PR #14, opened 2026-03-17
feat(nix): flake and hm module

Karrq:
This PR packages this really neat project with nix, and ships a nixos (and nix-darwin) module to set it up easily


jleni:
Very cool to see you around here @Karrq  :)

PR #32, opened 2026-04-27
[codex] add kunobi auth and ha support

jleni:
## Summary
- Integrates kache-service auth through `kunobi-auth` while keeping the existing bearer-token behavior.
- Adds optional `kunobi-ha` Kubernetes Lease leader election with readiness gating.
- Extends the Helm chart with HA values, ServiceAccount/RBAC, HA env wiring, and README/NOTES guidance.
- Documents that server-side kache is still work in progress.

## Validation
- `mise exec -- cargo test -p kache-service`
- `helm lint ./charts/kache-service`
- `helm template kache ./charts/kache-service --set ha.enabled=true --set replicaCount=2`
- `git diff --cached --check`

## Notes
- This PR does not deploy to `_platform`.
- For HA with PVC-backed planner state, the storage must support all scheduled replicas or `replicaCount` should stay at 1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -808,6 +808,12 @@ checksum = "4c7f02d4ea65f2c1853089ffd8d2787bdbc63de2f0d29dedbcf8ccdfa0ccd4cf"
 
 [[package]]
 name = "base64"
+version = "0.21.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


jleni:
Recreated as #33 using branch kache-kunobi-auth-ha without the codex/ prefix.

Issue #182, opened 2026-05-29
GC can delete a blob while a concurrent restore is mid-link


jleni:
## Problem
The cache-hit restore path holds no lock. `materialize_cached_artifact` (`src/wrapper.rs`) checks `store_path.exists()` (wrapper.rs:945) and then links the blob into place via `link_to_target` (wrapper.rs:983) / reads it. Concurrently the daemon's eviction (`evict` → `remove_entry` → `decrement_blob_refcount`, store.rs:938) can `remove_file` the blob once its refcount hits 0.

Between the `exists()` check and the actual hardlink/reflink/read, the blob can be deleted, so the link fails with ENOENT and the build aborts — on what was a valid cache hit.

## Trigger
Heavy concurrent builds + daemon eviction selecting the very entry being restored. Small window, but real under the concurrent-build workload kache targets.

## Relationship to other issues
- #181 (wrapper fallback-on-error) downgrades the **symptom** from a hard build failure to a transparent recompile — once #181 lands this is no longer build-breaking, just a wasted restore.
- The deeper fix belongs with the store-hardening / reachability-GC track (#180): never delete a blob that may be referenced or has an in-flight restore.

## Fix options
- Refcount-aware / reachability GC that never deletes a referenced blob.
- Treat ENOENT on restore as a miss and recompile (subsumed by #181).

jleni:
### Additional code-level depth (from a deep-review pass)

New depth on the GC-vs-restore TOCTOU: the restore **read** path (`store.get` at `wrapper.rs:681` →
`materialize_cached_artifact` which re-stats at `wrapper.rs:1074` then `std::fs::read` at `:1094` or
`link_to_target` at `:1113`) holds **neither** the SQLite write lock **nor** a KeyLock, and GC runs
in a **separate daemon process** with a different `Store`/`Connection` (`daemon.rs:~2242`) — so the
`store.rs:1007-1013` invariant (which only serializes put/import/remove_entry on one connection) does
**not** cover it.

Specifically, `get()`'s `last_accessed` bump (`store.rs:385`) does **not** make the entry
eviction-safe: `evict_older_than` (`store.rs:925`) and `evict_duplicate_entries` (`store.rs:959`)
select victims by age/duplicate predicates, not recency, so they can still target an entry being
restored.

Confirmed **safe-failing today** (unlink-not-truncate; open/read/hard_link/reflink all
[...]

Issue #181, opened 2026-05-29
Wrapper must fall back to the real compiler on cache errors (not abort the build)


jleni:
## Problem
The wrapper treats operational cache errors as fatal. In `run` (`src/wrapper.rs`) these all propagate via `?`:
- `store.get(&cache_key)?` — wrapper.rs:634
- `restore_from_cache(...)?` — wrapper.rs:647, :705, :749
- `store.try_lock(&cache_key)?` — wrapper.rs:740
- `store.wait_for_committed(&cache_key)?` — wrapper.rs:745

None of these are cache **misses** (a miss is `Ok(None)` and is handled correctly). They are operational **failures**: a locked/corrupt SQLite index, a missing/GC'd/truncated blob (`materialize_cached_artifact` bails "blob missing", wrapper.rs:946), or a hardlink/copy failure. Any of them aborts the user's build with a kache error — on a build that would otherwise compile fine.

## Why it matters
The wrapper's core invariant is **transparency**: on ANY cache problem it must run the real compiler with identical args/env/cwd and propagate the exact exit code/stdout/stderr. The miss path already does this; these error paths don't. Fixing this also makes the store-integrity races (e.g. GC deleting a blob mid-restore) degrade to a transparent recompile instead of a hard failure — high leverage.

## Fix
Route cache-layer errors (lookup, lock, restore) to the existing passthrough/recompile path instead of `?`-propagating; log at debug/warn first. Keep the genuine post-compile 'store the result' error handling separate.

## Notes
- Target: 0.4.0.
- Reduces the user-visible impact of #180 and the GC-mid-restore race.

jleni:
Shipped in 0.4.0. Fixed by #184 (`fix(wrapper): fall back to the real compiler on cache errors`), merged to `dev` and promoted to `main` via #174 (commit 6e5c108). Closing manually for parity with #178/#180 — the auto-close keyword didn't carry through the dev→main squash/merge.

PR #185, opened 2026-05-31
fix(store): atomic, durable entry registration (#180)

### LINKED ISSUE #180: Atomic entry registration in the local store
## Problem
The local store's `put` path is not transactional:
- `meta.json` / the DB row can be written before the blob contents are durably in place, leaving a window where a reader (or `contains`) can observe a half-written entry.
- Refcount mutations are read-modify-write split across separate autocommit statements, so concurrent wrapper processes can drift the count (leaked or prematurely-deleted blobs).
- No `fsync` before the "atomic rename", and `synchronous=NORMAL` — a crash after rename but before durability can leave a registered entry pointing at non-durable bytes.

## Proposal
An entry becomes *valid* only after its contents are fully and durably in place:
- Materialize all blobs first (write tmp → fsync → rename), then flip a single "valid" marker for the entry inside one transaction (`BEGIN IMMEDIATE … COMMIT`).
- Readers only ever see entries with the valid marker set.
- Wrap multi-statement refcount/registration mutations in transactions (or supersede refcounts entirely with reachability-based GC — tracked separately).

## Notes
- Removes the half-written-entry window and the non-transactional refcount drift surfaced in the store-integrity review.
- Pairs with the reachability-GC idea as the store-hardening track.

jleni:
Closes #180.

## Problem
The local store's write path was neither transactional nor durable:
- **Visibility before durability** — `meta.json` / the entry row could be written before the blob bytes were on disk (`synchronous=NORMAL`, no `fsync`). A crash could leave a `committed = 1` entry pointing at non-durable bytes.
- **Refcount drift** — blob refcount mutations were read-modify-write split across separate autocommit statements, so concurrent wrapper processes (the 300+ kache targets) could drift the count → leaked or prematurely-deleted blobs.
- **Shared temp path** — new blobs were copied to a fixed `<hash>.tmp`, so two writers of the same content could truncate each other mid-copy.

## Fix
An entry becomes visible only once it is fully durable, and all accounting is transactional:

1. **Phase 1** — hash and materialize every blob on disk first: copy to a **unique** temp name, **fsync**, atomic rename, mark read-only. No DB writes yet, so a crash here leaves at most orphan blob files (GC reclaims them), never a half-registered entry.
2. Write **and fsync** `meta.json`.
3. **Phase 2** — register every blob refcount **and** the `committed` entry row in **one transaction** (`unchecked_transaction`). Either the whole entry (with correct refcounts) becomes visible, or none of it does.

`remove_entry` and `import_downloaded_entry` get the same treatment: refcount mutations run inside one transaction, and blob files whose refcount hits zero are unlinked **only after commit** — a crash in that window leaves harmless orphans, never a DB row pointing at a deleted blob. `decrement_blob_refcount` is folded into the transactional `remove_entry`.

This also closes the read-side hazard structurally: `contains()` still requires `meta.json` **and** `committed = 1`, and now that flag flips atomically only after blobs + meta are durable.

## Testing
- New `test_concurrent_puts_sharing_blob_are_consistent`: **N=8 independent connections** (separate `Store`s, like real wrapper processes) concurrently cache distinct entries sharing identical content → exactly **one blob at refcount N**, every entry committed, **no `.tmp` debris**; removing all but one keeps the blob, removing the last reclaims it.
- Full store unit suite (62 tests) + end-to-end `stale_artifact_test` (12, incl. `stale_concurrent_builds_safe` and corruption recovery) pass; clippy `-D warnings` clean.

## Note for reviewers — fsync tradeoff
This adds an `fsync` per **new** blob + per `meta.json` on the cache-store (miss) path, which is what makes "durable before committed" meaningful. Dedup hits (existing blob) skip it. If this shows up as a cold-build latency regression under heavy parallelism, it's easy to gate behind a `durable_writes` config (default on) — happy to add that if you'd prefer. Flagging because kache's whole value is speed.

## Notes
- Target: 0.4.0. Second of the release-blocker series (after #181).
- Also makes #182 (GC mid-restore) degrade cleanly now that #181 + this are in: a missing/clobbered blob → transparent recompile + correct refcounts.

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -972,6 +972,14 @@ impl Store {
 
         // Reclaim blob files whose last reference is gone (post-commit, so the
         // DB never points at a file we've already deleted).
+        //
+        // Residual race (inherent to refcount GC, not closable across the
+        // FS/DB boundary): a concurrent `put` that dedups a *different* key
+        // against this blob can re-adopt it around this unlink, leaving its
+        // entry pointing at a deleted file. That is self-healing — `get`
+        // detects the missing blob, evicts the entry, and the wrapper
+        // recompiles (see #181). The complete fix is reachability-based GC;
+        // refcounts can't make the file unlink and the row state atomic.
         for hash in &zeroed {
             let blob = self.blob_path(hash);
             if blob.exists() {
--- a/src/store.rs
+++ b/src/store.rs
@@ -3,9 +3,29 @@ use rusqlite::{Connection, Error as SqlError, ErrorCode, params};
 use serde::{Deserialize, Serialize};
 use std::fs;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
 use crate::config::Config;
+
+/// Process-local counter for unique blob temp-file names.
+static BLOB_TMP_NONCE: AtomicU64 = AtomicU64::new(0);
+
+/// fsync a file so its bytes are durable on disk before we rename it into the
+/// content-addressed store or reference it from a committed entry. Opening
+/// read-only is sufficient for `fsync(2)`.
+fn fsync_file(path: &Path) -> std::io::Result<()> {
+    fs::File::open(path)?.sync_all()
+}
+
+/// A unique temp path beside the destination blob. Concurrent writers of the
+/// same hash must not share a temp file (a fixed `<hash>.tmp` lets one truncate
+/// the other mid-copy), so we disambiguate by pid + a process-local counter.
+fn blob_tmp_path(blob: &Path, hash: &str) -> PathBuf {
+    let pid = std::process::id();
+    let nonce = BLOB_TMP_NONCE.fetch_add(1, Ordering::Relaxed);
+    blob.with_file_name(format!(".{hash}.{pid}.{nonce}.tmp"))
+}
 
 /// Exclude a directory from Time Machine backups and Spotlight indexing.
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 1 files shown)


jleni:
### Concurrency note for reviewers

This PR **removes** the refcount-drift race (now one transaction per mutation) and the shared-`.tmp` corruption race, and tightens visibility (`committed=1` flips only after blobs+meta are durable). No new races, no transaction nesting (store access is serialized per process).

**One residual race remains** and is *inherent to refcount GC* (not introduced here, but its window is slightly wider with the phase split): a `put` that dedups a *different* key against a blob can race a concurrent `remove`/GC reclaiming that blob's last other reference → an entry briefly referencing a just-deleted file. It's **self-healing** — `get` evicts on the missing blob and the wrapper recompiles (#181) — so no corruption / wrong output / build failure. It can't be closed across the FS/DB boundary with refcounts; the complete fix is reachability-based GC, now tracked in **#186**. Documented inline in `remove_entry`.

jleni:
Update — the residual race I flagged earlier is now **closed**, not just documented (commit bcb8549).

Blob-file mutations and refcount mutations are now mutually exclusive: `remove` unlinks the blob **inside** its transaction, and `put`/`import` ensure/re-materialize the blob **inside** theirs. WAL allows one writer at a time, so a `put` adopting a blob and a `remove` reclaiming it can't interleave — whichever holds the write lock first wins cleanly (put re-materializes under the lock if remove got there first; remove won't hit zero if put got its reference first).

New test `test_concurrent_put_remove_never_dangles` (8 connections churning one shared blob) asserts no just-put entry ever dangles and the blob is fully reclaimed once settled. #186 is reframed as an optional future simplification, not a correctness gap.

Issue #82, opened 2026-05-15
Windows: run kache-e2e harness on Windows in CI


emmanuelm41:
## Gate

Subissue of #45. Windows support should not be considered "done" while the `kache-e2e` harness only runs on macOS/Linux. Until e2e runs on Windows, every Windows-specific code path lands without real validation and regressions are invisible until a user reports them.

## What's missing

- `crates/kache-e2e` does not run on Windows. The CI `windows-check` job only does `cargo check --target x86_64-pc-windows-msvc`.
- Windows-specific work already merged (#64 IPC, #76 Task Scheduler) has no e2e coverage.
- Path handling work in flight (#77) lands without test backing on Windows.

## Acceptance

- A Windows runner job in `.github/workflows/ci.yml` executes the existing `kache-e2e` harness.
- The job is required for merge (not exploratory).
- #45 stays open until this lands.

## Relation to #77

#77 bundles PathNormalizer fixes with the Windows e2e runner. This issue exists as a standalone gate so #45 can't close on the strength of the implementation alone — the harness must be running.

PR #204, opened 2026-06-01
fix(e2e): resolve inspected artifact with platform exe suffix (#202)

### LINKED ISSUE #202: Windows e2e: harness binary-inspection looks for the artifact without .exe
Surfaced by the Windows e2e arm (#82). Harness-only (`crates/kache-e2e`).

`multi-dep` and `rust-workspace` fail verify even though the binary runs fine (exit 0, correct stdout). The **binary-inspection** step can't find the artifact:

```
"failure_reason": "binary inspection: artifact not found at
  `C:/.../test-projects/multi-dep\\./target/release/multi-dep`"
```

On Windows the artifact is `multi-dep.exe` / `app.exe`; the inspection path (in `runner.rs`, the `forbidden_substrings`/`strings` step) doesn't apply the platform `.exe` suffix. Note also the `\\.` separator glitch in the path. Fix: resolve the inspected artifact with `std::env::consts::EXE_SUFFIX` (and normalize the join).

emmanuelm41:
Closes #202.

## Problem
The Windows e2e arm (#82) failed `multi-dep` and `rust-workspace` at the **binary-inspection** verify step even though the binaries ran fine (exit 0, correct stdout):

```
verify.failure_reason = "binary inspection: artifact not found at
  `C:/.../test-projects/multi-dep\./target/release/multi-dep`"
```

Two harness-only issues in `crates/kache-e2e/src/runner.rs::inspect_binary`:
1. It checked `cwd.join("./target/release/foo").exists()`, but on Windows the real artifact is `foo.exe` — the platform exe suffix was never applied.
2. `Path::join` left mixed `\.`/`/` separators in the reported path.

## Fix
- New `artifact_candidates(path, exe_suffix)` — pure, no filesystem — tidies the path with the existing `portable_path()` helper, then yields `<path>` followed by `<path><EXE_SUFFIX>` (skipping a double suffix). Testable on any host.
- New `resolve_artifact(path)` uses `std::env::consts::EXE_SUFFIX` and returns the first candidate that exists on disk.
- `inspect_binary` now resolves via `resolve_artifact` and reports the not-found path through `portable_path` for clean separators.

No-op on Unix (`EXE_SUFFIX` is empty), mirroring the `exe_suffix` matrix var already used in `ci.yml`.

## Tests
Added 5 unit tests for the candidate/resolve logic. `cargo fmt`, `cargo clippy --all-targets`, and `cargo test -p kache-e2e` (25 passed) all green locally.

Validation: the Windows e2e arm is non-blocking and runs on the self-hosted `kunobi-windows` runner; this PR's CI run will exercise it.

### DIFF
--- a/crates/kache-e2e/src/runner.rs
+++ b/crates/kache-e2e/src/runner.rs
@@ -14,7 +14,7 @@
 //! metric assertions impossible to write tightly.
 
 use anyhow::{Context, Result};
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 use std::time::Instant;
 use tempfile::TempDir;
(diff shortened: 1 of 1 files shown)


emmanuelm41:
Superseded by #209, which consolidates the remaining Windows e2e fixes (this PR's changes are included there as a dedicated commit) so the whole arm can be driven to green in one CI run. Closing in favor of #209.

Issue #196, opened 2026-06-01
Windows: kache fails to store cache entries ("flushing blob to disk")


emmanuelm41:
Surfaced by the new Windows e2e arm (PR #194, part of #77).

On the self-hosted Windows runner, **every** cache store fails:

```
WARN kache::wrapper: failed to store cache entry: flushing blob to disk
WARN kache::wrapper: failed to store cc cache entry for foo.c: flushing blob to disk
```

The compile itself runs, but the blob never lands in the cache, so the warm rebuild can't hit — every fixture with a `min_hits` / `max_compiler_runs = 0` warm assertion fails (c-depinfo, c-hello, cpp-hello, manifest-dir-runtime-workspace, multi-dep, out-dir-runtime, rust-check, rust-debug, rust-workspace). Fixtures that never store (c-passthrough, exclude-c, exclude-rust) pass.

Cache dir during the run was under the NetworkService profile temp (`C:\Windows\ServiceProfiles\NetworkService\AppData\Local\Temp\.tmpXXXX`). Likely a blob flush/rename/fsync path that doesn't hold on NTFS / under that profile.

This is the dominant blocker to making Windows e2e green. Caching does not currently function on Windows at all.

PR #205, opened 2026-06-01
fix(probe): exclude volatile env pseudo-vars from probe key (#201)

### LINKED ISSUE #201: Windows: cc compiler-probe not memoized across builds (warm re-runs the probe)
Surfaced by the Windows e2e arm (#82). **Not** a #196 regression — #196 (cache store) works here.

On the warm build the cc compiler-probe (`cc --version` / `cc -###`) re-runs instead of reading the memoized record, so `probe_runs = 1` on warm. `c-hello`/`c-depinfo` assert `max_probe_runs = 0` (probe memoized from cold) and fail. From the Windows `results.json`:

```
{"name": "max_probe_runs", "expected": "<= 0", "actual": "1", "passed": false}
```

Everything else on that phase passes — the compile is a `local_hit` (`compiler_runs: 0`), `diff_match[build/foo.o]` is byte-identical, verify ok. Only the probe memo isn't hitting.

Likely the probe record store/load in `src/probe/cache.rs` (`NamedTempFile::new_in` + `persist`, or the record lookup path) doesn't work on Windows. kache-core.

emmanuelm41:
Closes #201.

## Problem
On the Windows e2e arm (#82) the compiler-probe memo missed on the **warm** build, re-running the probe — `probe_runs = 1` where `c-hello` / `c-depinfo` assert `max_probe_runs = 0`:

```
{"name": "max_probe_runs", "expected": "<= 0", "actual": "1", "passed": false}
```

Everything else on that phase passed (compile `local_hit`, `compiler_runs: 0`, diff byte-identical, verify ok) — only the probe memo failed to hit. Not a #196 regression.

## Root cause
The probe key (`src/probe/cache.rs::probe_key`) mixes in `env_fingerprint()`, which hashes the **entire** process environment. On Windows `std::env::vars()` surfaces cmd.exe's hidden `=`-prefixed pseudo-variables:
- `=C:`, `=D:`, … — the per-drive current working directory
- `=ExitCode` — the previous child process's exit status

These shift between the cold and warm `make` invocations, so the probe key differed across builds and the warm `load()` looked up a `{key}.json` that cold never wrote → miss → re-probe.

This is Windows-only because Unix variable names cannot contain `=`. It breaks *only* the probe memo (not the compile cache) because the main cache key (`cache_key.rs`) hashes only a curated env subset (`RUSTFLAGS`, `CARGO_ENCODED_RUSTFLAGS`, `CARGO_CFG_*`) — which stays stable — exactly matching the observed "compile hits, probe re-runs".

## Fix
Filter `=`-prefixed names out of the env fingerprint. Refactored the hashing into a pure `fingerprint_env` core (+ `is_volatile_env_name` predicate) so the filtering is unit-testable without mutating the real environment. No-op on Unix.

## Tests
Added 3 regression tests (`fingerprint_env_ignores_windows_hidden_pseudo_vars`, `fingerprint_env_still_reflects_real_vars`, `is_volatile_env_name_flags_only_equals_prefixed`). `cargo fmt`, `cargo clippy --all-targets`, and `cargo test -p kache` (626 passed) all green locally.

Validation: the Windows e2e arm (non-blocking, self-hosted `kunobi-windows` runner) on this PR's CI run should now show `c-hello` / `c-depinfo` warm with `probe_runs = 0`.

### DIFF
--- a/src/probe/cache.rs
+++ b/src/probe/cache.rs
@@ -50,8 +50,7 @@ pub fn probe_key(prober_id: &str, req: &ProbeRequest<'_>) -> Option<String> {
     Some(h.finalize().to_hex().to_string())
 }
 
-/// Hash of the process environment — every `VAR=value` pair, sorted
-/// for determinism.
+/// Hash of the process environment — see [`fingerprint_env`].
 ///
 /// `cc -###` inherits this environment, so a change to it (a different
 /// `SDKROOT`, `CPATH`, …) can change the resolved invocation. Hashing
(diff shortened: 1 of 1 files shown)


emmanuelm41:
Superseded by #209, which consolidates the remaining Windows e2e fixes (this PR's changes are included there as a dedicated commit) so the whole arm can be driven to green in one CI run. Closing in favor of #209.

Issue #299, opened 2026-06-11
Firefox fails to build sandbox with kache on Windows


saschanaz:
## What happened

```
 1:50.46 E D:/firefox/security/sandbox/chromium/base/location.cc(67,15): error: static assertion failed due to requirement 'StrEndsWith("D:/firefox/security/sandbox/chromium/base/location.cc", kStrippedPrefixLength, "base\\location.cc")': The file name does not match the expected prefix format.
 1:50.46 E    67 | static_assert(StrEndsWith(__FILE__, kStrippedPrefixLength, "base\\location.cc"),
 1:50.47 E       |               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 1:50.47 E 1 error generated.
 1:50.47 E mozmake[4]: *** [D:/firefox/config/rules.mk:666: location.obj] Error 1
 1:50.49 E mozmake[3]: *** [D:/firefox/config/recurse.mk:79: security/sandbox/target-objects] Error 2
```

## Expected behaviour

Should not fail. Maybe kache converts backslashes into forward slashes?

## Steps to reproduce

1. `./mach build`

## Environment

- `kache --version`: 0.5.1-rc.4
- OS and version (e.g. macOS 15.4, Ubuntu 24.04): Windows 11
- Rust toolchain (`rustc --version`): rustc 1.95.0 (59807616e 2026-04-14)
- Install method (mise / cargo-binstall / cargo install): cargo install
- Remote backend (none / S3 / R2 / MinIO / other): none

## Logs

```
```

saschanaz:
Looks sus:

https://github.com/kunobi-ninja/kache/blob/fdd303a3bb0ab89f4868cd6168da181103f1c9c1/crates/kache-e2e/src/lib.rs#L37-L39

emmanuelm41:
Thanks for the detailed report — and thanks @saschanaz for digging in.

[...]

PR #303, opened 2026-06-11
fix(cc): suppress -ffile-prefix-map for the GNU driver on Windows (#299)

emmanuelm41:
Follow-up to #295, closing the residual half of #299.

## Context

#295 (clang-cl support) stopped injecting `-ffile-prefix-map` for the **clang-cl** driver, which fixed the reported #299: Firefox on Windows builds via `clang-cl`, so `__FILE__` is no longer rewritten.

But the plain `clang` / `gcc` **GNU driver on Windows** was left on the old path. `cc_prefix_maps_cfg` still derived maps for it, so `execute` injected `-ffile-prefix-map` — and clang's prefix mapper on Windows rewrites `__FILE__` and normalizes its `\` separators to `/`. That's the exact mechanism that broke Firefox's `static_assert(StrEndsWith(__FILE__, …, "base\\location.cc"))`. A GNU-driver Windows build that asserts on the native `__FILE__` spelling would hit the identical failure.

## Change

Extend the existing clang-cl guard in `cc_prefix_maps_cfg` to also return **empty maps for any compile on Windows**:

```rust
if parsed.family.dialect() == Dialect::Cl || is_windows {
    return Vec::new();
}
```

Empty maps keep the key path-literal (per-machine — misses, never a miscache) and inject nothing — exactly the clang-cl remedy. **Off-Windows GNU is unchanged**: there's no separator quirk on Unix, so cross-machine `-ffile-prefix-map` normalization stays on.

`cc_prefix_maps_cfg` now takes the host as an `is_windows` parameter (the caller passes `cfg!(windows)`), keeping its deterministic core pure so **both platform branches are unit-testable on any machine** — the same host-independent approach #295 used for the clang-cl shapes. Because both the cache-key and the compile read the same Windows-aware result, key and object stay consistent (no miscache).

## Tests

- `cc_prefix_maps_empty_for_gnu_on_windows` (new) — GNU driver + `is_windows=true` → empty; `is_windows=false` → non-empty.
- `cc_prefix_maps_empty_for_clang_cl` (extended) — clang-cl empty on both hosts.

715 bin tests pass; `cargo fmt --check` and `cargo clippy --all-targets -D warnings` clean. README's cache-key portability note now scopes gcc/clang portability to non-Windows hosts.

## Not yet validated on hardware

This is logically complete and safe regardless — worst case it makes Windows-GNU keys per-machine, which is *correct*, just less shareable. But I have **not** confirmed on a real Windows box that GNU-mode `clang` actually mangles `__FILE__` separators (my local test on macOS only proved the `-fmacro-prefix-map` rewrite, not the Windows-specific separator normalization). A minimal `clang -c sub\probe.c` repro on Windows is the remaining confirmation; happy to run it before merge if preferred.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


emmanuelm41:
Closing this — Windows-hardware testing disproved its premise.

On a real Windows box (clang/clang-cl 19.1.5, VS2022), I tested `__FILE__` under prefix-map injection directly:

| case | `__FILE__` result |
|------|-------------------|
| GNU `clang.exe`, `-ffile-prefix-map` no-match | `C:\...\probe.c` — **backslashes preserved** |
| GNU `clang.exe`, `-ffile-prefix-map` match | `SENT\sub\probe.c` — prefix replaced, **separators stay `\`** |
| `clang-cl.exe`, `-ffile-prefix-map` match | `C:\...\probe.c` — **ignored entirely** (not even prefix-replaced) |

So the GNU driver on Windows does **not** mangle `__FILE__` separators, and clang-cl ignores `-ffile-prefix-map` outright. This PR's premise (that `-ffile-prefix-map` forward-slashes `__FILE__` on the GNU-Windows path) is wrong, and merging it would disable Windows-GNU cross-machine caching for no correctness benefit.

This also means the original #299 root-cause theory (shared by the now-closed #301 and an earlier comment on #299) is incorrect — the forward-slashed `__FILE__` does not come from `-ffile-prefix-map`. Re-investigating the true mechanism with the rc.4 binary before proposing any fix.

PR #355, opened 2026-06-18
fix(extra_inputs): warn on unset env var in a pattern

emmanuelm41:
## Problem

An unset `$VAR` / `${VAR}` in an `extra_inputs` pattern was the **one** failure mode this module swallowed silently. The reference stays a literal, matches nothing on disk, and folds a *pattern-set-only* key that replays regardless of the files the author intended to track — while every other rejection path (control separator, bad glob, out-of-crate, filesystem-root walk, all-rejected) already warns loudly.

## Fix

Add `expand_env_vars_collecting` / `expand_exclude_pattern_collecting`, which return the names of referenced-but-unset vars alongside the expanded string, and warn in `normalize_pattern` when any survive. Vars with a built-in default (e.g. `CARGO_HOME`) are not reported. Folding behavior is otherwise unchanged (still fail-safe).

## Context

Surfaced by @lovesegfault's review in [rio-build#51](https://github.com/lovesegfault/rio-build/pull/51): *"warn on unset `$VAR` in `extra_inputs` patterns: it stays a literal, matches nothing, and folds a replayable pattern-set-only key. everything else in `extra_inputs.rs` fails loudly, this one is silent."*

## Test

`expand_collecting_reports_unset_vars_only_once` (dedup + first-seen order) and `expand_collecting_no_unset_when_resolved_or_defaulted` (resolved + defaulted vars not reported). Clippy clean.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -760,15 +760,25 @@ fn shellexpand(s: &str) -> PathBuf {
     PathBuf::from(s)
 }
 
-fn expand_env_vars(s: &str) -> String {
-    expand_env_vars_with(s, |key| std::env::var(key).ok())
-}
-
-fn expand_env_vars_with<F>(s: &str, lookup: F) -> String
+/// Core `$VAR` / `${VAR}` expander. Returns the expanded string plus the names
+/// of every referenced env var that was *unset* (no value and no
+/// [`default_env_var_value`]) and so was left as a literal `$VAR` in the output.
+///
+/// An unset reference matters to cache-key callers: it silently survives as
+/// text that matches nothing, so they fold a replayable pattern-set-only key
+/// while believing the intended files are tracked. Reporting the unset names
+/// lets those callers warn instead of degrading silently.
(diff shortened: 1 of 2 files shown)


emmanuelm41:
- Removed `expand_env_vars`/`expand_env_vars_with` wrappers — confirmed **no remaining callers** anywhere on the branch; `expand_exclude_pattern` now delegates to `expand_exclude_pattern_collecting(..).0`.
- `expand_exclude_pattern` output is **byte-identical** for existing cc/exclude callers (same lookup closure + `default_env_var_value` + shellexpand order).
- Both `$VAR` and `${VAR}` unset paths record the var; defaulted vars (e.g. `CARGO_HOME`) are **not** reported. Dedup is first-seen, no panic.
- Cache **key is unaffected** — `unset_vars` is consumed only by `tracing::warn!` in `normalize_pattern`; the folded value is unchanged.

emmanuelm41:

PR #356, opened 2026-06-18
fix(config): warn on malformed size strings instead of degrading silently

emmanuelm41:
## Problem

A `KACHE_MAX_SIZE` / `[cache] local_max_size` / `[cache] event_log_max_size` value that `ByteSize` can't parse (a typo'd unit like `100 gigs`, digit grouping like `1_000`, plain garbage) fell through **silently** to the next source and finally to the hardcoded default — so the cap the user asked for was ignored with no warning.

## Fix

Add `parse_size_checked(value, source)`, which `tracing::warn!`s (naming the source) when a *set* value is unparseable, and route the three load-time size reads through it. The TUI's own field validation keeps using the bare `parse_size`. Fall-through behavior is unchanged — only now it's visible.

## Context

Surfaced by @lovesegfault's review in [rio-build#51](https://github.com/lovesegfault/rio-build/pull/51): *"loud size-string validation: `KACHE_MAX_SIZE=1.5GiB` silently degrades through the file value to the 50GiB default."*

Note: `ByteSize` 2.x **does** accept fractional sizes now (`1.5GiB` → 1610612736), so the report's exact example parses today. This PR addresses the broader silent-degradation the point was about; the test pins the values that genuinely fail.

## Test

`parse_size_checked_rejects_malformed_and_mirrors_parse_size` covers the malformed set + a valid value passthrough. Clippy clean.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -254,14 +254,14 @@ impl Config {
 
         let max_size = std::env::var("KACHE_MAX_SIZE")
             .ok()
-            .and_then(|s| parse_size(&s))
+            .and_then(|s| parse_size_checked(&s, "KACHE_MAX_SIZE"))
             .or_else(|| {
                 file_config
                     .as_ref()
                     .ok()
                     .and_then(|c| c.cache.as_ref())
                     .and_then(|c| c.local_max_size.as_ref())
-                    .and_then(|s| parse_size(s))
+                    .and_then(|s| parse_size_checked(s, "[cache] local_max_size"))
             })
             .unwrap_or(50 * 1024 * 1024 * 1024); // 50 GiB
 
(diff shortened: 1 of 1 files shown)


emmanuelm41:
- `parse_size_checked` = `parse_size` + a `warn!` on `None`; for valid values the result is **byte-identical**, fallback chain (env→file→default) structurally unchanged.
- All 3 call sites only invoke it when a value is **present** (`.ok().and_then` / `.as_ref().and_then`), so an absent env var / missing TOML key never warns spuriously.
- Defaults (50GiB / 10MiB) unchanged; TUI still uses bare `parse_size`; no cache-key impact.
- Test pins genuinely-rejected values (`100 gigs`, `1_000`, `abc`, ``) — verified against bytesize 2.3.1 (which now *accepts* `1.5GiB`, so that's correctly avoided).

emmanuelm41:

PR #357, opened 2026-06-18
feat(observability): trace key_salt application and surface it in why-miss

emmanuelm41:
## Problem

`apply_key_salt` folded the salt into every key **silently**, so a miss caused by a stray machine-global `KACHE_KEY_SALT` (or a rotated salt) was invisible even at `KACHE_LOG=trace` — and the `why-miss` grep recipe (`[key:<crate>]`) couldn't show it either.

## Fix

- Emit a `[key:<crate>] key_salt=… -> <key-prefix>` trace line when a salt is applied, matching the other key-component trace lines (so the existing grep recipe surfaces it). `apply_key_salt` now takes the crate/source label. The unsalted path stays **byte-identical** and silent — no `CACHE_KEY_VERSION` impact.
- Surface the active `key_salt` in `kache why-miss` output. The salt isn't recorded per entry (a separate, larger change), so it can't be diffed against a stored entry — but showing it makes a salt-induced miss diagnosable.

## Context

Surfaced by @lovesegfault's review in [rio-build#51](https://github.com/lovesegfault/rio-build/pull/51): *"observability for the new key parts: `apply_key_salt` emits no tracing, so a salt-induced miss is invisible even at `KACHE_LOG=trace`… happy to PR the trace line + why-miss fields."*

## Test

Existing `apply_key_salt_*` tests updated for the new label param; unsalted-identity invariant still asserted. Clippy clean.

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -200,9 +200,22 @@ fn scrub_remap_value(value: &str) -> String {
 ///
 /// Shared by every compiler family (rustc and cc) so the salt applies
 /// uniformly regardless of which adapter produced `base`.
-pub(crate) fn apply_key_salt(base: String, salt: Option<&str>) -> String {
+///
+/// `label` is the crate/source name used in the `[key:…]` trace line so a
+/// salt-induced miss is visible under `KACHE_LOG=trace` alongside the other key
+/// components — previously the salt was the one key part that folded silently,
+/// so a miss caused by a (stray or rotated) salt was invisible to the
+/// `why-miss` grep recipe.
+pub(crate) fn apply_key_salt(base: String, salt: Option<&str>, label: &str) -> String {
     match salt {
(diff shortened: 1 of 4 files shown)


emmanuelm41:
- Cache **key is byte-identical**: `apply_key_salt` still returns `fold_labeled(base,"key_salt",salt)`; the new `label` is used **only** in the `tracing::trace!` macro, never hashed.
- Unsalted path (`None`/empty) remains a no-op returning `base`, emits no trace.
- `&keyed[..keyed.len().min(16)]` — safe: key is always 64 ASCII hex chars.
- Rebased onto #354: its `rustc_refuse_reasons` (incl. the `--pretty/--unpretty` guard) is intact; the salt change and the refuse-reason coexist cleanly in `rustc.rs`.
- `why-miss` change is output-only (`config.key_salt.as_deref().filter(..)`, no unwrap).

emmanuelm41:

PR #358, opened 2026-06-18
fix(passthrough): pre-clean read-only restores; log when outputs are unidentifiable

emmanuelm41:
## Problem

A prior cache hit restores read-only (0444) hardlinks into the target dir. The cached compile path pre-cleans them (`run_rustc`) and the disabled/re-entrant path does too (`run_compiler_directly`) — but a kache-**declined** passthrough (a refuse reason, non-primary, etc.) ran straight into the read-only outputs and rustc failed with EACCES. And when a lenient/partial parse couldn't recover `crate_name`/`extra-filename`, `pre_clean_outputs` identified nothing and **silently** cleaned nothing.

## Fix

- Call `pre_clean_outputs` in `wrapper::passthrough`, before both the fallback-wrapper and plain-rustc paths.
- Add a `debug!` line in `pre_clean_outputs` for the no-identity case. A conservative "remove every read-only file in the dir" sweep would risk deleting unrelated cached artifacts, so it **reports rather than guesses** — exactly the reviewer's "or at least a debug line" suggestion.

## Context

Surfaced by @lovesegfault's review in [rio-build#51](https://github.com/lovesegfault/rio-build/pull/51) (re kache PR #242): *"when the lenient parse can't recover crate_name/extra-filename, the passthrough cleans nothing. A conservative fallback (or at least a debug line) would close the gap for weird spawners."*

## Test

`test_pre_clean_without_identity_is_a_safe_noop` (no-identity → no files touched, no panic). Existing pre_clean + `run_compiler_directly` tests still green. Clippy clean.

### DIFF
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -418,6 +418,20 @@ pub(crate) fn pre_clean_outputs(
                 }
             }
         }
+    } else {
+        // Neither a concrete output path nor an (out_dir, crate_name) pair to
+        // scan: a lenient/partial parse (an unusual rustc spawner) left nothing
+        // to identify the outputs by, so we can't pre-empt a warm read-only
+        // restore and the compiler may then fail with EACCES. Surface it at
+        // debug so the failure mode is diagnosable rather than silent — a
+        // conservative "remove every read-only file in the dir" sweep would
+        // risk deleting unrelated cached artifacts, so we report instead of
+        // guess (rio-build#51 / kache#242).
+        tracing::debug!(
+            "pre_clean_outputs: nothing to identify outputs by \
+             (out_dir={out_dir:?}, crate_name={crate_name:?}); skipping read-only pre-clean — \
+             a warm restore in this directory may block this compile"
+        );
     }
 }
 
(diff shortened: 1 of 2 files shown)


emmanuelm41:
- `pre_clean_outputs` only ever removes **read-only** files (gated by `remove_if_readonly`) and matches crate-scoped output names — can't touch a user's writable sources/outputs.
- No double-clean: `passthrough` runs rustc directly (and the fallback wrapper), never via `run_rustc`; the three `pre_clean_outputs` call sites (run_rustc / run_compiler_directly / passthrough) are disjoint. (`remove_if_readonly` is idempotent anyway.)
- New call uses `args.output/out_dir/crate_name/extra_filename` — correct `RustcArgs` fields, same as `run_compiler_directly`.
- Cleaning before the fallback (sccache) path is correct — the fallback also needs writable outputs.
- New `else` branch in `pre_clean_outputs` is purely additive (debug log); the working if/else-if arms are byte-identical to main.

emmanuelm41:

PR #359, opened 2026-06-18
feat(doctor): warn when extra_inputs crates lack rerun-if-changed coverage

emmanuelm41:
## Problem

`extra_inputs` only re-keys when cargo invokes rustc. Editing a tracked file (e.g. `.sqlx/`) in a **warm in-place target** doesn't re-invoke rustc unless a build script emits a matching `cargo:rerun-if-changed` — so the stale artifact is restored while the author believes the file is tracked. There was no way to discover this footgun.

## Fix

`kache doctor` now scans the tree (skipping `target`/`.git`/hidden dirs, capped at 50k entries) for crates declaring `extra_inputs` and reports any whose build script won't re-trigger rustc:
- no `build.rs`, or
- a `build.rs` that emits no `rerun-if-changed`.

The check is **advisory and heuristic** — a static scan can't confirm the rerun path actually covers the declared globs, nor see a script that emits the directive without the literal string. It only appears when at least one crate declares `extra_inputs` (green when all covered, ✗ + guidance when not), so there's no noise for the common case.

## Context

Surfaced by @lovesegfault's review in [rio-build#51](https://github.com/lovesegfault/rio-build/pull/51) (point 1): *"maybe a `kache doctor` warning for `kache.toml`s in crates whose build script emits no matching `rerun-if-changed`."*

## Test

`audit_rerun_coverage_flags_only_uncovered_declaring_crates`: covers no-build.rs gap, covered build.rs, build.rs-without-directive gap, empty-declaration opt-out, and `target/` skipping. Clippy clean.

> Note: the companion docs caveat for this limitation lands in a separate docs PR.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2021,6 +2021,57 @@ pub fn doctor(
         });
     }
 
+    // extra_inputs warm-target coverage: a crate that declares extra_inputs but
+    // whose build script won't re-trigger rustc gets stale artifacts when a
+    // tracked file changes in a warm target. Only surfaced when at least one
+    // crate in the tree declares extra_inputs (no noise for the common case).
+    if let Ok(cwd) = std::env::current_dir() {
+        let audit = crate::extra_inputs::audit_rerun_coverage(&cwd);
+        if audit.declaring > 0 {
+            if audit.gaps.is_empty() {
+                checks.push(Check {
+                    label: "extra_inputs",
+                    pass: true,
+                    detail: format!(
+                        "{} crate(s), build scripts re-trigger rustc",
+                        audit.declaring
+                    ),
+                    fix: None,
+                });
+            } else {
+                let listed: Vec<String> = audit
+                    .gaps
+                    .iter()
+                    .take(5)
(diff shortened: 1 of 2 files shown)


emmanuelm41:
- **Diagnostic-only**: `audit_rerun_coverage` is called from `doctor()` only — not from `digest()` or any compile/cache-key path. Hot path untouched.
- Walk is safe: `DirEntry::file_type()` doesn't follow symlinks on Unix (no cycle/escape), DFS stack is bounded, and the 50k `AUDIT_WALK_LIMIT` early-returns. `target`/`.git`/hidden dirs pruned correctly.
- Malformed `kache.toml` is **silently skipped** (`let Ok(..) else { continue }`) — no panic, doctor proceeds.
- Output gating: the check only appears when ≥1 crate declares `extra_inputs`; a pass adds a green check (no issue-count bump), a gap adds one ✗. `doctor()` exit code unchanged.
- doctor now does one bounded DFS of cwd — acceptable for an on-demand command.

emmanuelm41:

Issue #394, opened 2026-06-20
E2E: validate path normalization across out-of-tree / sibling / symlinked build trees


jleni:
## Context

Spun out of #304. @mathstuf raised that an out-of-tree / sibling / symlinked build tree could degenerate kache's path normalization (the single "common ancestor" sentinel collapsing to `$HOME` and gaining nothing). Walking the code, the design already handles this — but **only by inspection**. We don't yet have an end-to-end scenario that *proves* keys converge across these topologies, and @emmanuelm41 committed in the thread to growing the nightly E2E suite to cover exactly these layouts. This issue tracks adding that coverage.

## What's already in place (so scope is validation, not implementation)

- **CC**: split sentinels `cwd → <CC_BUILD>` / `source_parent → <CC_SOURCE>` when there's no useful shared root (`cc_prefix_maps_for`, `src/compiler/cc.rs`); a `≥3 normal components` + reject-tempdir guard on the single common-root sentinel (`useful_cc_prefix` / `stable_cc_common_root`); roots computed on both literal and `canonicalize`d paths; `KACHE_BASE_DIR → <CC_BASE>` override.
- **Rust**: source identity is the content hash (path-independent), plus `PathNormalizer` rules (`<WORKSPACE>`/`<CARGO_HOME>`/`<RUSTUP_HOME>`/`<HOME>`).

## Scenarios to add to the E2E harness

1. **CMake out-of-tree, sibling build dirs** — `build`, `build-static`, `build-mpich` as siblings of one source tree. Assert cross-variant cache hits on shared (incl. generated) TUs.
2. **Symlinked build forest** — build/target dirs symlinked to a backup-excluded location (@mathstuf's actual layout: `builds → $buildsforest/$project`, `build → builds/build`). Assert keys converge through the symlink.
3. **Rust sibling / out-of-tree target** — `target → ../builds/target` symlink **and** a fully out-of-tree `CARGO_TARGET_DIR`. Assert warm cross-clone hits, and specifically check that build-script `OUT_DIR` / `CARGO_MANIFEST_DIR` env-deps still normalize when `target/` resolves outside the workspace (the one residual flagged in #304 — answer it concretely).
4. **`KACHE_BASE_DIR` override** — assert the explicit base-dir path strips to `<CC_BASE>` as expected.

## Acceptance

- Each topology is its own scenario in the e2e harness (`crates/kache-e2e`) with hit-rate assertions, wired into the nightly runs.
[...]

Issue #304, opened 2026-06-11
Concern about common root calculation


mathstuf:
## What happened

In [this recent blog post](https://kunobi.ninja/design-partner), I saw this:

> kache derives its path sentinel from the common ancestor of the working directory and the source file, which for an ordinary compile is the repo root

This seems to assume an under-source build tree. This may be true in CI (where you're generally given the repository checkout *as* the root you're allowed to use), but is definitely not true generally. At least I (and many other developers I know) use *sibling* directories to the source tree for build trees. This is more common in CMake land (FD: I am a CMake developer) where one might have multiple flavors of a build: `build-static`, `build`, `build-mpich`. Even with under-source build trees, these each would end up missing cache hits for generated sources that they do share.

Personally, I up the level more by using symlinks to shunt all build artifacts into a backup-excluded location:

```
- $project/
  - src/ (main checkout)
  - src-$foo/ (worktrees or related projects)
  - builds → $buildsforest/$project (symlink to build trees for $project)
  - build → builds/build
  - build-$foo/ → builds/build-$foo (project/worktree/variant builds)
```

For Rust projects, there is `builds/target`, `target → builds/target`, and `src/target → ../target`.

Any prefix calculation for such a setup on absolute paths means, basically, my home directory is the prefix and means the sentinel replacement gains nothing.

I think it is best to have a separate sentinel for the source tree and build tree in the general case.

## Expected behaviour

Separate sentinels for the source and build tree.

## Steps to reproduce

I've not actually tested this; just raising it based on the blog post mention.

emmanuelm41:
Thanks so much for raising this, and for the detail on your layout — the symlinked build-forest setup is exactly the kind of real-world configuration worth pressure-testing against, so this is genuinely useful.

You've read the blog sentence exactly as written, and the concern is spot-on *for what that sentence describes*. The post simplified things to keep the mental model easy to follow: it describes only the happy path, where the common ancestor of the build dir and the source file is the repo root. The actual implementation is more defensive than that one line implies, and I think it already covers your case — let me point at the specifics:

- **Separate source/build sentinels already exist.** When the common ancestor *isn't* a useful shared root, the code falls back to two distinct maps — `cwd → <CC_BUILD>` and the source dir → `<CC_SOURCE>` — which is precisely the "separate sentinels for source and build tree" you're asking for. (`cc_prefix_maps_for` in `src/compiler/cc.rs`, the `else` branch.)
[...]

PR #397, opened 2026-06-20
fix(cc): converge out-of-tree build keys + CMake e2e scenario (#304, #394)

jleni:
Closes the cc half of #304; part of #394.

## The bug
An **out-of-tree / sibling build tree** (a `build` / `cmake-build` dir *beside* the source tree, not under it — @mathstuf's #304 layout) failed to share the cc cache across locations: a relocated or cross-machine compile **missed** even though the produced object was **byte-identical**. A missed hit, not a correctness issue — but it defeats remote/cross-clone caching for exactly the CMake out-of-tree workflow #304 raised.

## Root cause
The cc prefix-map's single-`<CC_ROOT>` vs split-`<CC_BUILD>`/`<CC_SOURCE>` decision was made on **absolute-location** properties: the common ancestor's *depth* (`≥3` components) and whether it was a temp dir. For a sibling build the common ancestor is a *strict* ancestor of both build and source, and whether it earned a `<CC_ROOT>` sentinel **flipped with where the tree lived** (deep repo path → yes; `/tmp` or `$TMPDIR` → no). Because the prefix-map **sentinel set is hashed into the key**, the set flipping diverged the key.

## Fix — decide by topology, not location
`push_cwd_source_roots`:
- **Nested** (ordinary under-source build, common == build or source dir): single `<CC_ROOT>` — **unchanged**, no key change for existing builds.
- **Sibling / out-of-tree** (strict-ancestor common): always split into `<CC_BUILD>`/`<CC_SOURCE>`, and fold `<CC_ROOT>` whenever the common is not a **bare filesystem root** (has ≥1 normal component). That last test is **location-independent**, so the sentinel set is identical whether the tree lives in `~/proj`, a deep CI path, or `$TMPDIR`. Out-of-tree builds converge across machines / a relocate. (`/` is still excluded, to avoid over-mapping unrelated absolute paths.)

The objdir/Firefox cross-checkout case (also a sibling topology, deep repo root) is unaffected — it already folded `<CC_ROOT>` and still does.

## Tests
- **`e2e-cmake-out-of-tree`**: CMake builds into a sibling `../cmake-build`, generated header included via an absolute `-I` into the out-of-tree build dir, kache as `CMAKE_C_COMPILER_LAUNCHER`. `relocate` converges (`min_hits=1, max_misses=0`, byte-identical) and is **falsifiable** under `--negative-control`. (Requires `cmake`.)
- Unit test pinning the location-independent sentinel set (deep vs shallow root identical; bare `/` not folded).
- Full e2e suite (21 scenarios, no regressions) + 792 unit tests + clippy/fmt green.

### DIFF
--- a/crates/kache-e2e/src/fixture.rs
+++ b/crates/kache-e2e/src/fixture.rs
@@ -115,6 +115,16 @@ pub struct Fixture {
     #[serde(default)]
     pub requires: Vec<String>,
 
+    /// Operating systems this fixture supports, by `std::env::consts::OS`
+    /// value (`"linux"`, `"macos"`, `"windows"`). Empty (default) = all.
+    /// When non-empty and the current OS isn't listed, the harness SKIPS the
+    /// fixture (like a missing `requires` tool) instead of running it — for
+    /// fixtures that only make sense on some platforms, e.g. the cc out-of-tree
+    /// convergence scenario which exercises the gnu/clang prefix-map path and
+    /// not MSVC `cl` (kunobi-ninja/kache#394).
+    #[serde(default)]
+    pub os: Vec<String>,
+
     /// Overrides applied only when the harness runs on Windows (see
     /// [`OsOverride`]). Parsed on every platform but consulted only there.
     #[serde(default)]
--- a/crates/kache-e2e/src/fixture_runner.rs
+++ b/crates/kache-e2e/src/fixture_runner.rs
@@ -93,6 +93,21 @@ pub fn run(config: FixtureRunConfig) -> Result<()> {
     let mut fixture_results = Vec::new();
     let mut any_failed = false;
     for fixture in fixtures.values() {
+        if !fixture.os.is_empty() && !fixture.os.iter().any(|o| o == std::env::consts::OS) {
+            eprintln!(
+                "=> {}: skipped — not supported on {} (os = {:?})",
+                fixture.name,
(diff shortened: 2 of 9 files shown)


jleni:
**CI note:** the project deliberately doesn't provision `cmake` on its runners (it's not in `mise.toml`), so `e2e-cmake-out-of-tree` **skips** in CI via `requires = ["cmake"]` rather than failing. The cc fix itself is still CI-guarded by the new unit test `cc_prefix_maps_sentinel_set_is_location_independent_for_out_of_tree` (runs on every Test job), and the e2e scenario runs for real locally / on any cmake-equipped runner (verified locally: relocate converges + falsifiable under `--negative-control`).

jleni:
**Update — the e2e scenario now RUNS in CI** (addressing the skip concern):
- Added `cmake` to `mise.toml` so the Linux + macOS gate runners provision it.
- Added an `os` gate to the fixture loader: `os = ["linux", "macos"]` runs the scenario on the gnu/clang platforms (where the cc prefix-map fix applies) and skips it only on Windows, where MSVC `cl` is path-literal by design (out of scope for this fix — a separate track).

So the cc out-of-tree convergence fix is now exercised **end-to-end on Linux + macOS CI**, not just by the unit test. (Supersedes the earlier 'skips in CI' note.)

PR #416, opened 2026-06-22
feat(sync): add --workspace to scope sync --pull to workspace members

AlJohri:
## Summary

Adds `--workspace` to `kache sync`, scoping the pull's S3 key discovery to workspace members instead of every `Cargo.lock` dependency crate.

`sync --pull` issues one `ListObjectsV2` per crate in the Cargo.lock dependency set (`list_keys_for_crates`) to discover which artifacts are present — often hundreds-to-thousands of LIST calls. When the dependency artifacts are provided some other way (e.g. a prebuilt deps image whose compiled deps already sit in `target/`, so cargo never recompiles them and the wrapper never uploads them), only the workspace members are ever in S3, and the per-dependency LIST calls all find nothing.

`--workspace` reuses the existing crate-prefixed listing path with the workspace set kache already computes for the push filter (`cargo metadata --no-deps`), so discovery issues one LIST per workspace member (~tens) instead of one per dependency. Conflicts with `--all`.

This only narrows the **up-front batch** pull. Dependency artifacts are still resolved on demand during the build — the rustc wrapper fetches a local miss from S3 via the daemon (a remote hit) and the daemon prefetches by build intent — so deps are **not** recompiled. With deps already on disk they're local hits and never touch S3.

## Effect (real run, ~930-dependency / ~35-member workspace)

| | LIST calls | discovery time | keys found | pull plan |
|---|---|---|---|---|
| `sync --pull` | 932 | ~82s | 67 | pull 67 |
| `sync --pull --workspace` | 35 | ~4s | 67 | pull 67 |

Identical discovery result, ~27× fewer LIST calls. (`list_keys_for_crates` issues exactly one `list_objects_v2` per crate, so the crate count is the LIST count; the wall-clock difference is the eliminated round-trips.)

## Changes

- `--workspace` flag on `sync` (`conflicts_with = "all"`), threaded through `sync` → `sync_inner` → `sync_with_client` alongside `pull_all`.
- New pull arm: list `workspace_crates` instead of `lock_crates` when set; the Cargo.lock parse is skipped.
- Test asserting the workspace path lists only the workspace member(s).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2374,6 +2374,7 @@ pub fn sync(
     push_only: bool,
     dry_run: bool,
     pull_all: bool,
+    pull_workspace: bool,
 ) -> Result<()> {
     let remote = config
         .remote
@@ -2383,8 +2384,10 @@ pub fn sync(
     let store = Store::open(config)?;
     let workspace_crates = workspace_filter(manifest_path);
 
-    // For filtered pull: parse Cargo.lock to get all dependency crate names
-    let lock_crates = if !pull_all && !push_only {
+    // For the default filtered pull: parse Cargo.lock for every dependency crate
+    // name. Skipped under --workspace, which scopes the pull to workspace members
+    // only (the deps are expected to be provided some other way).
+    let lock_crates = if !pull_all && !pull_workspace && !push_only {
         parse_cargo_lock_crate_names()
     } else {
         None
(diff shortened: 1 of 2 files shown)


emmanuelm41 (approved):
Honestly, we are glad to see PRs contributing to the project. Looks correct. We may add some documentation updates to include this, and some more testing around it, but it looks solid. Thanks again!

jleni:
Thank you very much @AlJohri for your PR!! As Emmanuel said, we are very happy to see you contributing!! keep them coming!! :)

Issue #471, opened 2026-06-30
bench(surrealdb): cross-clone hit-rate gap — native-sys .a content divergence (quickjs, jemalloc) + DEBUG_OUTPUT_DIR/OUT_DIR cascade


emmanuelm41:
Sibling to #431 (substrate cross-clone gap). The new SurrealDB bench passes its gate but leaves most of the speedup on the table — and its root cause is a **different class** than substrate's `.rmeta` source-path cascade, so it's worth tracking separately.

## Data (run 28404901881, `just bench surrealdb --trace-keys`, linux-x86_64, v3.1.5)

- verdict **ok**; cold 28m09s → warm 27m54s; **speedup 1.01x**
- warm hit 97.3% by count but **47.5% weighted** — the expensive crates miss
- key stability 96.9% (560/578); 0 leak-detector firings
- costliest warm misses: `surrealdb_core` (5m), `surrealdb_server`, `surrealdb`, `surrealdb_mcp`, `wasmtime`, `surrealism_runtime` — all **pure `extern:` cascade** from the roots below

## Genuine roots (confirmed via warm hit/miss + key-diff)

**1. Native `-sys` static libs whose `.a` is byte-different across clones** — the C build bakes the absolute build path into the object, so the content hash itself diverges (distinct from #431's `.rmeta` path cascade, and distinct from the *path-only* false-positives in #470):

| crate | `libfoo.a` content (cold → warm) | warm |
|---|---|---|
| `rquickjs_sys` (quickjs) | `834656493be21f86` → `3dc0f4dbcc36e69f` | miss |
| `tikv_jemalloc_sys` (jemalloc) | `3a69e41d5f6e7b26` → `209e3d107161d91f` | miss |

(Most native libs — `ring`, `zstd_sys`, `lz4_sys`, `bzip2_sys`, `aws_lc_sys`, `cxx` — *are* byte-stable and hit; only quickjs and jemalloc bake the path.)

**2. Absolute build-path env deps kept in the key** (`kept absolute runtime path`):

| field | crates | warm |
|---|---|---|
| `env_dep:DEBUG_OUTPUT_DIR` | `wiggle_macro`, `wasmtime_internal_component_macro` | miss |
| `env_dep:OUT_DIR` | `scratch` (cxx) | miss |

[...]

Issue #421, opened 2026-06-22
fix(cache-key): native -l libs keyed by name only, content change yields stale hit


jleni:
## Summary

The cache key hashes the **name** of each native link library (`-l NAME`) but never its **content**. A native lib whose file content changes in place, while keeping the same `-l` name and the same (normalized) `-L` search path, produces an unchanged key and a stale cache hit.

## Where

`src/cache_key.rs` (the `link_libs` loop, currently around lines 688-697):

```rust
// Native libraries to link (`-l`). Machine-independent names; a
// build script repointing these (e.g. linking a different `libfoo`)
// changes the linked binary without touching RUSTFLAGS. Hashed raw,
// order preserved (static link order is significant).
for lib in &args.link_libs {
    hasher.update(b"link_lib:");
    hasher.update(lib.as_bytes());
    hasher.update(b"\n");
}
```

The comment correctly notes that a build script *repointing* `-l` to a different name changes the output. But it only catches a **name** change. The same name resolving to changed bytes is not covered. By contrast, `--extern` rlibs are content-hashed (around lines 559-587), so this is an asymmetry in the inputs we trust.

## Why it is a real false-hit (not just over-keying)

For `-l static=foo`, rustc bundles `libfoo.a` into the produced rlib / final binary, so its bytes are part of the output. The most plausible trigger is FFI crates using `cc`/`cmake` build scripts:

1. Build script compiles C/C++ into `libfoo.a` in `OUT_DIR` and emits `cargo:rustc-link-lib=static=foo` + `cargo:rustc-link-search=native=$OUT_DIR`.
2. The C source changes. `libfoo.a` content changes. The `-l` name (`foo`) and the normalized search path do not.
3. The Rust source of the linking crate is unchanged, so its cache key is unchanged.
4. kache serves the old rlib/binary built against the previous `libfoo.a`. Silent stale artifact.

This is the cardinal invariant (never serve an artifact when real inputs would produce different output).

## Scope / threat-model caveat

Narrow in practice:

- Only bites when a native lib's **content** changes while its **name and resolved path** stay constant. Many real changes also move the path (new `OUT_DIR` hash) or are caught upstream by cargo's own fingerprinting before kache is consulted.
- System libs (`-l z`, `-l ssl`) changing under an unchanged path is possible on a distro upgrade but uncommon, and resolving `-l NAME` to an actual file means replicating rustc/linker search order (`-L` dirs, sysroot, platform defaults), which is non-trivial and machine-local.

[...]

Issue #324, opened 2026-06-13
fix(cache): v16 key + fingerprint + content-hash hardening


jleni:
## Summary

One coordinated `CACHE_KEY_VERSION` 15→16 bump (`cache_key.rs:106`) closing several latent under-specification gaps. Bundled because they share the `FileFingerprint`/key surface and one invalidation event.

## Changes

- **inode in `FileFingerprint`** — today `{path, size, mtime_ns, ctime_ns}` (`cache_key.rs:1114`) plus the `file_hashes` schema. Without inode, an mtime/ctime-preserving swap memo-hits a stale hash.
- **length-prefix the free-text key fields** (cfg, env-dep, codegen `-Clink-arg`) so adjacent fields can't transpose without changing the digest.
- **rewrite `compute_content_hash`** (`store.rs:220`): it currently folds only sorted blob-hash strings (drops name/size/exec-bit) and truncates to 16 hex, so two entries differing only by a name↔hash transposition or an exec-bit collide → dedup-by-content can keep the wrong survivor. Fold a sorted list of `(relative-name, content-hash, size, exec-bit)`, each length-prefixed; keep 16-hex only as a non-authoritative dedup bucket hint. Add an exec/mode field to `CachedFile` (`store.rs:171`) captured at put time.
- **too-new-input guard** — record an `invocation_start`; after the pre-pass enumerates inputs, if any input's mtime/ctime falls within an FS-granularity margin of start, classify the invocation non-cacheable (serve a lookup hit, **refuse to store**) and emit a counted reason. The margin is **configurable** and derived from filesystem timestamp granularity, not a fixed constant; also gate the in-process hash memo put on a min-age floor.
- **residual-args** — the `_ => {}` catch-all (`args.rs:281`) silently drops unrecognized tokens; they survive only in `all_args` (`args.rs:60`/`:114`) and never enter the digest, so a codegen-affecting flag passed outside the modeled channels is invisible to the key. Fold the **normalized residual** tokens under a versioned `residual_args:` tag (strip the path-bearing classes already handled, canonicalize, sort — never the raw line, which would reintroduce path sensitivity).
- **golden-hash test** pinning an exact `content_hash` for a fixed multi-file entry so the serialization can't silently drift.

## Cost

One fleet-wide cold rebuild (the version bump). Warm-success path otherwise unchanged.

PR #476, opened 2026-06-30
fix(cache-key): build-path-portable hash for GNU static= archives (#471)

emmanuelm41:
Closes the root cause of #471: a `-l static=` archive is bundled into rustc's output so kache folds its content into the cache key (#421), but it hashed the **whole `.a`** — and the `cc` crate names archive members by a hash of the **absolute build path** (`cafca65b…-quickjs.o` vs `4af22b2a…`) while the object bytes are identical. So the key re-keyed per checkout and missed cross-clone (rquickjs-sys, wasm-opt-cc → cascades into `surrealdb_core`/`server`).

## Change

New `native_archive::portable_static_archive_hash`: for a **GNU/SysV `ar`** archive, hash each object member's DATA (length-framed, in archive order) + the symbol-table member (tagged `symtab32`/`symtab64` for `/` vs `/SYM64/`) + the `//` long-name table's **length** — and ignore the `//` content (the cc names) and all header metadata. **Fail-closed:** strict GNU layout (≤1 symtab first, ≤1 `//` next); BSD/macOS (`#1/N`, `__.SYMDEF*`), thin (`!<thin>`), COFF `.lib`, and anything malformed return `None` → whole-file fallback. So it can never produce a *wrong* (colliding) hash, only a less-portable one.

- Scoped to `-l static=` via `FileHasher::hash_static_lib` (preserves the too-new guard #324). `.rlib`s are also ar archives but stay whole-file via `FileHasher::hash`.
- **#421 preserved**: real object-byte changes, member reorders, symtab changes, and `//`-length changes all re-key.
- Portable digest tagged `gnu-ar-v1:` so it can't even textually collide with the whole-file fallback.
- `CACHE_KEY_VERSION` 18 → 19 (one-time clean re-key).

## Safety review (cross-family, codex)

Built design-first and adversarially reviewed across model lineages. The review caught **three real wrong-restore holes** my single-lineage passes missed, all now fixed + regression-tested:
1. ignoring the `//` table made raw GNU symtab offsets ambiguous (a `//`-length change could collide an archive that links differently) → now the `//` length is folded in;
2. COFF `.lib` / `__.SYMDEF_64` were accepted → now strict-layout fallback;
3. `/` vs `/SYM64/` armaps (different word widths, same bytes) shared a tag → now distinct.
Final codex pass: *"No remaining wrong-restore. Ship. Confidence: high."*

## Tests / scope

17 unit tests (portability across member names; `//`-length re-key; #421 content/order/symtab re-key; tag distinctness; BSD/thin/COFF/Darwin64/malformed/misplaced-special fallbacks; `/` vs `/SYM64/`). Full suite 1145 green; clippy + fmt clean.

**GNU only for now** (the Linux CI/S3 target). BSD symtab offset→ordinal normalization is deferred (kept on #471). Efficacy on the real SurrealDB cross-clone bench is being validated in CI now.

Refs #471.

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -126,7 +126,16 @@ use std::path::{Path, PathBuf};
 // whole substrate/crypto stack — kept an absolute build path in its key and
 // re-keyed per checkout, missing cross-clone. Normalizing it changes typenum's
 // (and any such crate's) key bytes, so bump to invalidate v17 entries cleanly.
-pub(crate) const CACHE_KEY_VERSION: u32 = 18;
+//
+// v19 (kunobi-ninja/kache#471): a `-l static=` GNU archive is now folded via a
+// build-path-PORTABLE member-content hash ([`crate::native_archive`]) instead of
+// a whole-file hash. The `cc` crate names archive members by a hash of the
+// absolute build path (`cafca65b…-quickjs.o` vs `4af22b2a…`) while the object
+// bytes are identical, so the whole-file hash re-keyed per checkout and missed
+// cross-clone (rquickjs-sys, wasm-opt-cc, …). The portable hash ignores those
+// names; non-GNU/unparseable archives fall back to the whole-file hash. Either
+// way the static-lib key bytes change, so bump to invalidate v18 entries cleanly.
+pub(crate) const CACHE_KEY_VERSION: u32 = 19;
 const MIN_PERSISTED_HASH_BYTES: i64 = 64 * 1024;
 
 /// Collapse runs of ASCII whitespace into single spaces and trim
@@ -1013,7 +1022,10 @@ fn resolve_native_static_lib(
         }
     }
     let path = found?;
-    let hash = file_hasher.hash(&path).ok()?;
+    // Build-path-portable archive hash (ignores the cc-derived member names that
+    // diverge across clones, #471); falls back to the whole-file hash for
+    // non-GNU / unparseable archives. See `FileHasher::hash_static_lib`.
+    let hash = file_hasher.hash_static_lib(&path).ok()?;
     Some((path, hash))
 }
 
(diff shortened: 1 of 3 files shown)


emmanuelm41:
## Efficacy validated on CI (SurrealDB cross-clone bench, run 28478442747, `--trace-keys`)

The fix does exactly what it targets — the native-`.a` crates are now cross-clone **hits** and have dropped out of the key-diff:

| crate | before | after |
|---|---|---|
| `rquickjs_sys` (libquickjs.a) | miss | **local_hit** |
| `wasm_opt_cxx_sys` / `wasm_opt_sys` | miss | **local_hit** |
| key stability | 97.1% (561/578) | **97.6% (564/578)** |
| warm misses | 18 | **14** (count hit 98.0%, verdict green) |

`link_lib_content` is no longer a diverging field — the GNU member-name-portable hash works on real Linux `cc`/binutils archives as predicted.

### Weighted hit rate is now gated by a different (out-of-scope) root
The remaining key-diff roots are the genuinely runtime-baked env-deps kache keeps absolute to avoid miscaching:
- `env_dep:DEBUG_OUTPUT_DIR` — `wasmtime-internal-component-macro`, `wiggle-macro`
- `env_dep:OUT_DIR` — cxx `scratch`

These cascade via `extern:` into `wasmtime` → `surrealdb_core`/`server`/`surrealdb`/`surrealdb_mcp` (the *expensive* crates), which still miss — so the compile-time-**weighted** rate stays ~47% (the 49.9%↔46.9% move is run-to-run compile-time noise on those few huge misses; count hit-rate and key-stability both improved). Closing that needs an upstream wasmtime/cxx fix (make those paths include-only / path-stable), not a kache change.

**Net:** this PR fully resolves the native-static-archive cross-clone class (benefits any project's cross-machine CI/S3 cache sharing), with correctness cross-family-confirmed. The SurrealDB *weighted* gain is bounded by the upstream-baked env-deps, tracked on #471.

emmanuelm41:
## Regression review (cross-family, codex) — one real regression found & fixed

Ran a codex regression pass over the whole diff (built-in `review --base main` + two focused `task` legs on codebase-wide interactions and cross-platform/perf).

**Confirmed clean (no regression):**
- Windows COFF `.lib` (two `/` linker members) → strict-layout **fallback** to the prior whole-file hash; BSD/macOS → fallback → identical prior hash. No behavior change on those platforms beyond the version bump.
- `CACHE_KEY_VERSION` 18→19 is a shared constant (Rust + cc keys); no `18` hardcoded elsewhere; remote/manifests/shards/planner carry 64-hex keys, not a schema-coupled version → effect is a **one-time global re-key** (incl. the cc cache), no breakage.
- Fallback digest == exact prior `hash_file` value; existing `resolve_native_static_lib` tests unaffected; no downstream consumer chokes on the `gnu-ar-v1:` prefix; `.rlib`/`dylib=` untouched; endian-safe; panic-safe.

**Regression found (leg B) & fixed:** `hash_static_lib` bypassed the persistent file-hash cache and **re-read the whole archive on every build** — where `main`'s cached path returned without reading. For a project linking a large `static=` lib (rocksdb ~100 MB), that's re-reading it on every incremental build.

**Fix:** cache the static-lib digest under a **scheme-namespaced fingerprint key** (`static-ar-v1\0` + path), restoring the warm-build fast path. Codex confirmed the fix: same staleness fields (incl. inode #324), no collision with whole-file rows (NUL-namespaced), too-new guard preserved. New test `hash_static_lib_caches_and_is_namespace_isolated`. Full suite **1146 green**, clippy + fmt clean.

Net: correctness confirmed across three review rounds; the one regression the pass surfaced is fixed and re-verified.

Issue #477, opened 2026-07-01
Very low hit rate for Firefox for daily pull


saschanaz:
## What happened

Hit rate is lower than 10% for daily pull

## Expected behaviour

I'd expect something higher than that?

## Steps to reproduce

1. `./mach build` on the latest build (or https://github.com/mozilla-firefox/firefox/commit/24e52d94)
2. Pull from latest after a day (or https://github.com/mozilla-firefox/firefox/commit/73c9f544)
3. `./mach build` again

## Environment

- `kache --version`: 0.8.0
- OS and version (e.g. macOS 15.4, Ubuntu 24.04): Windows 11
- Rust toolchain (`rustc --version`): 1.96.0
- Install method (mise / cargo-binstall / cargo install): cargo-binstall
- Remote backend (none / S3 / R2 / MinIO / other): none

## Logs

```
```

jleni:
@emmanuelm41 @saschanaz I started a discussion here about firefox optimizations. Maybe we can invite Lars too? I think it would be really nice if we can eventually get something similar to these patches upstream. It would not only help kache but also other tools.

https://github.com/orgs/kunobi-ninja/discussions/53

saschanaz:
But for the same clone the absolute path should not be a problem, right? https://github.com/orgs/kunobi-ninja/discussions/53 seems to only apply for workspaces.

emmanuelm41:
Thanks for the detailed report — it prompted us to build a **permanent daily-pull benchmark** for Firefox so we can track exactly this scenario (build → pull a day of upstream → rebuild, same checkout) on both Linux and Windows.

We ran it against a real ~1-day Firefox delta, and here's what we measured on a from-scratch rebuild (so kache is asked about *every* TU):

| | pull-phase hit (by count) | hit (compile-cost weighted) |
|---|---|---|
| Linux | 79.5% | 25.2% |
| **Windows (clang-cl)** | **79.9%** | **38.5%** |

So on our side the daily pull caches **~80% of translation units by count on both platforms** — we couldn't reproduce the <10%. That doesn't mean your number is wrong; it more likely points at something environment-specific, so it'd be a big help if you could revalidate on your setup:

- `kache doctor` — check the cache index opens and the cache dir is a fast, local, single-machine path (a store that can't persist between your two builds would look like ~all misses on the second one).
- `kache stats` right after the second `./mach build` — is `local` ≈ 0 while `miss` is huge? That's a persistence problem, not temporal invalidation.
[...]

Issue #130, opened 2026-05-23
Speculative cache prefetch on git pull / checkout


jleni:
## Motivation

When a developer pulls new commits or switches branches, the next `cargo build` (or `mach build`, etc.) is a *predictable* cold build: most crates haven't changed, but a few have, and rustc has to discover that one invocation at a time. If a remote cache backend (S3, etc.) holds entries that CI already populated for the new HEAD, those entries are reachable — but only at the moment the build asks for them, mid-compile, serialized behind whatever rustc was doing first.

The Firefox bench surfaced this pattern: gkrust at ~8 min as a single rustc call is fundamentally unaccelerable on a miss, but it's a trivial restore on a hit. The window where remote prefetch can hide latency is between `git pull` and `cargo build` — currently unused.

## Proposed feature

A background prefetcher that, given a source tree:

1. Triggered by an explicit `kache prefetch` (or, optionally, a `post-merge` / `post-checkout` git hook the user installs).
2. Walks the workspace, computes candidate cache keys for each crate the way kache would during a build (dep-info pre-pass + key composition).
3. Fans out `GET` requests to the remote backend for those keys, populating the local store ahead of time.
4. Exits silently if no remote backend is configured.

By the time `cargo build` runs, hits are local — restore latency only, no remote round-trip.

## Implementation sketch

- Reuse `compute_cache_key` + the dep-info pre-pass machinery already in `src/cache_key.rs`.
- Drive crate enumeration via `cargo metadata` (workspace + path deps) plus `cargo build --build-plan` for the per-crate rustc args. (Or invoke `cargo check --dry-run`-style probes.)
- Concurrency-limited fetch loop against the configured remote — bound by a CLI flag (`--jobs`).
- Honor existing local entries — never re-fetch what `Store::contains(key)` says we already have.
- Make it a no-op when:
  - No remote backend is configured
  - `--offline` is set
  - The workspace dep graph fails to resolve (e.g. mid-merge-conflict `Cargo.toml`)

## Open questions

- **Key skew tolerance**: prefetch computes a key based on the *current* environment (rustc version, RUSTFLAGS, target). If the actual build differs, prefetched entries miss. Acceptable for the common case; document the failure mode.
- **Git hook ergonomics vs surprise**: opting users into `post-merge` hooks via `kache prefetch install-hook` is explicit and easy to undo; auto-installing on `kache install` would be surprising. Lean explicit.
[...]

PR #479, opened 2026-07-01
bench(firefox): permanent daily-pull benchmark for #477

emmanuelm41:
## What & why

Issue #477 reports a <10% kache hit rate for the Firefox "daily pull" workflow (build, pull a day of upstream changes, rebuild — same checkout, Windows, local cache). Our existing Firefox benches can't reproduce or track this: they measure the **cross-clone / spatial** axis (same pinned ref at two different paths, on *patched* Firefox with `MOZ_BUILD_DATE` pinned). #477 is the **temporal** axis — same path, source moves forward in time, on *vanilla* Firefox.

This PR adds that missing quadrant as a **permanent, kept-around** benchmark — wired into the nightly bench and dispatchable on demand. ("Temporal" here means the *across-source-time* axis, as opposed to the spatial/cross-clone axis — **not** temporary/throwaway.)

|              | same source                  | changed source            |
|--------------|------------------------------|---------------------------|
| **same path**| `noop`                       | **`pull` ← new**          |
| **diff path**| `warm`/`relocate` (cross-clone) | `relocate-modified`    |

## What it does

- New `Phase::Pull` and a self-contained `run_pull_bench`: clone at ref A → cold build (populate cache) → `git checkout` ref B in the **same** worktree → forced-clean rebuild → report how much of the full TU set kache still hit. No cross-clone `warm`, no `clone-b`. The existing cold/warm path is untouched (purely additive branch, gated on `[source].ref_next`).
- Two **vanilla** scenarios — `bench-firefox-pull` (Linux) + `bench-firefox-pull-windows` — no reproducibility patches, no `MOZ_BUILD_DATE` pin, pinning the issue's **own** reproduction commits (`24e52d94…` → `73c9f544…`), so this doubles as a direct #477 regression.
- **CI wiring:** both scenarios are added to `bench.yml` — a `firefox-pull` Linux matrix arm and a standalone `firefox-pull-windows` self-hosted job — dispatchable and in the nightly schedule. (Also fixes a routing quirk so a windows-only dispatch no longer runs the whole Linux `all` matrix alongside.)
- Gating: `max_errors` + `max_passthrough_pct` blocking; hit rate **warn-only** while we baseline it (ratchet a floor later, as the cross-clone plan did for key stability).

## Measurement nuance (important)

The pull phase forces a full rebuild so kache is asked about **every** TU — the hit rate is *the fraction of the full TU set whose cache key survived the source delta*, a deterministic cache-key-stability number. A real user's post-pull build is **incremental** (mach skips unchanged files) and therefore *faster* than this hit rate implies — this is **not** a speedup claim. Documented in the README and the scenario headers.

## Not in scope

This measures the problem; it does not fix the hit rate. Once baselined (starting with the Windows arm, which matches the report), follow-ups can attack the temporal invalidation drivers (unified-chunk renumbering, reconfigure-regenerated force-included headers, unpinned `MOZ_BUILDID`). Related: #130 (prefetch on pull).

## Testing

- `cargo test -p kache-e2e`: 77/77, including unit coverage for the phase flags, `ref_next` parsing (both the bench and metadata-discovery config structs), two-commit offline checkout, result serialization, and **pull-phase gate binding**.
- `just lint` (clippy `-D warnings`) clean; `just fmt-check` clean.
- The full Firefox e2e is CI/nightly-gated (expensive) and not run locally; the new arms run on the bench runners.

## Notes for reviewers

- A whole-branch review caught a Critical gating bug (the `[checks.assert.pull]`/`[checks.measure.pull]` gates silently didn't bind because `PhaseAssertSpecs`/`PhaseMeasureSpecs` had no `pull` field/`for_phase` arm) — fixed, plus `deny_unknown_fields` hardening so a future unbound phase key fails loudly instead of silently.
- Deferred follow-up (cost, not correctness): `run_cold_phase` snapshots the whole cache after cold on every pull run (a `--retry` artifact the pull path never uses) — worth a `snapshot: bool` param.

Refs #477.


scenarios/bench-firefox-pull/scenario.toml
```diff
@@ -0,0 +1,56 @@
...
+
+# Pull-phase gates. max_errors + max_passthrough_pct are blocking (they catch a
+# broken/uncached build and unclassified driver flags); the hit rate is
+# warn-only for now — we are baselining a number, not yet enforcing a floor.
+[checks.assert.pull]
+max_errors          = 0
+max_passthrough_pct = 40.0
+
+[checks.measure.pull]
+min_hit_rate_pct = 50.0
```
jleni, line 56: You are not applying the patches

https://github.com/kunobi-ninja/kache/blob/ab08707ad11d1d8e6d007c6563292504eee23854/scenarios/bench-firefox/scenario.toml#L106-L125

so performance will not be good

jleni (approved):

PR #491, opened 2026-07-03
feat(remote): add read-only remote consumer mode

### LINKED ISSUE #328: feat(remote): read-only-remote consumer mode
## Problem

A consumer with GET-only remote credentials (fork/PR CI, shared read-only caches) still enqueues uploads — doomed PUTs per entry, log spam, and a risk of tripping the resilience breaker on **expected** AccessDenied.

The existing `local_only` mode (`config.rs:85`, `KACHE_LOCAL_ONLY`) is a different thing: it suppresses **all** remote access. This wants remote **reads**, no **writes**.

## Fix

- A `KACHE_REMOTE_READONLY` config/env gate at the upload-enqueue and upload call sites: read/restore from remote as normal, never write.
- Optional latched auto-detect: classify a 403/AccessDenied on first PUT, log once, and stop further uploads (the flag is primary; auto-detect is a latched fallback so a transient 403 can't permanently mask a real misconfig).
- Classify "denied" distinctly from "flaky" so read-only mode never hides a genuinely broken upload config.

Ysh204:
## Summary

Closes #328

Adds support for a read-only remote consumer mode via `KACHE_REMOTE_READONLY`.

When enabled, kache continues to perform remote cache reads/restores as normal, but suppresses all remote uploads. This is useful for environments with GET-only credentials (e.g. fork/PR CI or shared read-only caches), preventing repeated failed PUT requests and unnecessary upload work.

### Changes

- Added `remote_readonly` configuration support via:
  - `KACHE_REMOTE_READONLY`
  - `[cache] remote_readonly` in the config file
- Added upload guards at both:
  - the upload enqueue path (`handle_upload`)
  - the upload execution path (`do_upload`)
- Preserved existing remote download behavior.
- Updated configuration handling (including the config TUI) to preserve the new option.
- Added tests covering read-only upload behavior.

## Test plan

Verified locally by running:

- `cargo test --all`
- Verified that:
  - uploads are skipped when `remote_readonly` is enabled
  - remote download behavior is unchanged
  - uploads continue to function normally when the option is disabled

## Checklist

- [x] Commit messages follow conventional commits (`feat:`, `fix:`, `docs:`, …)


emmanuelm41:
Thanks for this @Ysh204 — the plumbing is clean and mirrors the existing `local_only` pattern nicely. A few notes before merge:

**1. `kache sync` push phase bypasses the guard (main one).** `upload_entry` (the actual S3 `put_object`) has a second caller besides the daemon: the Push phase in `sync_with_client` (`src/cli.rs:~2786`). It calls `upload_entry` directly with no `remote_readonly` check, so with `KACHE_REMOTE_READONLY=1` and GET-only creds, `kache sync` still attempts PUTs and every one fails — exactly the "repeated failed PUT requests" this is meant to prevent. Suggest short-circuiting the push phase (skip `to_push` / early return) when `config.remote_readonly` is set, so the "suppresses *all* remote uploads" promise holds.

**2. Docs.** `docs/getting-started/configuration.mdx` documents `KACHE_LOCAL_ONLY` / `cache.local_only` in the config table but there's no row for the new `KACHE_REMOTE_READONLY` / `cache.remote_readonly`. Worth adding one mirroring the `local_only` entry (incl. the "env wins over file; explicit `0` overrides" note — which the code already implements correctly).

**3. Field doc comment.** `Config::remote_readonly` (and `CacheFileConfig::remote_readonly`) were added without the doc comment every neighboring field carries. Minor, just for consistency with the rest of the file.

Nothing wrong with the daemon-side guards themselves — they're placed correctly and the env/file precedence matches `local_only`. Mainly #1 to make the feature complete.

---
_Disclosure: this review was LLM-assisted._

Ysh204:
Thanks! Good catch on the sync push path—I missed that upload_entry has a second caller outside the daemon. I'll update sync_with_client to respect remote_readonly, add the missing configuration docs and field comments, and verify whether save_manifest also needs the same guard.

Ysh204:
Thanks for the review! I've pushed a follow-up commit addressing all of the feedback.

### Changes
- **Guarded the push phase** in `sync_with_client` so no remote uploads are attempted when `remote_readonly` is enabled.
- **Guarded `save_manifest`**, since it also performs remote uploads, and now exits early in read-only mode with a `tracing::debug!` message to avoid log spam.
- **Added missing doc comments** for `remote_readonly` in `Config`, `CacheFileConfig`, and `EnvOverrides`.
- **Updated the configuration documentation** by adding `KACHE_REMOTE_READONLY` / `cache.remote_readonly` to `configuration.mdx`.
- **Added behavioral tests** covering the guarded sync push path and skipped manifest upload behavior in read-only mode.

### Verification
- ✅ `cargo fmt --all`
- ✅ `cargo test --all`

emmanuelm41:
The functional changes look complete — all three of my earlier notes are addressed, plus you caught the `save_manifest` path. 

CI's `Check (Linux)` job was failing on a clippy `-D warnings` lint in the new test: `writeln!(file, "{}", event.to_string())` trips `clippy::to_string_in_format_args`. Since maintainer edits are enabled, I pushed a one-line fix (`writeln!(file, "{event}")`) to unblock CI so this can land — hope that's alright. All 3 `remote_readonly` tests pass locally and clippy is clean.

_Disclosure: LLM-assisted._

Issue #493, opened 2026-07-04
Shell Completions


aksheyd:
## Problem

I would like zsh completions personally. Can expand to bash, fsh, etc using `clap` of course.

## Proposed solution

Add clap native completions. Should work out the box given all commands are typed (no variable inputs).

## Alternatives considered

N/A

## Use case

Using kache personally where I can just hit ka -> tab -> (becomes kache) -> space -> mo -> tab -> (becomes monitor). 

---

Seems like a straightforward fix, happy to take this on unless it's already planned or against some rule.

jleni:
Nice contribution!! Keep them coming!! 🚀

PR #494, opened 2026-07-05
feat(cli): add shell completions from clap

aksheyd:
## Summary

I wrote [#493](https://github.com/kunobi-ninja/kache/issues/493) and decided to implement for fun. Happy to close/change/etc the PR however y'all see fit! 

This adds `clap`'s completions to `kache` so users can get shell completions via:
```bash
kache completions <shell>   # bash | zsh | fish | elvish | powershell
```

## Test plan

- [x] Add tests to `tests/cli_commands_test.rs`
- [x] Ran `just check`

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1297,6 +1297,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.6.7"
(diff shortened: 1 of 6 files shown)


jleni:
Thank you @aksheyd !!! keep them coming!! :)

PR #502, opened 2026-07-07
fix(cli): honor CARGO_HOME instead of hardcoding ~/.cargo in init and migrate

### LINKED ISSUE #501: kache init, kache doctor --fix doesn't honor $CARGO_HOME
## What happened

`kache init`, `kache doctor --fix` always use `~/.cargo`, ignoring $CARGO_HOME

## Expected behaviour

If CARGO_HOME is present, use it.

## Steps to reproduce

e.g. `kache init --check` with different $CARGO_HOME

I guess `kache doctor --fix` requires migration from sccache, so it's a little bit harder to reproduce?

## Environment

- `kache --version`:
- OS and version (e.g. macOS 15.4, Ubuntu 24.04):
- Rust toolchain (`rustc --version`):
- Install method (mise / cargo-binstall / cargo install):
- Remote backend (none / S3 / R2 / MinIO / other):

## Logs

```
```

## Additional context

~~Currently I'm working on a fix.~~ Fix implemented. See #502.

CookieHCl:
Closes #501 

## Summary

Honor `CARGO_HOME` instead of hardcoding `~/.cargo`.

### What changed

- In `cli.rs`, these will now use `$CARGO_HOME`, and fall back to `~/.cargo/` if env is missing.
  - kache init (`cargo_config_target_path()`)
  - kache doctor --fix (`migrate()`)
- Also changed README.md, CONTRIBUTING.md that refers `~/.cargo/`.
- Add two tests, enabled one existing test for windows, see **### Test code** below.

### Manual test

Tested `kache init`

image
image

### Test code

- Add two tests, which are nearly similar to existing tests but uses custom `CARGO_HOME`.
  - `doctor_fix_migrates_sccache_config_in_custom_cargo_home()`
  - `init_writes_config_to_custom_cargo_home()`
- **Enabled one existing test for windows**: `init_noninteractive_writes_isolated_cargo_config`
  - Now `kache init` use `CARGO_HOME` first, so this test can be enabled.

> [!NOTE]
> Theoretically, current `doctor_fix_migrates_sccache_cargo_config_to_kache()` works in windows.
> However, this is because current test doesn't checks whether `~/.zshrc` changed.
>
> Since `~/.zshrc` still have the same `dirs::home_dir()` issue, I didn't enabled it and just changed the comments.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ just build
 ### Install locally
 
 ```sh
-just install   # installs to ~/.cargo/bin and registers the daemon service
+just install   # installs to $CARGO_HOME/bin (default ~/.cargo/bin) and registers the daemon service
 ```
 
 > **Note:** The `Justfile` exports `RUSTC_WRAPPER=` to avoid a bootstrapping loop (kache building itself through kache).
--- a/README.md
+++ b/README.md
@@ -91,8 +91,8 @@ winget install kunobi-ninja.kache.Unstable
 ## Quick start
 
 ```sh
-# Interactive setup: configures ~/.cargo/config.toml, installs the
-# background daemon as a login service, and starts it.
+# Interactive setup: configures $CARGO_HOME/config.toml (default ~/.cargo),
+# installs the background daemon as a login service, and starts it.
 kache init
 
 # Or accept all defaults non-interactively:
@@ -102,7 +102,7 @@ kache init -y
 kache doctor
 ```
 
-`kache init` is idempotent — re-run it any time to repair configuration. Use `kache init --check` to preview the changes without touching any files. If you prefer to configure things by hand, just export `RUSTC_WRAPPER=kache` or add it to `~/.cargo/config.toml` under `[build]`.
+`kache init` is idempotent — re-run it any time to repair configuration. Use `kache init --check` to preview the changes without touching any files. If you prefer to configure things by hand, just export `RUSTC_WRAPPER=kache` or add it to `$CARGO_HOME/config.toml` (usually `~/.cargo`) under `[build]`.
 
 ## Use in CI
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2408,7 +2408,12 @@ fn migrate(purge_sccache: bool) -> Result<()> {
             && output.status.success()
         {
             let path = String::from_utf8_lossy(&output.stdout).trim().to_string();
-            if std::path::Path::new(&path).starts_with(cargo_dir.join("bin")) {
+            let sccache_path = std::path::PathBuf::from(&path);
+            let cargo_bin = cargo_dir.join("bin");
+            let resolved_sccache = sccache_path.canonicalize().unwrap_or(sccache_path);
+            let resolved_cargo_bin = cargo_bin.canonicalize().unwrap_or(cargo_bin);
+
+            if resolved_sccache.starts_with(resolved_cargo_bin) {
                 println!("Uninstalling sccache via cargo...");
                 let status = std::process::Command::new("cargo")
                     .args(["uninstall", "sccache"])
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2339,9 +2339,12 @@ fn migrate(purge_sccache: bool) -> Result<()> {
         actions.push("Stopped sccache daemon".into());
     }
 
-    // 2. Replace sccache in ~/.cargo/config.toml
+    // 2. Replace sccache in $CARGO_HOME/config.toml (fallback to ~/.cargo)
+    let cargo_dir = std::env::var_os("CARGO_HOME")
+        .map(std::path::PathBuf::from)
+        .unwrap_or_else(|| home.join(".cargo"));
     for name in ["config.toml", "config"] {
-        let cargo_config = home.join(".cargo").join(name);
+        let cargo_config = cargo_dir.join(name);
         if let Ok(content) = std::fs::read_to_string(&cargo_config)
             && content.contains("sccache")
         {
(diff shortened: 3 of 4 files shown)


emmanuelm41:
## Review — honoring `CARGO_HOME`

**Verdict: looks good, approve with optional polish.** The core change is correct and well-scoped. `cargo_home_dir()` centralizes resolution cleanly, the doctor hints now print the resolved paths, and — importantly — the reworked tests genuinely catch a regression: the old code wrote to `HOME/.cargo`, and the new tests point `CARGO_HOME` elsewhere and assert the config lands there *and* that `~/.cargo` stays untouched, so they'd fail against the old hardcoded path on Unix. Enabling the previously Unix-gated init test on Windows is safe — with `--no-service`, `init` skips `service::install()` and only does a harmless `service_file_path().exists()` read, so no real-profile mutation path remains.

Everything below is **Low / nit** — none blocking. Some are pre-existing, flagged only because the PR touches the adjacent code.

| # | Sev | Finding |
|---|-----|---------|
| 1 | Low | **Fallback path is untested.** The test helper always sets `CARGO_HOME`, and both new tests override it — so "unset `CARGO_HOME` → `~/.cargo`" has no coverage. Such a test would need to be `#[cfg(unix)]` (on Windows `dirs::home_dir()` ignores the test's `HOME` and would touch the real profile). |
| 2 | Low | **`migrate()`'s sccache-uninstall hardcodes `which`** (`cli.rs:2394`) while `cli.rs:1825` already uses `if cfg!(windows) { "where" } else { "which" }`. On native Windows `which` doesn't exist, so uninstall detection is silently skipped — meaning the new `starts_with` Windows-portability improvement never actually runs there. Pre-existing, but directly adjacent to the changed lines. |
| 3 | Low | **`cargo_home_dir()` diverges from cargo's actual semantics on two edges.** Empty `CARGO_HOME=""` → this writes `config.toml` relative to CWD, whereas cargo's `home` crate filters empty and falls back to `~/.cargo`. Relative `CARGO_HOME` → this keeps it relative (file I/O works by luck, but doctor hints/comparisons then show the unresolved path), whereas cargo resolves it against CWD. Trivial to fix by mirroring that logic or depending on the `home` crate that cargo itself uses. |
| 4 | Low | **`starts_with(cargo_dir.join("bin"))` can false-negative under symlinks.** If `which sccache` returns a canonicalized path while `cargo_dir` isn't canonicalized, detection misses. It's strictly better than the old `.contains(".cargo/bin")` for a custom `CARGO_HOME`; the marginal regression risk is only a symlinked default home, only on the `--purge` path. |
| 5 | Nit | **Assertions could be stronger.** The new/enabled tests assert file *existence*, not that it contains `rustc-wrapper = "kache"`, and assert `~/.cargo` *non-existence* rather than seeding sentinel content and asserting it's left unchanged. |
| 6 | Note | **Pre-existing (not introduced here):** `dirs::home_dir().unwrap_or_default().join(".cargo")` writes a CWD-relative `.cargo` if home can't be determined. Carried over verbatim from the old code. |

Nice, focused fix overall — the findings are polish, not corrections.

---
<sub>🤖 This review was assisted by LLMs — cross-checked across two independent model families (Claude + OpenAI codex), with every falsifiable claim verified against the source.</sub>

emmanuelm41 (changes requested):
Thanks for the contribution 🙏 . Just a few minor discoveries that we could tackle on the same PR, and we ship it 🚀

CookieHCl:
Here's my thoughts about your review. Please request changes again if you disagree with this.

## 1. Fallback path is untested.

I don't think this test is necessary. [`rustup` requires](https://rust-lang.github.io/rustup/installation/index.html#choosing-where-to-install) you to set `CARGO_HOME` and `RUSTUP_HOME`.
If user's environment lacks `CARGO_HOME`, it's ill-formed and lots of rustup/cargo related things fails when it's unset.

`~/.cargo` is for "fallback", it is not meant to be a documented behavior or something that should be tested.

## 2. migrate()'s sccache-uninstall hardcodes which.

Addressed in 0b46cc98a530f7d6356b4092ee3e6f711e834673.

## 3. cargo_home_dir() diverges from cargo's actual semantics on two edges.

Addressed in aa1e2b08a31f8f75032a56c8530a3ed12e12af98.

## 4. starts_with(cargo_dir.join("bin")) can false-negative under symlinks.

Addressed in 806318544fbe6bde1db1a5272e1029ef373fd320.

## 5. Assertions could be stronger.

True, but those checks are the same as the existing tests.
I decided to follow the existing tests for simplicity.

## 6. Pre-existing (not introduced here)

I agree with this, but I'm not sure how to fix this. Maybe we should panic with `unwrap()/expect()`? I don't think there is a good default value when `dirs::home_dir()` is `None`.

emmanuelm41:
Thanks for the quick turnaround — the three follow-up commits address findings #2, #3, and #4 cleanly:

- **#2** (`0b46cc98a`) — `where` on Windows / `which` elsewhere, matching the existing pattern at `cli.rs:1825`. ✅
- **#3** (`aa1e2b08a`) — `cargo_home_dir()` now filters empty `CARGO_HOME`, resolves a relative value against CWD, and uses an absolute one as-is. That matches cargo's own `home`-crate semantics. ✅
- **#4** (`806318544`) — both the `which`/`where` result and `$CARGO_HOME/bin` are canonicalized before the `starts_with` check, with an `unwrap_or` fallback when a path doesn't resolve. ✅

One tiny residual (optional, low severity): on Windows `where sccache` can print **multiple** lines (one per match). The code does `.trim()` on the full stdout without `.lines().next()`, so with >1 sccache on `PATH` the resulting `PathBuf` is a two-line string — `canonicalize()` fails, and the uninstall is silently skipped. It fails safe (a miss, never a wrong uninstall), so not blocking; `.lines().next().unwrap_or("")` before building the path would close it.

Findings #1 (a `#[cfg(unix)]` test for the unset-`CARGO_HOME` → `~/.cargo` fallback) and #5 (assert config *content* rather than mere existence) were the optional ones — fine to leave as-is. LGTM.

---
<sub>🤖 Follow-up review assisted by LLMs; changes verified against the source.</sub>

emmanuelm41 (approved):

Issue #287, opened 2026-06-10
Failed to run `cargo clippy`


greenhat616:
## What happened

```
2026-06-10T22:57:36.7937767+08:00 ERROR Flycheck failed to run the following command: CommandHandle { program: "G:\\packages\\cargo\\bin\\cargo.exe", arguments: ["clippy", "--workspace", "--message-format=json-diagnostic-rendered-ansi", "--manifest-path", "G:\\Programs\\Rust\\.ccg\\clash-nyanpasu\\refactor-nyanpasu-config\\backend\\Cargo.toml", "--keep-going", "--all-targets"], current_dir: Some("G:\\Programs\\Rust\\.ccg\\clash-nyanpasu\\refactor-nyanpasu-config\\backend") }, error=Cargo watcher failed, the command produced no valid metadata (exit code: ExitStatus(ExitStatus(101))):
error: process didn't exit successfully: `kache G:\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\bin\clippy-driver.exe rustc -vV` (exit code: 2)
--- stderr
error: unrecognized subcommand 'G:\.rustup\toolchains\nightly-x86_64-pc-windows-msvc\bin\clippy-driver.exe'

Usage: kache [COMMAND]

For more information, try '--help'.
```

## Expected behaviour

`cargo-clippy` run as expected

## Steps to reproduce

1. run `cargo clippy`

## Environment

- `kache --version`: 0.5.0
- OS and version (e.g. macOS 15.4, Ubuntu 24.04): Windows 11 25H2
- Rust toolchain (`rustc --version`): rustc 1.97.0-nightly (f964de49b 2026-05-07)
- Install method (mise / cargo-binstall / cargo install): cargo install
- Remote backend (none / S3 / R2 / MinIO / other): None

## Logs

```
[...]

PR #506, opened 2026-07-10
fix(rustc): detect unrecognized RUSTC_WORKSPACE_WRAPPER drivers by inner rustc

### LINKED ISSUE #505: cargo dylint fails with kache as RUSTC_WRAPPER: unrecognized subcommand
When kache is set as `RUSTC_WRAPPER` and a tool also sets `RUSTC_WORKSPACE_WRAPPER` (e.g. `cargo dylint`, `cargo clippy` with non-clippy-driver wrappers), cargo invokes the wrapper chain as `kache <workspace-wrapper> rustc <args>` for workspace-member crates.

kache only recognizes `rustc`, `rustc*`, and `clippy-driver` as valid compiler basenames in wrapper position (`rustc.rs:71`). Any other workspace wrapper — like dylint's `dylint-driver` — falls through to clap CLI parsing and errors:

```
error: unrecognized subcommand '/Users/…/.dylint_drivers/…/dylint-driver'
```

Same class of bug as #287 (clippy-driver, fixed in #289 by adding the name to the list). The hardcoded approach means each new `RUSTC_WORKSPACE_WRAPPER` tool needs its own patch.

### Reproduce

```toml
[build]
rustc-wrapper = "kache"
```
```sh
cargo install cargo-dylint
cargo dylint --all
```

### Workaround
```sh
CARGO_BUILD_RUSTC_WRAPPER= cargo dylint --all
```

### Environment
- kache 0.9.0, macOS aarch64 (all platforms affected)

aqeelat:
## What

When `RUSTC_WRAPPER=kache` and a tool also sets `RUSTC_WORKSPACE_WRAPPER` (e.g. `cargo dylint`), cargo invokes the wrapper chain as `kache <workspace-wrapper> rustc <args>` for workspace-member crates.

kache's `RustcCompiler::recognizes` only matched `rustc`, `rustc*`, and `clippy-driver` basenames. Any other workspace wrapper — like `dylint-driver` — fell through to clap CLI parsing and errored:

```
error: unrecognized subcommand '/Users/…/.dylint_drivers/…/dylint-driver'
```

Same class of bug as #287 (clippy-driver, fixed in #289 by adding the name to the list).

## Fix

Rather than extending the hardcoded basename list, this PR detects the double-wrapper chain structurally: if `args[0]` is a path (contains a separator) and `args[1]` is recognized as rustc, it's a `RUSTC_WRAPPER + RUSTC_WORKSPACE_WRAPPER` chain — regardless of the workspace wrapper's name.

New helper `is_workspace_wrapper_chain` in `compiler/mod.rs` is wired into two call sites:

1. **`detect_log_mode`** (main.rs) — entry gate: routes the invocation to Wrapper mode instead of falling through to clap.
2. **`run_wrapper_mode`** (main.rs) — adapter fallback: when `detect_compiler` returns None but the chain matches, falls back to the rustc adapter.

The path-separator constraint on `args[0]` prevents false positives on CLI subcommands (cargo always resolves wrappers to absolute paths).

No changes to `recognizes`, `RustcArgs::parse`, `compile.rs`, or `wrapper.rs` — the existing double-wrapper infrastructure (inner-rustc split at args.rs:155, inner-rustc forwarding at compile.rs:44-48) handles the rest unchanged.

## Why caching is safe

The cache key probes `<compiler> -vV` and hashes the version string (cache_key.rs:360). Different workspace wrappers (dylint-driver, clippy-driver, plain rustc) produce different `-vV` output, so cache keys are naturally differentiated.

## Testing

- Unit tests for `is_workspace_wrapper_chain` (positive: dylint-driver path, Windows backslash path; negative: no path separator, inner arg not rustc, too few args)
- `detect_log_mode` test: dylint-driver chain routes to Wrapper; negative case (CLI subcommand with `rustc` in position 2 but no path separator) stays Cli
- `RustcArgs::parse` double-wrapper test with dylint-driver path

All tests are pure string-logic — run on every host OS, no fixtures.

Closes #505

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1445,7 +1445,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e75b2483e97a5a7da73ac68a05b629f9c53cff58d8ed1c77866079e18b00dba5"
 dependencies = [
  "digest 0.10.7",
- "spin 0.10.0",
+ "spin 0.10.1",
 ]
 
 [[package]]
@@ -2692,7 +2692,7 @@ dependencies = [
  "atomic-polyfill",
  "hash32 0.2.1",
  "rustc_version",
- "spin 0.9.8",
+ "spin 0.9.9",
  "stable_deref_trait",
 ]
 
@@ -3612,7 +3612,7 @@ version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bbd2bcb4c963f2ddae06a2efc7e9f3591312473c50c6685e1f298068316e66fe"
 dependencies = [
- "spin 0.9.8",
+ "spin 0.9.9",
 ]
 
 [[package]]
@@ -6105,18 +6105,18 @@ dependencies = [
 
 [[package]]
 name = "spin"
-version = "0.9.8"
+version = "0.9.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6980e8d7511241f8acf4aebddbb1ff938df5eebe98691418c4468d0b72a96a67"
+checksum = "3763264f6b73151db08c50ff20d7d8a0b8796e021cdea7ceedad07b80155fa0e"
 dependencies = [
  "lock_api",
 ]
 
 [[package]]
 name = "spin"
-version = "0.10.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d5fe4ccb98d9c292d56fec89a5e07da7fc4cf0dc11e156b41793132775d3e591"
+checksum = "023a211cb3138dbc438680b32560ad89f699977624c9f8dbb95a47d5b4c07dd3"
 
 [[package]]
 name = "spki"
--- a/crates/kache-e2e/src/runner.rs
+++ b/crates/kache-e2e/src/runner.rs
@@ -1074,7 +1074,7 @@ fn inspect_binary(run_cmd: &str, cwd: &Path, forbidden: &[String]) -> Option<Str
                 "binary contains forbidden substring `{}` in {}: ...{}...",
                 needle,
                 bin_path.display(),
-                &haystack[start..end].replace('\n', " ")
+                haystack[start..end].replace('\n', " ")
             ));
         }
     }
--- a/src/args.rs
+++ b/src/args.rs
@@ -984,6 +984,39 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_double_wrapper_unrecognized_driver() {
+        // Issue #505: dylint-driver (or any future RUSTC_WORKSPACE_WRAPPER
+        // tool) in the double-wrapper chain. The split keys off the inner
+        // rustc, so an unrecognized workspace wrapper is forwarded correctly.
+        let args: Vec<String> = vec![
+            "/Users/dev/.dylint_drivers/nightly/dylint-driver",
+            "/home/user/.rustup/toolchains/stable/bin/rustc",
+            "--crate-name",
+            "foo",
+            "src/lib.rs",
+            "--crate-type",
+            "lib",
+        ]
+        .into_iter()
+        .map(String::from)
+        .collect();
+
+        let parsed = RustcArgs::parse(&args).unwrap();
+        assert_eq!(
+            parsed.rustc,
+            PathBuf::from("/Users/dev/.dylint_drivers/nightly/dylint-driver")
+        );
+        assert_eq!(
+            parsed.inner_rustc,
+            Some(PathBuf::from(
+                "/home/user/.rustup/toolchains/stable/bin/rustc"
+            ))
+        );
+        assert_eq!(parsed.crate_name.as_deref(), Some("foo"));
+        assert!(!parsed.all_args.iter().any(|a| a.contains("rustc")));
+    }
+
+    #[test]
     fn test_parse_single_wrapper_unchanged() {
         // Normal case: kache /path/to/rustc --crate-name foo src/lib.rs
         // After main.rs strips argv[0], parse receives: [/path/to/rustc, ...]
--- a/src/compiler/mod.rs
+++ b/src/compiler/mod.rs
@@ -586,6 +586,18 @@ pub fn detect_compiler(args: &[String]) -> Option<&'static CompilerAdapter> {
         .find(|adapter| adapter.recognizes(args))
 }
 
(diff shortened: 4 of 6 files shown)


emmanuelm41 (commented):
Thanks @aqeelat for your contribution. I suggested a few changes to make this wrapper feature more resilient. 🙏

emmanuelm41 (commented):
Detection approach is right — recognizing the chain by the inner rustc beats a basename allowlist. Two problems below it, though.

**1. The `-vV` safety argument is false for dylint.** It holds for clippy (`clippy 0.1.96 …` vs `rustc 1.96.0 …` — I checked), but `dylint-driver` forwards `--version --verbose` to its embedded rustc and prints *rustc's* banner ([source](https://docs.rs/dylint_driver/latest/src/dylint_driver/lib.rs.html#399-406)). Dylint builds are still invalidated correctly, just for an unrelated reason — it registers `DYLINT_LIBS` and its lint libraries as dep-info inputs, which do reach the key.

So the key identifies the compiler only by `get_rustc_version(&args.rustc)` (`cache_key.rs:360`) — self-reported, exit status unchecked, and neither the wrapper's bytes nor `inner_rustc` are folded in. That was fine while the set was a hand-vetted allowlist; this PR opens it to any path-shaped `argv[0]` followed by a rustc-ish name, where nothing requires a distinct banner.

**Suggestion:** fold the wrapper's identity (path + mtime/size, or a content hash) and the inner rustc's `-vV` into the key. That's what cargo's own rustc fingerprint does — it hashes the wrapper executables' resolved paths and metadata, not their version output. It closes the collision class for every wrapper, keeps this PR's cache benefit for dylint, and makes clippy's key sounder too. If that's more than you want here, ship detection + uncached passthrough for unknown wrappers and do the key work as a follow-up — #505 is a hard error, so passthrough alone fixes it.

**2. The path-separator guard has a false negative.** "Cargo always resolves wrappers to absolute paths" isn't true — a separator-free `RUSTC_WORKSPACE_WRAPPER` is passed through bare and resolved via `PATH`. Verified with a logging wrapper: cargo hands `argv[1]=mydriver`, no separator, so it still falls through to clap with the #505 error. `cargo dylint` sets an absolute path (so the reported case is fixed), but the bare-name config isn't. Consider also matching `args[0]` against `env::var_os("RUSTC_WORKSPACE_WRAPPER")`.

**Minor:** the three probes (`get_rustc_version`, sysroot, `run_dep_info_pass` at `cache_key.rs:520`) don't forward `inner_rustc`, so "any future workspace wrapper works" isn't quite true yet — it fails safe via passthrough, but a strict wrapper won't cache.

<sub>Disclosure: LLM-assisted review; the version banners and the cargo bare-wrapper behavior above were verified by running them.</sub>

emmanuelm41 (changes requested):
Left a few suggestion to address! But thanks a lot for the interest 🙏

aqeelat:
Thanks for the thorough review. Pushed `75f01b6` addressing all three points.

**1. Cache key — went with your passthrough suggestion.** Switched from the rustc adapter fallback to `run_compiler_directly` (uncached exec). Unrecognized workspace wrappers now compile correctly without cache-collision risk. The key work can be a follow-up if caching lint tools is ever warranted, but I think it's the wrong default — dylint/clippy emit lint output on stderr, which kache doesn't replay on cache hits, so caching would silently suppress warnings.

**2. Bare-name wrapper — fixed.** `is_workspace_wrapper_chain` now also checks `args[0]` against `env::var_os("RUSTC_WORKSPACE_WRAPPER")`, so bare names resolved via PATH are caught. Verified with the dylint-driver binary: installed kache 0.9.0 errors with "unrecognized subcommand", the patched build passes through and compiles cleanly.

**3. Probe forwarding — moot under passthrough.** The probes only run in the cached path (`wrapper::run` → `compute_cache_key`). Unrecognized wrappers take `run_compiler_directly` and never touch them. This becomes relevant only if caching is enabled for unknown wrappers in the future.

emmanuelm41:
Amazing! Thanks for the addressed changes. Just small things now, and let's get this merged 🙏 

Revision looks good — all three points addressed. I pulled `75f01b6` and ran the suite locally: 1176 passed, including the three new `workspace_wrapper_chain_*` tests. The passthrough correctly preserves the chain (`run_compiler_directly` execs `args[0]` with `args[1..]`, which still carries the inner rustc), and you're right that probe forwarding is moot now that unknown wrappers never reach `compute_cache_key`.

One correction, since it affects the follow-up rather than this PR:

> dylint/clippy emit lint output on stderr, which kache doesn't replay on cache hits, so caching would silently suppress warnings

kache does replay cached stderr — on every hit path (`wrapper.rs:410-414`, `973-977`, and the remote hits at `1045-1049` and `1122-1125`). clippy-driver is already in the allowlist, so `cargo clippy` is cached today and its warnings survive exactly because of that replay. Same conclusion (don't cache unknown wrappers), but the hazard is the inverse of what's described: caching a lint tool replays **stale** lint output, it doesn't suppress it. Flagging it so nobody reads that comment later and concludes clippy caching is broken.

Nit on `workspace_wrapper_chain_detects_bare_name_via_env`: the manual `match prev` restore doesn't run if the `assert!` panics, leaking `RUSTC_WORKSPACE_WRAPPER=mydriver` into the rest of the test process. There's an RAII guard for this already — `TestEnvGuard` at `src/wrapper.rs:2088`.

Non-blocking: nothing currently pins the new behavior. All the tests are string classification, so none of them would fail if the chain quietly went back to being cached. A fake-wrapper integration test asserting it gets executed on every build would cover that.

aqeelat:
Pushed `d2b071f` addressing the two nits.

**Env-var guard.** Replaced the manual save/restore with an RAII guard (same pattern as `TestEnvGuard`), so the env var is cleaned up even if the assert panics.

**Integration test.** Added `workspace_wrapper_passthrough_executes_every_time`: builds twice through a fake workspace wrapper that records each execution to a marker file. Asserts the wrapper ran both times — would fail (1 run instead of 2) if unknown wrappers were ever routed back through the cache.

**Correction on my earlier stderr comment.** You're right — kache does replay cached stderr (wrapper.rs:410-414, etc.). The hazard I should have described is stale lint replay, not suppression. Same conclusion (uncached passthrough for unknown wrappers), but thanks for catching the factual error.

emmanuelm41:
@aqeelat Ci is failing, could you please fix it?

aqeelat:
@emmanuelm41 the issue was because of the upgrade to rust 1.97, which added a clippy lint that flags `&` on an already-owned value inside `format!`

Fixed.

aqeelat:
@emmanuelm41 
~~actually there is another issue as well in deny. Let me investigate.~~
all good.

emmanuelm41 (approved):

Issue #497, opened 2026-07-07
kache store directory grows twice over the default-configured max store size?


saschanaz:
## What happened

```
> du store
╭───┬──────────────────────┬──────────┬──────────╮
│ # │         path         │ apparent │ physical │
├───┼──────────────────────┼──────────┼──────────┤
│ 0 │ D:\kache-cache\store │ 113.2 GB │ 113.2 GB │
╰───┴──────────────────────┴──────────┴──────────╯
```

```
 kache config
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 [General]
 > Cache directory     D:\kache-cache
   Max store size      (default: 50GiB)
   Disabled            [ ]  (env: KACHE_DISABLED)
```

## Expected behaviour

It should not go above the max store size?

## Steps to reproduce

1. Do some builds with kache
2. Check the directory size

## Environment

- `kache --version`: 0.9.0-rc.1
- OS and version (e.g. macOS 15.4, Ubuntu 24.04): Windows 11
- Rust toolchain (`rustc --version`): 1.96.1
- Install method (mise / cargo-binstall / cargo install): cargo binstall
- Remote backend (none / S3 / R2 / MinIO / other): none

## Logs

```
```

jleni:
@Ysh204 do you want to take this? or you think it is unrelated?

Ysh204:
Thanks for the report! I'll take a look.

Since this might be related to #329, I'll investigate whether the store growth is caused by unreclaimed worktree artifacts or by a separate issue with max store size enforcement, and update this issue with what I find.

emmanuelm41:
@Ysh204 I have experienced the same thing locally on my machine, and I do cleanup worktrees all the time, so it may be the case that it is related to it.

Ysh204:
### Findings

- Store-wide eviction is currently triggered only from daemon-owned paths:
  - the periodic GC task, and
  - the post-upload eviction check after a successful remote upload.
- Local cache writes through the `RUSTC_WRAPPER` hot path do not perform any store-size checks or trigger eviction.
- `sweep_orphan_blobs()` only removes blobs that exist on disk but are no longer referenced by the SQLite `blobs` table. It does not reclaim entries belonging to deleted Git worktrees because those entries and blob references still exist in the database.
- This means #329 (reclaiming orphaned worktree cache entries) would help reduce persistent store growth from deleted worktrees, but it doesn't appear to be the only factor involved here.

[...]

PR #507, opened 2026-07-11
feat(store): reclaim orphaned worktree cache entries during GC

### LINKED ISSUE #329: feat(store): reclaim orphaned-worktree cached bytes (config-driven)
## Problem

Many git worktrees of the same repo each leave a `target/` whose artifacts are already content-addressed in the kache store. When a worktree is pruned, its cached bytes are not reclaimed automatically, so the store grows across throwaway worktrees.

## Fix

A config-driven reclaim during kache's own GC / `kache clean`: detect cached entries/links belonging to worktrees that no longer exist and release them, bounded by config (which roots to manage, age/grace window, on/off).

Stays entirely within kache's existing store + eviction machinery (`sweep_orphan_blobs`, `store.rs:1051`); immutable-artifact scope only; no separate orchestrator process.

**Out of scope (future, not committed):** generalizing the store to dedup non-Rust artifacts.

Ysh204:
## Summary

Closes #329.

This PR adds config-driven reclamation of cache entries belonging to deleted Git worktrees during `kache gc` and `kache clean`.

The implementation integrates with the existing GC/eviction pipeline and reuses the existing eviction machinery (`remove_entry_guarded`) so blob reference counting, metadata cleanup, and active-restore safety remain consistent with the rest of the store.

### Configuration

Added optional configuration for worktree reclamation:

- `cache.reclaim_orphaned_worktrees`
- `cache.worktree_reclaim_roots`
- `cache.worktree_reclaim_grace_secs`

Environment variable equivalents:

- `KACHE_RECLAIM_ORPHANED_WORKTREES`
- `KACHE_WORKTREE_RECLAIM_ROOTS`
- `KACHE_WORKTREE_RECLAIM_GRACE_SECS`

The feature is disabled by default.

### Store

- Record the producing Git worktree for newly created cache entries.
- Reclaim entries whose recorded worktree no longer exists once they exceed the configured grace period.
- Reuse the existing eviction path (`remove_entry_guarded`) so blob refcounts and cleanup behavior remain unchanged.
- Clean up worktree metadata when entries are removed.

### GC integration

Reclamation now runs before normal age/size-based eviction in:

- daemon background GC
- `kache gc`
- `kache clean`

This allows reclaimable worktree entries to reduce cache size before LRU eviction considers removing active entries.

## Design note

This implementation records the **producing worktree** for each cache entry rather than tracking every worktree that subsequently restores it.

This intentionally keeps cache hits read-only and avoids introducing additional bookkeeping on the hot path.

The trade-off is that if the producing worktree is deleted while another worktree is still reusing the entry, the entry may eventually be reclaimed after the configured grace period, resulting in a cache miss and recompilation rather than incorrect artifacts.

I'd appreciate feedback on whether this trade-off matches the intended design for #329, given the earlier discussion around "deleting != dereferencing".

## Test plan

Verified locally with:

```bash
cargo fmt --all
cargo clippy --workspace -- -D warnings
cargo test --all
```

Added coverage for:

- reclaiming entries from deleted worktrees
- respecting configured reclaim roots
- grace period behavior
- lifecycle trade-off:
  - Worktree A produces an entry
  - Worktree B restores the entry
  - Worktree A is deleted
  - GC respects the grace period
  - entry is reclaimed after expiry
  - Worktree B recompiles and becomes the new producing worktree

## Notes

This PR addresses the worktree reclamation described in #329.

The separate question of when size-based eviction should be triggered in daemon-less/local-only workflows (discussed in #497) is intentionally left out of scope.


emmanuelm41:
Hi @Ysh204 — thank you for this. You asked the right design question twice *before* coding, and the implementation is clean (correct `remove_entry_guarded` reuse, additive nullable migration, real tests).

I hit something while reviewing that I want to walk through — but **I'm not certain about it**, and you've been in this code more recently than me. If I've got it wrong, please say so.

## What I think I'm seeing

Two worktrees on two feature branches. Each touches only `feature_lib`; both share `core_lib` unchanged:

```
crate_name   origin
-----------  ------
core_lib     wt1     <- wt2 HIT this, didn't create its own
feature_lib  wt1
feature_lib  wt2
```

Deleted `wt1` (merged, thrown away), ran `gc` with your feature on. It reclaimed **two** entries: `feature_lib@feat-1` (correct — genuinely dead), and `core_lib` — which `wt2` was still using. `wt2` then recompiled a crate it never touched.

My reading: kache's keys are deliberately path-independent (`PathNormalizer` → stable sentinel), so worktrees *share* entries rather than each owning a copy. If that's right, "which worktree produced this?" can't tell a dead entry from a shared one — and in a real workspace, where a branch touches a few crates and shares the rest, the shared ones are most of them.

That's not a bug in your code. Your code does what it says.

## Where I might be wrong

- My repro is a **toy workspace** — two crates. If you tested something realistic and saw meaningful reclamation, that's evidence against me and I'd like to see it.
- I may be **misreading the key normalization**, or missing a case where worktrees don't share keys.
- Nobody ever asked @rirze whether his 250 GB was the **store** or accumulated **`target/` dirs**. Different fixes entirely. That's a hole in *my* reasoning, not yours.

Easy to check: two worktrees on different branches sharing an untouched crate → build both → delete the first → `gc`. If the second still hits `core_lib`, I'm wrong.

## Two things worth flagging either way

1. **`clean`'s reclaim path doesn't take `gc.lock`** (`cli.rs:1638`), which `gc` and the daemon both do. `remove_entry_guarded` reads the blob list *before* its transaction and never re-checks the row inside it, so two concurrent removers can double-decrement a shared blob. Filed as #510 — it's latent on `main` regardless of this PR.
2. **`kache config` would silently drop your three settings** — `config_tui.rs` needs an explicit `preserved_*` param per field; `..Default::default()` writes them back as `None`. Undocumented trap, I only caught it by accident.

Also: `origin_worktree` correctly never reaches `meta.json` or the remote packs — no local paths leak to S3. Real risk, and you avoided it.

## Where this leaves us

If my reading holds, I don't think worktree-origin is the right foundation — and I'd rather say so now than have you polish something we'd back out later. But I'm holding that loosely.

Meanwhile: this sent me back to **#497**, where you'd *already* diagnosed the root cause before I did. I've posted a repro confirming it. That's the one I'd love to have you on — I think it's the fix that actually helps @rirze.

Tell me where I've got this wrong.

emmanuelm41 (changes requested):
Holding the merge button while we work through the design question — a hold, not a judgement. The work is good.

Substance (and the parts I'm unsure about) are in my comment. Short version: I think the producing-worktree model can't tell a dead entry from one another worktree is still sharing, because kache's keys look to be deliberately path-independent. I reproduced a merged worktree's deletion evicting a crate a live worktree was using — but on a toy workspace, so if you've seen otherwise on something real, that matters.

Independent of all that: `reclaim_deleted_worktrees_for_clean` (`cli.rs:1638`) runs bulk removal without `gc.lock`, which `gc` and the daemon both hold. Filed as #510 — latent on `main` already.

@Ysh204 — push back if I've read this wrong.

Ysh204:
Thanks for the detailed review and for taking the time to reproduce the scenario.

I think the `core_lib` example makes the limitation of the producer-origin model much clearer than my earlier testing did. My implementation was validating that model, but your repro shows that "producing worktree" isn't a reliable proxy for "no longer referenced" because identical worktrees intentionally converge on the same cache entry.

I'm glad we went through this exercise though ,the discussion helped narrow down the actual issue behind the reported disk growth, and it was great to see the investigation around #497 hold up with a concrete reproduction.

Thanks as well for pointing out the `gc.lock` and `config_tui` issues. I'll keep those conventions in mind for future store work.

I'll shift my focus to #497 and work on a fix there instead.

emmanuelm41:
@Ysh204 should we close this one right?

Issue #510, opened 2026-07-13
fix(store): `remove_entry_guarded` can double-decrement shared blob refcounts (safe only by `gc.lock` convention)


emmanuelm41:
## Summary

`remove_entry_guarded` (`src/store.rs`, ~L1907) reads the entry's blob-hash list from `meta.json` **before** opening its write transaction, and never re-checks inside the transaction that the entry row still exists. It is safe today **only** because every bulk remover happens to hold `gc.lock`. That's an undocumented invariant, and the next caller who misses it corrupts the store.

## The race

Two removers, A and B, targeting the same entry:

1. A and B both read `meta.json` and get the same list of blob hashes. *(outside any transaction)*
2. A opens its transaction, decrements each blob's refcount, deletes the entry row, commits.
3. B opens its transaction. The only in-transaction guard is the recency check:
   ```sql
   SELECT EXISTS(SELECT 1 FROM entries WHERE cache_key = ?1 AND last_accessed >= datetime('now', ?2))
   ```
   The row is **gone**, so this returns 0 — meaning "not recently accessed" — and B proceeds.
4. B decrements every blob **a second time**.
5. A blob shared with another live entry goes 2 → 1 → 0, gets `DELETE`d from `blobs`, and is `unlink`ed from disk.

Result: a still-valid entry now points at a missing blob file. Blobs are shared by design (that's what the content-addressed store and `blobs.refcount` are *for*), so this is not an exotic case.

The recency guard cannot catch it: absence of a row is indistinguishable from "not recently accessed" in that query.

## Why it hasn't bitten us

Every current bulk remover takes `gc.lock` first — `kache gc` (`cli.rs:1223`, with the explicit #326 comment) and the daemon (`daemon.rs:2520`). The lock is doing all the work; the function itself is not self-protecting.

PR #507 added a new bulk remover (`reclaim_deleted_worktrees_for_clean`, `cli.rs:1638`) that *didn't* take the lock — which is how this surfaced. That PR isn't merging, but the underlying fragility is on `main` today.

## Suggested fix

Re-check row existence inside the transaction, before any decrement:

```rust
let row_exists: i64 = tx.query_row(
    "SELECT EXISTS(SELECT 1 FROM entries WHERE cache_key = ?1)",
    params![cache_key],
    |row| row.get(0),
)?;
if row_exists == 0 {
    return Ok(false);  // someone else already released this entry's references
}
```

[...]

PR #511, opened 2026-07-14
feat(gc): trigger background GC opportunistically for local workflows

### LINKED ISSUE #497: kache store directory grows twice over the default-configured max store size?

Ysh204:
## Summary

Closes #497.

This PR ensures that the configured `max_store_size` is enforced for local-only, daemonless workflows.

Previously, store-wide eviction (`store.evict()`) was only triggered from daemon-owned paths (the background GC loop and the post-upload hook). As a result, local cache writes could continue growing the store indefinitely when no daemon was running.

To address this, the wrapper now opportunistically triggers a background `kache gc` after successful cache writes when the store exceeds the configured size budget. The trigger is throttled to avoid repeated checks and runs asynchronously, preserving the fast cache-hit path while reusing the existing GC and eviction machinery.

No changes are made to the eviction algorithm itself—this PR only changes **when** it is triggered for local-only workflows.

## Test plan

Verified locally by running:

```bash
cargo fmt --all
cargo clippy --workspace -- -D warnings
cargo test --all
```

Additionally verified that:

- Local-only, daemonless builds now trigger background GC once the configured store size is exceeded.
- The auto-GC trigger is throttled to avoid repeated background launches.
- Cache-hit performance remains unchanged (no additional work is performed on cache hits).
- Existing GC and eviction behavior is preserved by continuing to use the existing `store.evict()` implementation.

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/) (`fix:`, …)


emmanuelm41:
Thanks @Ysh204 — right shape. Follows @jleni's steer, sane hysteresis, and you wired `preserved_auto_gc` through `config_tui.rs` without being asked. Nice.

I tested it against the #497 repro though, and I don't think it bites yet. Tell me if I set it up wrong.

**A normal build never triggers GC.** Fresh 20 KiB-capped store, one `cargo build`:

```
auto-gc log lines: (none)
Store: 189.5 KiB / 20.0 KiB (948%)
```

I think it's the ordering in `auto_gc_wanted` (`wrapper.rs:186`) — the stamp is written *before* the size check. So the first put, while still under budget, claims the stamp and burns the 5-min throttle. Every later put in that build skips. Build ends 9× over, nothing retries.

**And when it does trigger, it evicts nothing:**

```
auto-gc: store size 194088 exceeds max 20480, triggering background GC
auto-gc: spawned background `kache gc`
   -> evicted 0 entries.     Store: 948%
```

Detection and spawn are perfect. But `EVICTION_IDLE_GRACE` (120s) pins every entry the build just wrote, so the gc walks the store and evicts nothing. Age those same entries → same gc evicts 23 → 76%. Machinery's fine, it's the timing. And the triggering check already re-stamped, so no retry for 5 min.

**Two small fixes:**
1. Check the size *first*, stamp *second* — kills the no-op.
2. Have the worker retry after the grace instead of one-shotting: sweep, sleep ~120s, sweep again.

**Smaller stuff:**
- `kache gc` auto-starts a daemon (`send_gc_request`'s error arm → `start_daemon_background()`), so each trigger forks two daemon attempts. Mine die instantly and it falls back to local gc — but where the daemon *does* start, this quietly makes daemonless users daemon users. Probably cleaner to evict locally under `gc.lock` than to shell out to `kache gc`.
- Child isn't detached (no `setsid`), so Ctrl-C on a build also SIGINTs the gc — and `remove_entry_guarded` unlinks blobs before committing (#510).
- `cargo fmt --all -- --check` fails (stray blank lines), and `auto_gc`/`KACHE_AUTO_GC` aren't in the docs — both checklist boxes are ticked.
- Auto-gc fires even when the `put` errored (it's outside the success arm).
- The tests check the *decision* but never that anything gets evicted — which is why they're green while it no-ops. An end-to-end "build past the cap → assert it comes back under" is the one I'd most want.

**My setup, so you can check me:** `local_max_size = "20KiB"`, `KACHE_LOCAL_ONLY=1`, no daemon, ~25 tiny crates, fresh store, one build, then `kache stats`. If it bounds the store for you, I've got something wrong.

Good direction — both fixes are small. Ping me when you push.

Ysh204:
Hi @emmanuelm41, thanks again for the detailed review.

I've pushed an update addressing the points you raised:

- Check size before updating the throttle stamp.
- Only trigger auto-GC after successful `put()` operations.
- Background worker now performs a second local GC pass after `EVICTION_IDLE_GRACE`.
- Background execution now bypasses daemon startup and runs local GC directly.
- Detached the background worker so it isn't interrupted by `Ctrl-C` from the parent build.
- Fixed the `remove_entry_guarded` transaction race described in #510.
- Added an end-to-end integration test covering the full flow (build past the configured cap → background GC → store returns under the configured limit).

emmanuelm41:
@Ysh204 the changes look correct, but so far CI is failing. Can you fix it?

Ysh204:
@emmanuelm41 
Updates

I've pushed another round of changes addressing the review feedback:

### SQLite concurrency fix (#510)

- Reworked `remove_entry_guarded` to perform the `DELETE` inside the transaction before blob refcount updates, using the affected row count to determine whether cleanup should proceed.
- This prevents concurrent removers from both decrementing shared blob refcounts while preserving cleanup of partial/untracked entries that exist on disk without a corresponding database row.

### Config initialization

- Added the new `auto_gc` field to all `Config` struct initializers across the unit and integration test suites (`cli.rs`, `daemon.rs`, `fallback_planner.rs`, `remote_layout.rs`, `remote_plan.rs`, `report.rs`, `store.rs`, and `tui.rs`).

### Dependency audit

- Updated the `spin` dependency from `0.10.0` to `0.10.1` in `Cargo.lock` to resolve the `cargo deny` audit warning.

### Integration test improvements

- Made the background worker retry delay configurable via `KACHE_AUTO_GC_RETRY_DELAY_SECS` (default remains `120s`).
- Updated `test_auto_gc_bounds_store_size` to:
  - enable executable caching,
  - use a shorter retry delay for testing,
  - correct the database path (`index.db`),
  - age entries within the test environment,
  - and verify the complete background auto-GC flow in a few seconds.

Everything is now passing locally, including `cargo check --all-targets`, `cargo test`, and the new end-to-end auto-GC integration test.

emmanuelm41:
@Ysh204 amazing! I think CI will be green now! I just caught two small more things to tackle, and we are done. Sorry for the long iteration, I know it can be tedious. But happy to have you contributing to the project.  1 and 2 are quick and worth doing before merge. Ping me when it's up and I'll re-run.



**1. Retry sleep is one second short.** The worker sleeps exactly `EVICTION_IDLE_GRACE` (default 120s, `cli.rs:1227`), but the active-pin check is `>=` at SQLite's 1-second resolution — so the very entry that *triggered* the GC can still be pinned when the second sweep runs, and survive both. Your e2e test doesn't catch it because it backdates entries by 300s. Bumping the default to `EVICTION_IDLE_GRACE + 1` fixes it.

**2. The under-budget path SUMs on every miss.** When you moved the size check before the stamp (good change), it meant the stamp is now only written *once you're over budget*. So while the store is under budget, every cache miss runs a full `SELECT SUM(size)` — O(entries) on the hot path, growing with cache size. That's also the case the `config.rs` doc comment says *doesn't* happen ("one stat per compile … one SUM per interval"). Simplest fix: also refresh the stamp on the under-budget branch so the SUM is genuinely throttled to once per interval.

Ysh204:
Hi @emmanuelm41,

I've pushed the updates addressing the latest points and resolved the merge conflicts with main. All unit and integration test suites compile and pass successfully on both Linux and Windows!

"Additionally, I noticed and patched a concurrency race/truncation risk in the log rotation logic (src/events.rs) where parallel wrapper processes could cause silent event loss. I added a regression test for it as well."

Ysh204:
ill just create a different pr for that #518

emmanuelm41 (approved):

Issue #485, opened 2026-07-02
debuginfo: make default remap targets resolvable by profilers (follow-up to #480)


jleni:
## Problem

#484 adds `KACHE_RUSTC_PATH_NORMALIZE=0` as the escape hatch for #480: with remapping off, samply and the Firefox Profiler see real paths and everything resolves. But the opt-out gives up cross-worktree and cross-machine hits, which is the whole point of remapping. The default should be good enough that the knob stays a last resort.

Today the remap targets are invented sentinels (`<WORKSPACE>`, `<RUSTUP_HOME>`, `<CC_ROOT>`, ...). Two problems:

1. No tool resolves them without a manually configured source map. The docs cover gdb/lldb (#239, #243); samply has no equivalent setting at all.
2. They are relative-looking strings, so DWARF consumers join file entries onto `DW_AT_comp_dir` and produce composed garbage like `<WORKSPACE>/<RUSTUP_HOME>/toolchains/.../core/src/ptr/mod.rs` (the exact log in #480).

## Proposal

Keep remapping on. Change only the target spellings to forms existing tooling already resolves, and make them all absolute.

rustc:

| prefix | today | proposed |
|---|---|---|
| `{sysroot}/lib/rustlib/src/rust` (new rule) | falls under `<RUSTUP_HOME>` | `/rustc/<commit-hash>` |
| workspace / cwd | `<WORKSPACE>` | `/proc/self/cwd` on Linux, `/kache/workspace` elsewhere |
| `$CARGO_HOME` | `<CARGO_HOME>` | `/kache/.cargo` |
| `$RUSTUP_HOME`, `$HOME`, tempdir, ... | `<...>` | `/kache/rustup`, `/kache/home`, `/kache/tmp`, ... |

Why these spellings work (verified against samply's source and by local dwarfdump experiments):

- `/rustc/<rev>/...` is the upstream virtual form for prebuilt std. samply and the Firefox Profiler already fetch those from GitHub, and rust-gdb knows them. When the rust-src component is installed, rustc de-virtualizes std paths in user-crate debug info to the local rust-src dir (verified with dwarfdump), so this rule just restores the upstream form. The commit hash comes from the `rustc -vV` output kache already caches.
- samply's crates.io matcher accepts any prefix ending in `/.cargo` before `registry/src/<registry>/<crate>-<version>/`, so `/kache/.cargo` keeps dep sources fetchable from crates.io.
[...]

Issue #414, opened 2026-06-22
feat(remote): filesystem remote backend for shared-folder caching without S3


jleni:
## Motivation

#412 is a user who wants to share one kache cache across several Linux hosts via a network folder, without standing up S3. This is a common setup (a team or CI fleet or HPC cluster with an existing NFS/SMB share) where operating an object store is unwanted overhead.

Today the only cross-host mechanism is the remote backend, and it speaks S3 only (`RemoteLayout` holds a concrete `aws_sdk_s3::Client`). Pointing `KACHE_CACHE_DIR` straight at a shared folder does not work, for two independent reasons:

- The SQLite index (`index.db`, WAL mode) cannot be shared across hosts or live on a network filesystem. From sqlite.org/wal.html: "All processes using a database must be on the same host computer; WAL does not work over a network filesystem."
- The zero-copy restore path (reflink, then hardlink) does not work over a network or virtual filesystem, so restores degrade to full byte copies or fail outright (`src/link.rs`).

## Proposal

Add a filesystem remote backend that reuses the existing v3 manifest/pack sync with a plain-folder transport instead of S3. The remote layout is already content-addressed and lock-free:

```
{prefix}/v3/manifests/{crate}/{cache_key}.json
{prefix}/v3/packs/{crate}/{cache_key}.tar.zst
```

The backend surface is small: HEAD, GET, PUT, LIST-by-prefix. A folder maps them 1:1: stat / read file / write-temp-then-atomic-rename / readdir. Each host keeps its own LOCAL cache (local index, local blobs, local reflink restore); the shared folder holds only immutable packs and manifests, written with atomic rename. That sidesteps both the WAL and the reflink problems by construction, because nothing on the shared folder is a database and nothing is ever linked off it. It is the same architecture as the S3 backend, just a different transport.

## Scope

Phase 1 (this issue):
- Introduce a backend trait over the four operations and move the current S3 path onto it.
- Add a `std::fs` backend, configured as a remote of kind `filesystem` (or `folder`) pointing at a path.
- Pull-on-miss, push-on-store, and prefetch work unchanged through the trait.

Phase 2 (separate follow-up): shared-folder GC and eviction. The immutable pull/push path is lock-free, but deletes need cross-host coordination (a lock file, or a single designated GC host). Until then, document that eviction on a shared folder is single-host or manual.

## Acceptance
- Two or more hosts sharing one folder get cross-host cache hits with no S3 server.
[...]

PR #515, opened 2026-07-15
refactor(remote): extract RemoteBackend trait for transport-specific concerns (#414)

jlucaso1:
Phase 1 of #414 opens with:

> - Introduce a backend trait over the four operations and move the current S3 path onto it.

This is that bullet on its own. The `std::fs` backend — the second bullet — follows in a separate PR: CONTRIBUTING asks to keep PRs focused, and splitting here means the folder backend arrives as a pure addition rather than a rewrite tangled up with a refactor. Structurally a refactor; `S3Backend` is the only implementor. Not *bit-for-bit* no-op, though: a few things are observable, and they are called out inline below — GET miss-classification for shards, the pack body no longer being copied, and some error/log strings that promised S3.

## Why this has to come first

`aws_sdk_s3::Client` was the concrete type threaded through `remote.rs`, `remote_layout.rs`, `remote_plan.rs`, `daemon.rs`, `cli.rs` and `fallback_planner.rs`. `RemoteLayout` held one directly, and four manifest/shard functions in `remote.rs` took one by argument alongside `bucket`. A second transport had nowhere to attach.

## The seam

The trait sits at the object level #414 describes — `head` / `get` / `put` / `list`-by-prefix — rather than at the entry level. That keeps pack framing, zstd, blake3 validation and key naming above it, shared by every transport; a folder backend only has to answer four questions about bytes. `bucket` moves inside the backend and off the call signatures.

`RemotePlan::layout()` and the daemon's `OnceCell` now speak `RemoteBackend`, so the plumbing is transport-agnostic end to end.

## Design notes

- **Absence is not an error.** `head` answers `false`, `get` answers `None`. The `NoSuchKey`/`404`/`NoSuchObject` spread of AWS and its clones stays inside the backend, where it belongs. `EntryNotFound` is still minted by the layout, so the #485 Phase 0 miss path is untouched.
- **`get` takes `max_bytes`.** The metadata cap has to be a pre-buffering guard — checking the length after the body is collected would defeat the point. Letting callers check afterwards would have quietly dropped that protection, so the cap crosses the trait.
- **`get` returns the request/body timing split** that `DownloadResult` reports, so transfer telemetry keeps its resolution instead of collapsing to one number.
- **`Arc<dyn RemoteBackend>`, not `Box`.** Shard prefetch and shard upload fan out over `tokio::spawn`, which needs an owned `'static` handle. The alternative — `join_all` — would have quietly serialized them onto one task.
- **Naming.** Where the code is transport-agnostic, messages that promised S3 now say remote; "S3 client init failed" would be a lie over a folder. Internals still named for S3 (`s3_semaphore`, `S3KeyCache`, `s3_concurrency`) are left alone — renaming those is churn that belongs with the backend that makes them wrong.

## Tests

The mock-injection points that already existed (`sync_with_client`, `upload_manifest_and_shards`, the daemon's `OnceCell`) now take the trait, so every existing wire-mock round-trip — remote-check hit/miss, GET-404-as-clean-miss, prefetch, upload-skip — still exercises the S3 path, just through the seam.

Two test groups moved with the code they cover:

- head-miss classification (`NotFound`/`NoSuchKey` → miss, `SignatureDoesNotMatch` → not a miss) → `remote_backend.rs`.
- the metadata cap: `reject_oversized_metadata` was unit-tested directly; the equivalent now drives a real GET against the mock (under cap, over cap, uncapped, absent), which covers the guard on the path it actually protects.

## Verification

`just check` is green: fmt, `clippy --workspace --all-targets -- -D warnings`, and the full
`cargo test --workspace` including the integration suite (1182 unit + integration all passing).

Correcting something from the first revision of this description: I had flagged
`cargo test --workspace` as flaky here and on `main`. That was wrong. The integration tests build
into `/tmp`, which is a 16G tmpfs on my machine, and my own repeated runs had filled it — the
failures were `No space left on device`, not test interference. With space free, the suite is
green on both.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,8 @@ resolver = "3"
 
 [dependencies]
 blake3 = "1"
+# Already transitive via aws-sdk-s3; declared for RemoteBackend's body type.
+bytes = "1"
 # TLS provider: `ring`, NOT aws-lc-rs. We avoid `aws-lc-sys` entirely
 # because its cmake/NASM C build hangs when cross-compiling to
 # `aarch64-pc-windows-msvc` under cargo-xwin; `ring` ships prebuilt asm
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -3,6 +3,8 @@ use bytesize::ByteSize;
 #[cfg(unix)]
 use std::os::unix::fs::MetadataExt;
 
+use std::sync::Arc;
+
 use crate::config::Config;
 use crate::daemon;
 use crate::events;
@@ -2587,11 +2589,11 @@ async fn sync_inner(
     lock_crates: Option<&std::collections::HashSet<String>>,
     pull_workspace: bool,
 ) -> Result<()> {
-    let client = crate::remote::create_s3_client(remote, config.s3_pool_idle_secs)
+    let backend = crate::remote_backend::create_backend(remote, config.s3_pool_idle_secs)
         .await
-        .context("connecting to S3 — check credentials and endpoint")?;
+        .context("connecting to the remote — check credentials and endpoint")?;
     sync_with_client(
-        &client,
+        backend.as_ref(),
         config,
         store,
         remote,
@@ -2606,12 +2608,12 @@ async fn sync_inner(
     .await
 }
 
-/// The S3-driven body of `sync`, with the client injected so tests can drive it
-/// against a mock S3 server. Lists remote keys, diffs against the local store,
+/// The remote-driven body of `sync`, with the backend injected so tests can
+/// drive it against a mock. Lists remote keys, diffs against the local store,
 /// then (unless `dry_run`) pulls missing artifacts and pushes local-only ones.
 #[allow(clippy::too_many_arguments)]
 async fn sync_with_client(
-    client: &aws_sdk_s3::Client,
+    backend: &dyn crate::remote_backend::RemoteBackend,
     config: &Config,
     store: &Store,
     remote: &crate::config::RemoteConfig,
@@ -2646,7 +2648,7 @@ async fn sync_with_client(
             eprint!("Listing S3 keys for {} workspace crates...", crates.len());
             let keys = planner
                 .plan(crate::remote_plan::RemoteWorkload::KeyDiscovery)
-                .layout(client, remote)
+                .layout(backend, remote)
                 .list_keys_for_crates(crates)
                 .await
                 .context("listing S3 keys for workspace crates")?;
@@ -2659,7 +2661,7 @@ async fn sync_with_client(
             eprint!("Listing S3 keys for {} crates...", crates.len());
             let keys = planner
                 .plan(crate::remote_plan::RemoteWorkload::KeyDiscovery)
-                .layout(client, remote)
+                .layout(backend, remote)
                 .list_keys_for_crates(crates)
                 .await
                 .context("listing S3 keys for dependency crates")?;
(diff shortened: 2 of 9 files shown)


emmanuelm41:
Really nice work on this — the seam is in exactly the right place, and thanks for the writeup 🙏

I ran a second pass on it, and a few things surfaced. None are hard merge-blockers for the S3-only present, but the first three are cheap to fix now and get harder once the folder backend exists — I'd lean toward doing them here. Fine with any landing in the follow-up though.

1. **`GetObject.body` copies every download.** `body.into_bytes().to_vec()` in `S3Backend::get` — `into_bytes()` already yields a contiguous `bytes::Bytes`, so `.to_vec()` is a second full heap copy of the whole object, and on the pack path (`download_entry`, uncapped) those can be big. Two extra wrinkles: it's now on the hot restore path where it wasn't before, and the copy happens after `body_ms` is recorded and before `extract_ms` — so it's an untimed gap in the transfer telemetry, not just latency. Making `GetObject.body: bytes::Bytes` drops the copy; `from_slice(&..)` and the pack path both take `&[u8]` unchanged, and the folder backend can `Bytes::from(vec)` for free. `bytes` is already in the tree.

2. **`max_bytes` isn't actually a hard cap.** The check only fires when `resp.content_length()` is `Some`; if Content-Length is absent (chunked responses from some S3 clones/proxies, and eventually the folder backend) the body is buffered unbounded — but the trait doc now promises "a hostile or corrupt object can't force an unbounded allocation." For real S3 today GetObject always sends the length so it works, but the doc is stronger than the code. I'd either fail-closed when capped-and-no-length, or soften the doc to "checks the advertised size." (This matches the old `reject_oversized_metadata`, which also allowed `None` — the change is that the trait now documents it as a guarantee.)

3. **GET miss-classification quietly widened for shards.** Routing `download_shard` through `backend.get()` means it now uses `is_missing_get_object`, which checks raw HTTP 404 first. `download_shard` previously returned `None` only on `is_no_such_key()`; now a structured `NoSuchBucket` (also a 404) classifies as object-absence, so a missing/misconfigured bucket silently becomes an ordinary cache miss instead of an error. The pack path already behaved this way, but spreading it to shards is new. Letting the structured error code be authoritative, with raw-404 only as a code-less fallback, keeps the clone-compat behavior without swallowing config errors.

4. **Two small ones.** `handle_prefetch` in `daemon.rs` still returns `"S3 client init failed"` while every sibling got switched to `"remote backend init failed"` — just missed in the sweep. And `describe_head_object_error` (now in `remote_backend.rs`) changed its empty-case fallback from Display (`err.to_string()`) to Debug (`{err:?}`) and dropped the old no-status/no-code branch — cosmetic, only affects a non-miss head error log, but flagging since this is a "no behavior change" PR. Intentional?

On the "no behavior change" banner — it's almost true, and the structural work is clean, but the shard miss-widening, the pack copy, and a handful of error/log string changes are observable. Worth a one-line acknowledgment in the description rather than dropping the claim.

Test-wise, coverage is well preserved by wrapping the existing wire-mock client in `S3Backend`. If you touch this again, the gaps worth filling are: a multi-page LIST test (pagination moved modules but only single-page is covered), exact-at-cap and no-Content-Length cases, and a `NoSuchBucket`-vs-`NoSuchKey` test so #3 can't regress silently.

Ownership is solid — the `Arc` + `tokio::spawn` fan-out in `upload_shards` / `shard_prefetch_for_deps` is correct, and `Arc` over `Box` is the right call there. Nothing here should hold up the merge. Great first slice of #414. 🚀

emmanuelm41 (approved):

emmanuelm41 (approved):
Approving — clean, well-motivated extraction and the seam is in the right place. Left inline notes above; none are blockers, the first three are just worth folding in here or in the folder-backend follow-up. 🚀

emmanuelm41 (approved):

PR #518, opened 2026-07-15
fix(events): make log rotation atomic and synchronized

Ysh204:
## Summary

Fixes a race in log rotation where concurrent wrapper processes could silently lose events.

`log_event()` and `log_transfer()` already serialize writes using the existing sidecar lock, but `rotate_if_needed()`, `rotate_transfers_if_needed()`, and `clear_events()` previously performed unlocked read-modify-write operations. Under concurrent builds, a rotation could overwrite events appended by another process.

This PR makes log rotation participate in the existing locking protocol and performs atomic file replacement to avoid exposing partially-written log files.

### Synchronization

- Acquire the existing `open_log_lock()` before checking file size, reading, trimming, or rewriting the log.
- Update `clear_events()` to use the same locking mechanism.

### Atomic rotation

- Write rotated contents to a unique temporary file.
- Flush the temporary file before replacing the original log.
- Atomically replace the original log with the rotated version.

### Refactoring

- Extract a shared `rotate_log_impl()` used by both:
  - `rotate_if_needed()`
  - `rotate_transfers_if_needed()`

### Tests

Added a regression test that exercises concurrent log appends and rotation to verify that events are not lost or corrupted under parallel execution.

## Test plan

Verified locally with:

- [x] `cargo fmt --all`
- [x] `cargo test`

Including the new concurrent log rotation regression test.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -301,28 +301,137 @@ impl EventTailer {
 
 /// Rotate the event log if it exceeds the max size.
 /// Keeps the last `keep_lines` lines.
+fn rotate_log_impl(
+    log_path: &Path,
+    max_size: u64,
+    keep_lines: usize,
+    log_label: &str,
+) -> Result<()> {
+    if !log_path.exists() {
+        return Ok(());
+    }
+
+    // 1. Acquire the lock before querying metadata or reading
+    let lock = open_log_lock(log_path).context("opening log lock")?;
+    lock.lock().context("locking log for rotation")?;
+
+    let res = (|| -> Result<()> {
+        let meta = fs::metadata(log_path)?;
+        if meta.len() <= max_size {
+            return Ok(());
+        }
+
+        // Clean up stale temp files in the same directory (older than 5 minutes)
+        if let Some(parent) = log_path.parent() {
+            if let Ok(entries) = fs::read_dir(parent) {
+                let file_prefix = log_path.file_name().and_then(|n| n.to_str()).unwrap_or("");
+                let temp_marker = format!("{}.tmp.", file_prefix);
+                for entry in entries.flatten() {
+                    let path = entry.path();
+                    if let Some(name) = path.file_name().and_then(|n| n.to_str()) {
+                        if name.starts_with(&temp_marker) {
+                            if let Ok(meta) = fs::metadata(&path) {
+                                if let Ok(modified) = meta.modified() {
+                                    if let Ok(age) = modified.elapsed() {
+                                        if age > std::time::Duration::from_secs(300) {
+                                            let _ = fs::remove_file(&path);
+                                        }
+                                    }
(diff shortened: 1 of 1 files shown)


emmanuelm41 (changes requested):
Thanks @Ysh204 — the append/rotation race fix is right, and moving to temp-file + atomic rename is the correct instinct. Two things I'd like sorted before this goes in, though, both real:

**1. This breaks `kache tui`'s live view after the first rotation.** `EventTailer::poll` follows an already-open file handle and only detects rotation via `len < position`. The old in-place `fs::write` shrank the same inode, so that worked — but the rename now swaps in a *new* inode, and the tailer keeps stat-ing the orphaned old one (length frozen), so it returns empty forever and never reopens. The TUI polls this live, so it silently freezes on any long build. Fix: have `poll()` stat the *path* and reopen when the inode/length changes (rotation-mechanism-agnostic).

**2. The concurrent regression test doesn't actually prove no-loss.** With a 300-byte cap and ~500-byte records, rotation collapses the log to a single line, so the gap check ends up `[x]` vs `x..=x` — trivially true. It'd also pass if every append errored (those errors are ignored). Worth a version that keeps enough lines for a real gap to be observable, and ideally one that exercises the actual `EventTailer` across a rotation so #1 can't regress again.

Two smaller follow-ups (fine after merge): the temp-file + fsync + rename + Windows-retry logic is nearly identical to what `store.rs` already has (the Windows errno handling especially) — worth extracting a shared helper so they don't drift; and the rotation path now takes the lock before the size pre-check, adding a lock to every compile — a lock-free stat gate in front keeps the hot path cheap.

Nice work on the core fix — just want the tailer regression and a non-vacuous test in before it lands.

emmanuelm41 (approved):
CI is green across the board now — including Test (Windows), which matters here because it's the first time the `#[cfg(windows)]` `get_file_identity` FFI has been compiled *and* exercised: `test_event_tailer_handles_rename_rotation` ran on the Windows runner and passed.

I also checked that guard test is real rather than decorative — neutering the reopen-on-rotation fix makes it fail (`left: 0, right: 3`, exactly the freeze), and it goes green with the fix restored. That's a genuine regression test.

Both of my asks are addressed: the tailer now compares dev/inode (Unix) and file identity (Windows) and reopens on rotation, and the concurrent test is no longer vacuous (4000-byte cap / keep 20, plus it drives a real `EventTailer` across a rotation).

Clearing my changes-requested. Thanks for sticking with this one through several rounds — the race fix is real and the tailer regression would have been genuinely nasty to debug in the wild.

Filing a follow-up for the one thing left: the temp-file + fsync + rename + Windows-errno-retry logic here is nearly identical to what `store.rs` already has, and I'd rather they share one helper than drift apart.

PR #519, opened 2026-07-15
fix(compiler): detect bare workspace wrappers configured by Cargo

### LINKED ISSUE #516: Workspace-wrapper bare-name detection misses CARGO_BUILD_* and .cargo/config.toml
Follow-up from #506 (merged in 86e7b9c). Not a regression — #506 strictly improves the situation — but the bare-name detection it added is narrower than the test implies.

## Summary

`is_workspace_wrapper_chain` (`src/compiler/mod.rs:593`) detects a bare (separator-free) workspace-wrapper name only when it exactly matches `env::var_os("RUSTC_WORKSPACE_WRAPPER")`. Cargo produces a bare workspace-wrapper name through **three** config surfaces; the other two are not covered and still fail with the original #505 error.

## Repro (real cargo + kache at 86e7b9c)

```sh
# helper: a bare-named wrapper on PATH
mkdir -p bin && printf '#!/bin/sh\nexec "$@"\n' > bin/mydriver && chmod +x bin/mydriver
export PATH="$PWD/bin:$PATH"
cargo new --lib demo && cd demo

# 1) CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER (env), RUSTC_WORKSPACE_WRAPPER unset
env -u RUSTC_WORKSPACE_WRAPPER RUSTC_WRAPPER="$(command -v kache)" \
    CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER=mydriver cargo build
#  -> error: unrecognized subcommand 'mydriver'

# 2) .cargo/config.toml — NO env var set at all
mkdir -p .cargo && printf '[build]\nrustc-workspace-wrapper = "mydriver"\n' > .cargo/config.toml
env -u RUSTC_WORKSPACE_WRAPPER -u CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER \
    RUSTC_WRAPPER="$(command -v kache)" cargo build
#  -> error: unrecognized subcommand 'mydriver'
```

Both invocations reach `kache mydriver <abs>/rustc -vV` and fall through to clap.

## Why env-matching can't fully fix it

Case 2 (`.cargo/config.toml build.rustc-workspace-wrapper`) sets **no environment variable** in the compiler process, so checking additional env spellings (e.g. also reading `CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER`) fixes case 1 but not case 2. A complete fix needs a structural signal instead of an env match — e.g. resolve `args[0]` against `PATH` and treat `<resolved-exe> rustc <args>` as a chain when `args[0]` is not one of kache's own subcommands, or otherwise detect the chain from shape alone.

## Note on the existing test

`workspace_wrapper_chain_detects_bare_name_via_env` (`src/compiler/mod.rs:728`) passes and reads as "bare names are handled," which overclaims — it only exercises the `RUSTC_WORKSPACE_WRAPPER` env spelling. Whatever fix lands should add coverage for the `CARGO_BUILD_*` and `.cargo/config.toml` surfaces (ideally a real-cargo integration test with a PATH-only wrapper), or the test/comment should be scoped honestly.

## Scope

The core #505 case (`cargo dylint`, which sets an absolute-path `RUSTC_WORKSPACE_WRAPPER`) works. This is an extension of the bare-name handling, low priority.

Ysh204:
## Summary

Closes #516.

This PR extends workspace-wrapper chain detection to support bare (separator-free) wrapper names configured through Cargo, not just `RUSTC_WORKSPACE_WRAPPER`.

Previously, bare wrapper names were only recognized when they matched the `RUSTC_WORKSPACE_WRAPPER` environment variable. As a result, wrappers configured through `CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER` or `.cargo/config.toml` (`build.rustc-workspace-wrapper`) were misidentified and treated as `kache` subcommands.

### Compiler detection

- Keep the existing absolute/relative path detection.
- Keep the existing `RUSTC_WORKSPACE_WRAPPER` environment matching.
- Add detection for bare wrapper names by:
  - ignoring known `kache` subcommands and global flags,
  - resolving the wrapper executable on `PATH` (including Windows `PATHEXT` handling),
  - verifying that the remaining arguments form a recognized `rustc` compiler invocation.

### Compiler execution

- Update `run_compiler_directly` to execute the resolved wrapper path, allowing Windows batch-script wrappers (`.bat` / `.cmd`) to work correctly.
- Add a regression test to ensure the hardcoded subcommand list stays synchronized with the Clap command definitions.

### Tests

Added coverage for:

- PATH-resolved bare wrapper detection.
- `CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER`.
- `.cargo/config.toml` (`build.rustc-workspace-wrapper`).
- Negative cases to ensure unrelated commands are not misidentified as wrapper chains.

### DIFF
--- a/src/compiler/mod.rs
+++ b/src/compiler/mod.rs
@@ -590,14 +590,90 @@ pub fn detect_compiler(args: &[String]) -> Option<&'static CompilerAdapter> {
 /// unrecognized workspace wrapper. Cargo passes `<wrapper> rustc <args>`;
 /// the wrapper may be an absolute path or a bare name resolved via PATH.
 /// We match the inner rustc, not the wrapper name.
+#[cfg(unix)]
+fn is_executable(path: &std::path::Path) -> bool {
+    use std::os::unix::fs::PermissionsExt;
+    std::fs::metadata(path)
+        .map(|metadata| metadata.is_file() && metadata.permissions().mode() & 0o111 != 0)
+        .unwrap_or(false)
+}
+
+#[cfg(not(unix))]
+fn is_executable(path: &std::path::Path) -> bool {
+    path.is_file()
+}
+
+pub(crate) fn is_kache_subcommand_or_flag(s: &str) -> bool {
+    if s.starts_with('-') {
+        return true;
+    }
+    use clap::CommandFactory;
+    let mut cmd = crate::Cli::command();
+    cmd.build();
+    cmd.find_subcommand(s).is_some()
+}
+
+pub(crate) fn resolve_program_on_path(program: &str) -> Option<std::path::PathBuf> {
+    if program.contains('/') || program.contains('\\') {
+        return Some(std::path::PathBuf::from(program));
+    }
+    let path = std::env::var_os("PATH")?;
+    let dirs: Vec<std::path::PathBuf> = std::env::split_paths(&path).collect();
+
+    let extensions: Vec<String> = if cfg!(windows) {
+        if let Some(pathext) = std::env::var_os("PATHEXT") {
+            std::env::split_paths(&pathext)
+                .filter_map(|p| p.to_str().map(|s| s.to_string()))
+                .collect()
+        } else {
+            vec![
+                ".exe".to_string(),
+                ".bat".to_string(),
+                ".cmd".to_string(),
+                ".com".to_string(),
+            ]
+        }
+    } else {
+        vec!["".to_string()]
+    };
+
+    for dir in dirs {
+        let p = dir.join(program);
+        if is_executable(&p) {
+            return Some(p);
+        }
+        for ext in &extensions {
+            if ext.is_empty() {
+                continue;
+            }
+            let mut suffixed = p.clone().into_os_string();
+            suffixed.push(ext);
+            let suffixed_path = std::path::PathBuf::from(suffixed);
+            if is_executable(&suffixed_path) {
+                return Some(suffixed_path);
+            }
+        }
+    }
+    None
+}
+
+fn is_program_on_path(program: &str) -> bool {
+    resolve_program_on_path(program).is_some()
+}
+
 pub fn is_workspace_wrapper_chain(args: &[String]) -> bool {
     if args.len() < 2 || !rustc::RustcCompiler::recognizes(&args[1..]) {
         return false;
     }
-    args[0].contains('/')
-        || args[0].contains('\\')
-        || std::env::var_os("RUSTC_WORKSPACE_WRAPPER")
-            .is_some_and(|w| w == std::ffi::OsStr::new(&args[0]))
+    if args[0].contains('/') || args[0].contains('\\') {
+        return true;
+    }
+    if std::env::var_os("RUSTC_WORKSPACE_WRAPPER")
+        .is_some_and(|w| w == std::ffi::OsStr::new(&args[0]))
+    {
+        return true;
+    }
+    !is_kache_subcommand_or_flag(&args[0]) && is_program_on_path(&args[0])
 }
 
 /// Extract the bare command name from an `argv[0]`, splitting on both Unix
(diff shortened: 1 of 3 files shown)


emmanuelm41:
Thanks @Ysh204 — this is basically right, and the `CARGO_BUILD_RUSTC_WORKSPACE_WRAPPER` / `.cargo/config.toml` coverage plus the clap-sync test are good instincts. Two things I'd like folded in before merge:

**1. `resolve_program_on_path` matches on `is_file()` with no executable check.** On Unix that has two consequences: a non-executable file shadowing a name on PATH gets misclassified as a wrapper, and `run_compiler_directly` then execs it and fails EACCES where the old `Command::new` would've kept searching PATH. (The tell: the new unit test creates the wrapper with `File::create` and no `+x`, and passes *because* of this gap.) An exec-bit check — or the `which` crate — fixes both.

**2. The hardcoded subcommand list is a sync-forever tax with an alias hole.** The sync test only checks subcommand *names*, so the day someone adds e.g. `alias = "ls"` to a command, the test stays green and `kache ls rustc` resolves `/bin/ls` on PATH and execs it. Since the list is fully derivable, I'd rather drive it off `Cli::command().find_subcommand()` (covers names, aliases, and hidden commands for free, and deletes the sync test), treat `starts_with('-')` as CLI for flags, and let clap be the source of truth. No current bug — there are no aliases today — but it removes the landmine before it's laid.

Minor, not blocking: the PATHEXT lowercasing isn't needed (and can hurt on a case-sensitive dir), and the Windows test creates both an extensionless and a `.exe` file so it doesn't actually prove `.exe` resolution.

Solid fix for #516 overall — just those two before it merges and I think it's in good shape.

Ysh204:
Thanks for the feedback @emmanuelm41! I have updated the PR to address all points:

1. **Executable-Bit Check**: Added a lightweight helper `is_executable` (`metadata.permissions().mode() & 0o111 != 0` on Unix, and standard `is_file()` on Windows/non-Unix) rather than adding the `which` crate as a new dependency to keep things minimal. Also updated the Unix unit test to set the execute permissions (`0o755`) on the mock wrapper.
2. **Dynamic Subcommand Resolution**: Replaced the hardcoded list with `Cli::command().find_subcommand(s)` (along with a check for flag arguments starting with `-`) so that Clap is the single source of truth for subcommands, aliases, and hidden commands. Deleted the now-obsolete `test_hardcoded_subcommands_match_clap` sync test.
3. **Windows/PATHEXT & Test Hardening**: Removed the `PATHEXT` lowercasing, and modified the Windows unit test to *only* create the `.exe` wrapper file (no extensionless file) to verify that bare-name `.exe` suffix resolution is working correctly.

emmanuelm41:
Thanks @Ysh204 — the exec-bit helper and the clap-derived check are both exactly right, and dropping the PATHEXT lowercasing plus fixing the Windows test to only create the `.exe` are nice touches. Pulled it and it's green here: fmt, clippy, unit + the two e2e cargo-build tests.

One catch, and it's my fault — the refactor I suggested has a hole. `Cli::command().find_subcommand("help")` returns **false**: clap only inserts its auto-generated `help` subcommand during `build()`, so it isn't in `get_subcommands()` yet. The old hardcoded list had `"help"` explicitly, so this silently lost it. Probed it directly:

```
PROBE token=help       is_cli=false   <-- regression
PROBE token=-h         is_cli=true
PROBE token=--help     is_cli=true
PROBE token=gc / list  is_cli=true
```

Unix is fine (`help` is a shell builtin, not a file on PATH, so it falls through to CLI). **Windows isn't**: `help.exe` is a real binary in `System32`, which is on PATH — so `kache help rustc` would resolve it, classify as a wrapper chain, and exec `help.exe` instead of printing kache's help. Easy to miss precisely because it only bites on Windows.

The fix keeps clap as the source of truth — just build the command first:

```rust
let mut cmd = crate::Cli::command();
cmd.build();
cmd.find_subcommand(s).is_some()
```

Verified both directions locally:

```
Cli::command().find_subcommand("help")            -> false
Cli::command() + .build() -> find_subcommand("help") -> true
                           + find_subcommand("gc")   -> true
```

Worth a small test asserting `help`/`-h`/`--help`/`-V`/`--version` all classify as CLI, so this can't silently regress again.

Sorry for the round-trip — that one was on my suggestion, not yours. With that in, I think #519 is good to go.

emmanuelm41 (approved):
CI is green across the board, including Test (Windows) — which is the one that mattered here, since the `help` gap only bites on Windows (`help.exe` lives in System32 and is on PATH).

Verified the fix locally too:

```
PROBE token=help       is_cli=true   <-- was false
PROBE token=-h         is_cli=true
PROBE token=--help     is_cli=true
PROBE token=-V         is_cli=true
PROBE token=--version  is_cli=true
PROBE token=gc / list / doctor  is_cli=true
```

`cmd.build()` before `find_subcommand` does exactly what's needed, and the test you added pins `help`/`-h`/`--version` so it can't silently regress again.

Thanks for the patience on this one — you ended up fixing a hole in my own suggestion, and #516 is properly closed: exec-bit check in place, clap as the single source of truth for subcommands/aliases/hidden commands, no hand-maintained list to drift.

Issue #411, opened 2026-06-22
Support more flags needed for Firefox build


saschanaz:
## Problem

```
 0:29.89 W 2026-06-22T13:56:17.549952Z DEBUG kache::compiler::cc: [cc:Unified_cpp_dom_ipc5.cpp] flag classify: 1 modeled / 12 probe / 66 preprocessor / 60 no-effect / 1 parser-handled / 0 user-allowed / 11 unmodeled
 0:29.89 W 2026-06-22T13:56:17.549993Z DEBUG kache::compiler::cc: cc unsupported flag(s): -TP -ffile-reproducible -Xclang -Xclang -MP -Xclang -dependency-file -Xclang -Xclang -Xclang -fansi-escape-codes — not yet — passthrough
 0:29.89 W 2026-06-22T13:56:17.550000Z DEBUG kache::wrapper: cc: passthrough (cc unsupported flag(s): -TP -ffile-reproducible -Xclang -Xclang -MP -Xclang -dependency-file -Xclang -Xclang -Xclang -fansi-escape-codes — not yet)
 0:29.95 W 2026-06-22T13:56:17.614186Z DEBUG kache::compiler::cc: [cc:Unified_cpp_dom_media_doctor0.cpp] flag classify: 1 modeled / 12 probe / 33 preprocessor / 60 no-effect / 1 parser-handled / 0 user-allowed / 11 unmodeled
 0:29.96 W 2026-06-22T13:56:17.614232Z DEBUG kache::compiler::cc: cc unsupported flag(s): -TP -ffile-reproducible -Xclang -Xclang -MP -Xclang -dependency-file -Xclang -Xclang -Xclang -fansi-escape-codes — not yet — passthrough
 0:29.96 W 2026-06-22T13:56:17.614240Z DEBUG kache::wrapper: cc: passthrough (cc unsupported flag(s): -TP -ffile-reproducible -Xclang -Xclang -MP -Xclang -dependency-file -Xclang -Xclang -Xclang -fansi-escape-codes — not yet)
 0:29.98 W 2026-06-22T13:56:17.638651Z DEBUG kache::compiler::cc: [cc:Unified_cpp_dom_ipc6.cpp] flag classify: 1 modeled / 12 probe / 66 preprocessor / 60 no-effect / 1 parser-handled / 0 user-allowed / 11 unmodeled
[...]

Issue #286, opened 2026-06-10
Firefox fails to build swgl


saschanaz:
## What happened

```
 3:32.45 E error: failed to run custom build command for `swgl v0.1.0 (D:\firefox\gfx\wr\swgl)`
 3:32.45 E note: To improve backtraces for build dependencies, set the CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.
 3:32.45 E Caused by:
 3:32.46 E   process didn't exit successfully: `D:/firefox/obj-aarch64-pc-windows-msvc\release\build\swgl-a292cf735a64f6bf\build-script-build` (exit code: 101)
 3:32.46 E   --- stdout
 3:32.46 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/brush_blend.glsl
 3:32.46 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/shared.glsl
 3:32.46 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/base.glsl
 3:32.46 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/prim_shared.glsl
 3:32.46 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/rect.glsl
 3:32.48 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/render_task.glsl
 3:32.49 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/transform.glsl
 3:32.50 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/image_source.glsl
 3:32.50 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/gpu_buffer.glsl
 3:32.50 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/brush.glsl
 3:32.50 E   cargo:rerun-if-changed=D:\firefox\gfx\wr\swgl/../webrender/res/blend.glsl
 3:32.52 E   OPT_LEVEL = Some(0)
[...]

Issue #95, opened 2026-05-16
cc: user-configurable flag classification — extend the allow-list locally


jleni:
## Idea

Let advanced users **extend kache's cc flag classifier via config** — opt a flag into caching *before* kache ships built-in support for it in a release.

## Why

PR #94 made cc caching an **allow-list**: a flag is cacheable only if kache recognizes it as cache-safe (modeled codegen / preprocessor-captured / no-object-effect); everything else passes through, and the refusal **names the rejected flags**.

That's correct, but conservative — e.g. the `cc` crate passes `-ffunction-sections -fdata-sections --target=… -mmacosx-version-min=…`, so `cc`-crate C builds currently don't cache. Built-in support for those will land via the support-matrix work, but a user blocked by one flag shouldn't have to wait for the next kache release.

A config knob closes that gap:
- **Pairs with the rejected-flags report** — the report tells you *what's* blocked; the config lets you act on it *now*.
- **Demand signal** — flags users keep adding to their configs are exactly the ones kache should classify built-in next.
- It's a *user overlay* on kache's built-in matrix: matrix = what kache ships, config = what you've added locally. When built-in support lands, you drop the entry.

## Safe design (important)

The config must not become a miscache footgun. Scope it narrowly:

**The config can only add flags to the "hash into the key" set.** The user declares a flag; kache then (a) stops refusing it and (b) hashes the flag verbatim into the cache key. Sketch:

```toml
[cc]
# Also treat these as cacheable; each is hashed verbatim into the key.
extra_codegen_flags = ["-ffunction-sections", "-fdata-sections", "-fno-rtti"]
```

This is **correct by construction** for any flag whose effect is fully determined by its string — which is ~every `-f…` codegen flag users actually hit. Two builds differing by the flag get different keys; no collision.

### Non-goals (deliberately excluded)

- **No "ignore this flag" mode.** That is the actual footgun: if the user is wrong about a flag being inert, it's a silent miscache. (ccache's `sloppiness` is the cautionary precedent — powerful, explicitly at-your-own-risk.)
- **No override of the structural refuses** — link mode, `--coverage`, multi-source, response files. The config extends *flag classification* only, never "cache something kache architecturally can't."

### Residual sharp edge

[...]

PR #520, opened 2026-07-16
cc: recognize cargo-zigbuild zigcc wrappers as Clang-family compilers

lwintermelon:
## Summary

Fix a build failure when kache is used together with
[cargo-zigbuild](https://github.com/rust-cross/cargo-zigbuild) as `rustc-wrapper`.

cargo-zigbuild generates zig cc wrapper scripts named
`zigcc-{target}.{glibc_ver}-{hash}.sh` for cross-compilation with glibc version
pinning. When kache acts as `rustc-wrapper`, Cargo wraps CC invocations from
build scripts (e.g., `ring`) through kache — the command becomes:

```
kache /path/to/zigcc-x86_64-unknown-linux-gnu.2.35-a940.sh -Oz -c foo.c -o foo.o
```

kache's name-based compiler detection (`recognizes()`) only matches
https://github.com/kunobi-ninja/kache/blob/1b7382ce636fb646f8530809a2c422caf84ac392/src/compiler/cc.rs#L3452-L3471

basenames `cc`, `c++`, `gcc`, `g++`, `clang`, `clang++` and their versioned
variants. The zigcc wrapper doesn't match, so `detect_log_mode` falls to CLI
mode, and clap errors with "unrecognized subcommand".

Fix: add `zigcc` to the recognized stems in both `recognizes()` and
`ToolFamily::detect()`, classifying it as `Clang` (zig's cc is clang-based).

A better long-term approach is dynamic compiler detection via `-E` probing (as
[sccache](https://github.com/mozilla/sccache) does in its `detect_c_compiler()`):
https://github.com/mozilla/sccache/blob/4fcb1611767704dcbf969ee10a5bdb35ef3eb959/src/compiler/compiler.rs#L1644
when the name doesn't match, run the binary with `-E` on a test file containing
`#if defined(__clang__)` / `#elif defined(__GNUC__)` to discover the compiler's
identity. This would cover *any* cc-compatible wrapper without allowlist
maintenance. kache already has `-E` awareness in `recognizes_family_probe()`:
https://github.com/kunobi-ninja/kache/blob/1b7382ce636fb646f8530809a2c422caf84ac392/src/compiler/cc.rs#L3473-L3496

but it is used for passthrough in the opposite direction
(the `cc` crate probing kache rather than kache probing the compiler).

## Test plan

- [x] `cargo build` with kache as rustc-wrapper — rustc caching intact
- [x] `cargo zigbuild` with kache as rustc-wrapper — zig cc wrappers
  recognized as Clang, C/C++ compilation cached by kache

## Checklist

- [ ] New behaviour is covered by tests where it makes sense


src/compiler/cc.rs
```diff
@@ -3465,6 +3467,22 @@ impl CcCompiler {
...
+        // scripts are named `zigcc-{target}.{glibc_ver}-{hash}.sh` and
+        // delegate to `cargo-zigbuild zig cc -- ...`. Zig's cc is clang-based
+        // (defines `__clang__`), so we treat it as a clang-family compiler.
+        // See https://github.com/rust-cross/cargo-zigbuild.
+        //
+        // NOTE: A better long-term approach is dynamic compiler detection
+        // via `-E` probing (as sccache does). That would cover *any*
+        // cc-compatible wrapper regardless of its filename, rather than
+        // maintaining a name-based allowlist. Tracked in follow-up.
+        if stem == "zigcc" {
```
emmanuelm41, line 3482: Three things here:

**Add `zigcxx`** (see the main review comment) — otherwise C++ build scripts under `cargo zigbuild` still fail with the same clap error this PR fixes.

**Consider folding both stems into the existing lists** rather than a special-cased early return. This preserves both behaviours (bare name and suffixed) and keeps one place to edit:

```rust
if matches!(name, "cc" | "c++" | "gcc" | "g++" | "clang" | "clang++" | "zigcc" | "zigcxx") {
    return true;
}
let stem = name.split('-').next().unwrap_or("");
matches!(stem, "cc" | "c++" | "gcc" | "g++" | "clang" | "clang++" | "zigcc" | "zigcxx")
    && name.len() > stem.len()
    && name.as_bytes()[stem.len()] == b'-'
```

**"Tracked in follow-up" is a dangling reference** — no issue is linked. The convention in this module is a concrete number (`"Issue #411 —"`, `"issue #286"`, `"#299/#312"`, `"issue #95"`). Either file the `-E` probing issue and cite it here, or drop the sentence. FWIW the sccache-style dynamic-detection rationale from the PR description is good material for that issue — better there than as a 12-line inline comment.

src/compiler/cc.rs
```diff
@@ -85,7 +87,7 @@ impl ToolFamily {
             return ToolFamily::ClangCl;
         }
         let stem = name.split('-').next().unwrap_or("");
-        if stem == "clang" || stem == "clang++" {
+        if stem == "clang" || stem == "clang++" || stem == "zigcc" {
```
emmanuelm41, line 90: Worth knowing (not asking you to change it): this half is a behavioural no-op today. Every consumer of `family` goes through `.dialect()` — lines 870, 929, 1179, 2515, 2610, 2785, 2824, 3016, 3753 — and both `Gnu` and `Clang` map to `Dialect::Gnu`. So the `recognizes()` change alone is what unbreaks the build.

Keep it — zig cc is clang-based and defines `__clang__`, so this is the correct classification for whenever `family` gains a non-dialect consumer. Just flagging that it changes nothing observable right now, and nothing tests it.

emmanuelm41 (commented):
🤖 **AI-generated review** — produced by Claude Code at @emmanuelm41's request, verified by reading the code at base `1b7382c`. Treat it as review input, not a verdict; a human still owns the call.

---

Nice diagnosis — the mechanism checks out end to end. `command_basename` yields the full script name, `.sh` isn't stripped (only `.exe` is), `split('-').next()` gives `zigcc`, and placing the check ahead of the versioned-variant branch also lets a bare `zigcc` match, which the dash guard would otherwise reject. That looks intentional and I think it's the right call.

I also went looking for miscache risk and didn't find any: `cache_key` hashes the wrapper's file name (which already embeds cargo-zigbuild's arg hash + glibc version) and, separately, the first line of `<wrapper> --version` — for zig that resolves to `clang version X (zig-bootstrap <sha>)`, so a zig upgrade that moves the bundled clang moves the key. Good.

One real gap before this ships, plus smaller things inline.

## `zigcxx` is missing

cargo-zigbuild generates two wrappers, not one — from its `src/zig.rs`:

- `zigcc-{file_target}-{:x}.{sh,bat}`
- `zigcxx-{file_target}-{:x}.{sh,bat}`

`"zigcxx".split('-').next()` is `"zigcxx"`, which matches neither the new check nor the `matches!` list — so C++ build scripts under `cargo zigbuild` still hit the exact clap error this PR fixes.

That's not hypothetical: kache treats `CXX` as a first-class wrapped variable — `is_cc_family_env_key` matches `CXX` / `CXX_*` / `TARGET_CXX`, and `recover_probe_compiler` carries dedicated `is_cxx_env_key` precedence logic. I suspect the test plan passed because `ring` is C-only.

Adding `zigcxx` is safe: language comes from the source extension and `-x`, never from the compiler name, so both wrappers correctly classify as `Clang`.

## Tests

The checklist leaves this unchecked, and this module tests name-matching pretty exhaustively — `recognizes_canonical_command_names`, `recognizes_versioned_variants`, `recognizes_rejects_non_c_compilers`, and `tool_family_detect_*` (~line 4210) are the natural slots. Given the whole change *is* name matching, coverage is cheap and high-value:

- a real-world `zigcc-x86_64-unknown-linux-gnu.2.35-a940.sh`
- the `zigcxx-` counterpart
- the `.bat` variant (cargo-zigbuild emits `.bat` on Windows)
- `ToolFamily::detect("zigcc-...sh", &[]) == ToolFamily::Clang`

## Minor, non-blocking

- On Windows the wrapper is `.bat`, and `CcProber::probe` does `Command::new(req.compiler).arg("--version")`. Rust std routes `.bat`/`.cmd` through `cmd.exe` (since the CVE-2024-24576 fix in 1.77.2), so this should work — just noting nobody's exercised it.
- Small thing in the description: `RUSTC_WRAPPER` doesn't cause Cargo to route CC through kache. kache lands in `CC`/`CXX` because those are set to `kache <compiler>` — that's exactly what `recover_probe_compiler` exists to untangle. Doesn't affect the code, but the body currently reads as if Cargo does the wrapping.

Happy to see this land once `zigcxx` is covered and there's a test or two.

emmanuelm41:
One more thing I missed on the first pass — sorry for the drip-feed.

Adding a recognized compiler is a user-visible compatibility change, and kache asserts its recognized-compiler list in three places this PR doesn't touch:

- `README.md:119` — "It recognizes `cc`, `c++`, `gcc`, `g++`, `clang`, and `clang++` (plus versioned variants like `gcc-13`), and `clang-cl` or any `--driver-mode=cl` invocation"
- `docs/getting-started/c-cpp.mdx:3` — frontmatter: "Wrap cc, c++, gcc, clang, and clang-cl for local object-compile caching."
- `docs/getting-started/c-cpp.mdx:31` — "kache recognizes `cc`, `c++`, `gcc`, `g++`, `clang`, `clang++`, versioned variants like `gcc-13`, `clang-cl`, and `clang --driver-mode=cl`."

The checklist has "User-visible changes (CLI, config, output) are reflected in the README or relevant docs" ticked, so I read this as an oversight rather than a disagreement about scope.

Worth naming cargo-zigbuild explicitly in at least the README C/C++ section and `c-cpp.mdx:31` — "does kache work with cargo-zigbuild" is the kind of thing people search for before they try it, and today the support would be silent. The mdx frontmatter one is easy to miss since it's not in the body.

*(AI-assisted review, same as above — a maintainer still owns the call.)*

lwintermelon:
Nice feedback, I'll look into it when I have time.

emmanuelm41 (approved):
Thanks @lwintermelon — nice catch, and a well-scoped fix. CI is green across all platforms.

The `zigcc` stem classification is right: `zig cc` literally reports `clang version 21.1.0`, so treating it as Clang-family is accurate rather than approximate.

The thing I wanted to be sure of before merging was cache-key safety, since the wrapper name encodes real semantics (`zigcc-{target}.{glibc_ver}-{hash}.sh`) and kache deliberately keys compilers by probed identity rather than path — so two zigcc wrappers must not collide. Checked it locally, and it's safe, though the margin is narrower than it looks:

- `version_line` is captured first-line-only, and `zig cc --version`'s first line is `clang version 21.1.0` for *every* target — so compiler identity alone does **not** distinguish them.
- The `-###` resolved tokens do carry it, including the glibc pin:
  ```
  -triple x86_64-unknown-linux5.10.0-gnu2.35.0    target-cpu x86-64
  -triple aarch64-unknown-linux5.10.0-gnu2.17.0   target-cpu generic
  ```

So different targets/glibc pins key differently — no wrong-artifact risk. And no `CACHE_KEY_VERSION` bump is needed here, since zigcc invocations previously fell to CLI mode and were never cached: this only adds recognition, it doesn't change any existing key.

Your point about `-E` probing being the durable answer is a good one and I agree — a basename allowlist is a list someone has to remember to update forever. Filed it as #535 with credit to you, including the `version_line` first-line detail above, in case you or anyone wants to pick it up.

lwintermelon:
@emmanuelm41 Ah, looks like this got merged!

The cargo-zigbuild does have a `zigcxx` stem, see
https://github.com/rust-cross/cargo-zigbuild/blob/ffdcd2fb3f038cf4bba90084a0d2e74adb9a8ace/src/zig.rs#L1905-L1907

And yeah the docs needs to be updated.

One more thing I missed: `zig cc` itself has a more fine-grained local cache(yeah, it's a compiler, not just a wrapper), will it affect how kache works? I'm wondering if multi-layer local caching might add unnecessary overhead, since kache knows nothing about cache policy of `zig cc`.

Issue #399, opened 2026-06-20
Out-of-tree build keys do not converge on Windows (relocate misses)


jleni:
## Problem

Out-of-tree builds (build tree outside the source tree, e.g. a sibling `CARGO_TARGET_DIR` or a symlinked build forest) converge across a relocate on Linux and macOS but NOT on Windows. The relocated build recompiles instead of hitting, even though the produced artifact is byte-identical. This is a missed hit (wasted recompile), not a wrong artifact.

Found while adding the out-of-tree e2e scenarios (#394): `e2e-rust-out-of-tree-target` and `e2e-rust-symlinked-target` both had to be gated `os = ["linux", "macos"]` because the relocate phase fails on Windows.

## Likely cause

The path that diverges is the `OUT_DIR` env-dep (and similar build-script paths). On Windows the build-tree path looks like `C:\...\source\..\oot-target\release\...` with mixed separators and an unresolved `..` component. The Rust path normalizer's prefix matching does not normalize this the same way at two locations, so the key diverges. The cc fix in #397 (topology-based prefix-map scheme) addressed the C/C++ side on Linux/macOS; the Windows analog for the Rust normalizer is still open.

## Scope

- Rust out-of-tree `OUT_DIR` / `CARGO_MANIFEST_DIR` env-dep normalization on Windows.
- Note: MSVC `cl` keeps path-literal keys by design (a separate track); this is about gnu/clang and the Rust path normalizer on Windows.

## Done when

- The two gated scenarios can drop their `os` gate and converge on Windows, or a Windows-specific scenario asserts convergence.

Follow-up from #304 / #394.

jleni:
### Update: OUT_DIR convergence fixed on Windows; one residual field remains

PR #403 lands the first half of this.

[...]

PR #521, opened 2026-07-16
fix(args): derive target_dir and workspace_root correctly for cross-compilation

Ysh204:
## Summary

Fixes incorrect `target_dir()` derivation for cross-compilation.

When cross-compiling (`cargo build --target <triple>`), Cargo places outputs under:

<workspace>/target/<triple>/<profile>/deps

The current implementation always walks up two parent directories, yielding:

<workspace>/target/<triple>

instead of the logical Cargo target directory:

<workspace>/target

This causes two issues:

1. **Incorrect workspace root derivation**

   `workspace_root()` is derived from `target_dir().parent()`. During cross-compilation it therefore resolves to `<workspace>/target` instead of `<workspace>`, preventing workspace source paths from being normalized consistently.

2. **Incomplete dep-info path normalization**

   Cross-compilation still produces host-built artifacts (such as build scripts and proc-macros) under:

   `<workspace>/target/<profile>/...`

   while target-built artifacts live under:

   `<workspace>/target/<triple>/...`

   Using `<workspace>/target/<triple>` as the rewrite anchor means host-layout paths no longer share the same prefix and therefore cannot be relativized consistently.

## Solution

This PR makes `target_dir()` aware of `--target`.

- Host builds continue to walk up **2** parent directories.
- Cross builds walk up **3** parent directories, stripping the target triple and deriving the logical Cargo target directory (`<workspace>/target`).

With this change:

- Both host-built and target-built outputs share the same rewrite anchor (`<workspace>/target`), allowing dep-info rewriting to normalize both layouts consistently.
- `workspace_root()` continues to derive the correct workspace root (`<workspace>`) from `target_dir()` for both host and cross-compilation.

## Test plan

Verified locally with:

- [x] `cargo fmt --all`
- [x] `cargo test`

Added regression tests covering:

- cross-compilation `target_dir()` derivation
- cross-compilation build-script outputs
- cross-compilation `workspace_root()` derivation

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -431,12 +431,8 @@ impl RustcArgs {
     /// rules for the two consumers and the cache key wouldn't reflect
     /// the actual remap injection.
     pub fn workspace_root(&self) -> Option<PathBuf> {
-        self.out_dir
-            .as_ref()
-            .and_then(|p| p.parent())
-            .and_then(|p| p.parent())
-            .and_then(|p| p.parent())
-            .map(std::path::Path::to_path_buf)
+        self.target_dir()
+            .and_then(|t| t.parent().map(Path::to_path_buf))
     }
 
     /// Derive the cargo target directory (e.g. `<workspace>/target`) from
@@ -463,8 +459,14 @@ impl RustcArgs {
     /// `rustc -o /tmp/prog`), so dep-info rewriting is skipped rather than
     /// anchored to a wrong directory.
     pub fn target_dir(&self) -> Option<PathBuf> {
+        let is_cross = self.target.is_some();
         if let Some(od) = &self.out_dir {
-            return od.parent()?.parent().map(Path::to_path_buf);
+            let mut p = od.parent()?;
+            p = p.parent()?;
+            if is_cross {
+                p = p.parent()?;
+            }
+            return Some(p.to_path_buf());
         }
         let out = self.output.as_deref()?;
         let mut cursor = out.parent();
@@ -472,7 +474,12 @@ impl RustcArgs {
             if let Some(name) = dir.file_name()
                 && (name == "deps" || name == "build")
             {
-                return dir.parent()?.parent().map(Path::to_path_buf);
+                let mut p = dir.parent()?;
+                p = p.parent()?;
+                if is_cross {
+                    p = p.parent()?;
+                }
+                return Some(p.to_path_buf());
             }
             cursor = dir.parent();
         }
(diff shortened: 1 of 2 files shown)


emmanuelm41:
Thanks for this, and for the unusually clear write-up — the diagnosis is right, and I checked the mechanism rather than taking the description's word for it.

**One blocking item before this can land — a `CACHE_KEY_VERSION` bump — plus one test I'd like removed. The fix itself is correct.**

The derivation fix checks out. I logged real rustc argv through a `RUSTC_WRAPPER` on `cargo build --target wasm32-unknown-unknown`: host units (build scripts, proc-macros) get no `--target` and land under `target/debug/…`; only target units carry `--target` and land under `target/<triple>/…`. So `self.target.is_some()` tracks cargo's layout exactly. On those captured inputs:

| invocation | before | after |
|---|---|---|
| host lib | `/ws/target` | `/ws/target` (unchanged) |
| cross target unit | `/ws/target/<triple>` | `/ws/target` ✓ |
| `workspace_root()`, cross | `/ws/target` | `/ws` ✓ |

`just check` is green with your branch merged onto current main.

**Blocking — bump `CACHE_KEY_VERSION` to 22.** `src/cache_key.rs:160` is at `21`, and the contract there is "bump when cache key logic changes in a way that could have produced incorrect entries". Since v17 (#399) the remap prefix set is deliberately *not* folded into the key (only the on/off bit), and sources are hashed by content, not path — so this change moves the dep-info anchor and the injected remap for cross builds **without changing the key**:

```
PathNormalizer::from_env(Some(<ws>/target))  [before] -> aa5bc59fd318c54d…
PathNormalizer::from_env(Some(<ws>))         [after]  -> aa5bc59fd318c54d…
```

Byte-identical. And `rewrite_depinfo_content` (`src/link.rs:818`) is a blind string replace: an entry stored by today's kache relativizes `<ws>/target/<triple>/…` to `__kache_root__/…`, and a build on your code expands that against `<ws>/target` — so the triple silently drops out of every restored `.d` path, under the same key. `v20` and `v21` are the same shape. `CACHE_KEY_VERSION = 22` plus a short `v22:` note in that comment block invalidates the old cross entries cleanly. Nothing you could have known from the diff — it's a repo convention.

**Please drop `test_target_dir_build_script_cross_compiling`.** It builds a state cargo can't produce: it hands a *host* build script `--target`, puts it under the triple, and models `-o` where cargo passes `--out-dir`. It passes, but it pins a fictional layout. The real shape is `target: None, out_dir: "/work/proj/target/debug/build/serde-abc123"`, which returns `/work/proj/target/debug` — which is also why "both host-built and target-built outputs share the same rewrite anchor" isn't quite right: proc-macros and host libs do share `<ws>/target`, build scripts don't. That mismatch is pre-existing (identical in a plain host build on main today), so it isn't yours to fix — but the test shouldn't imply otherwise. Your other two tests cover real states and are good.

Minor, none of it blocking:

- The branch is 18 commits behind main and re-adds `test_parse_double_wrapper_unrecognized_driver` from #506. It merges clean (git deduplicates it), but a rebase drops those ~34 lines and leaves the diff as just your fix.
- `workspace_root()` now returns `Some` for `-o`-only invocations where it returned `None`; and with an external `CARGO_TARGET_DIR`, cross builds now derive the target dir's parent rather than the target dir. Both look unreachable or pre-existing, just noting them as unintended side effects.

I read the code and reproduced cargo's layout locally; I did not push a cross-compiled build through kache end to end.

emmanuelm41:
That's exactly it — thanks for the quick turnaround. The `v22` note reads right, the rebase dropped the #506 duplicate, and removing the build-script test leaves the two that cover real states. `just check` is green on my side at `80f4f7b`.

Nothing further from me.

emmanuelm41 (approved):
Verified at `80f4f7b`: the `v22` key bump closes the false-hit hazard on existing cross entries, the two remaining tests cover states cargo actually produces, and `just check` plus all seven required checks are green.

*(AI-assisted review, approved on a maintainer's instruction.)*

PR #522, opened 2026-07-16
fix(daemon): add `status` subcommand the help text already promised

aqeelat:
## Problem

`kache daemon --help` advertised `status` alongside the real subcommands:

> Daemon management (status, start, stop, install, uninstall, log)

…but no `status` subcommand existed, so `kache daemon status` failed with `unrecognized subcommand 'status'`. Status was only reachable via the bare `kache daemon` invocation, which the help text never mentioned.

## Fix

- Add a `Status` variant to `DaemonCommands` routing to the **existing** `service::status()` — the same function the bare command already uses. No logic duplicated.
- Rewrite the misleading `about` line; clap already enumerates the subcommands below it.
- Document `kache daemon status` as an explicit alias in `docs/commands/reference.mdx`.

Both `kache daemon` and `kache daemon status` now print the status block (service install state, running/offline, socket path, log location, version/epoch).

## Verification

- `just check` passes (fmt + clippy + tests).
- New test `daemon_status_subcommand_succeeds` mirrors the existing `daemon_without_subcommand_succeeds`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,7 @@ See [`scenarios/README.md`](scenarios/README.md) for the scenario format.
 | `kache config` | Open the TUI configuration editor |
 | `kache completions <shell>` | Print shell completion script (bash, zsh, fish, elvish, powershell) |
 | `kache daemon` | Show daemon and service status |
+| `kache daemon status` | Show daemon and service status (explicit form) |
 | `kache daemon run` | Start the persistent background daemon (foreground) |
 | `kache daemon start` | Start daemon in background (returns immediately) |
 | `kache daemon stop` | Stop a running daemon |
--- a/docs/commands/reference.mdx
+++ b/docs/commands/reference.mdx
@@ -361,6 +361,7 @@ kache completions fish | source   # fish
 
 ```sh
 kache daemon                  # show daemon status: service install state, running/offline, socket path, log location, daemon version/epoch
+kache daemon status           # same as above (explicit alias)
 kache daemon start            # start in background, wait until ready
 kache daemon stop             # graceful shutdown
 kache daemon restart          # restart (via launchd/systemd/Task Scheduler if installed, else manual)
--- a/src/main.rs
+++ b/src/main.rs
@@ -186,7 +186,7 @@ enum Commands {
         namespace: Option<String>,
     },
 
-    /// Daemon management (status, start, stop, install, uninstall, log)
+    /// Daemon management. With no subcommand, shows daemon status.
     #[command(subcommand_required = false)]
     Daemon {
         #[command(subcommand)]
@@ -249,6 +249,8 @@ enum Commands {
 
 #[derive(Subcommand)]
 enum DaemonCommands {
+    /// Show daemon status (alias for bare `kache daemon`)
+    Status,
     /// Run the daemon server in the foreground
     Run,
     /// Start daemon in background (returns immediately)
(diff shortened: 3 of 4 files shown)


emmanuelm41:
Thanks for this — and for chasing down the mismatch instead of just working around it. Nothing here blocks merge; two things below, one of which is a real miss.

**Safety:** CLEAR
**Purpose:** ACCOMPLISHES
**Tests:** GAP — the new test asserts exit status only, not the equivalence it documents
**Docs:** REQUIRED & MISSING — `README.md:246-253`

The mechanism checks out end to end. `Status` routes to the same `service::status()` as the `command: None` arm, and nothing between argv and that call distinguishes the two spellings: `detect_log_mode` only special-cases `monitor`/`config` at argv[1], and the wrapper-detection path can't match `["daemon", …]`. I built the branch and diffed both invocations in one isolated env — byte-identical stdout, both exit 0. The consumers that *do* pattern-match the daemon argv (`pgrep -f "kache daemon run"`, the launchd/systemd/Task Scheduler units, `nix/module.nix:192`) all want the `run` form and are correctly untouched.

Rewriting the `about` line rather than repairing its list was the better call — the old one was wrong in both directions, since it also omitted `run` and `restart`.

**The real miss:** `README.md:246-253` has a `## Commands` table enumerating the bare form plus all seven daemon subcommands. It's a complete inventory rather than an illustrative sample, so it's now the one list left in the repo that describes the daemon surface without `status` — the same drift this PR is fixing, one level up. One row:

```
| `kache daemon status` | Show daemon and service status (explicit form) |
```

(`docs/daemon/lifecycle.mdx` and `docs/daemon/overview.mdx` also show daemon commands, but those blocks are deliberately partial — they already omit `install`/`uninstall`. No change needed there.)

**Minor, your call:** `daemon_status_subcommand_succeeds` only asserts `.success()`, so it would still pass if `Status` were wired to some other handler that happens to exit 0. Since the claim is equivalence, comparing the two forms in one env pins the routing:

```rust
let bare = e.cmd().arg("daemon").output().unwrap();
let explicit = e.cmd().args(["daemon", "status"]).output().unwrap();
assert!(bare.status.success());
assert_eq!(explicit.status, bare.status);
assert_eq!(explicit.stdout, bare.stdout);
```

Mergeable as-is if you'd rather leave it.

emmanuelm41:
Both addressed — thanks for the quick turnaround.

**Safety:** CLEAR (re-ran on the new head)
**Purpose:** ACCOMPLISHES
**Tests:** COVERED
**Docs:** REQUIRED & PRESENT

The test change is the interesting one. I checked it does what it claims rather than just watching it go green: I miswired `Status` to a handler that exits 0 without printing, and the new test fails on it (`left: []` against the status block). The version it replaced passed that same mutation. So it's now pinning the routing, not just the exit code.

Nothing further from me. A maintainer owns the merge call.

aqeelat:
@emmanuelm41 out of curiosity, what are you using for the ai review? is it a specific skill or tool?

emmanuelm41 (approved):
Approving — all four verdicts are clean and CI is green across the seven required checks.

For the record: `status` routes to the same `service::status()` as the bare form, and that routing is now pinned by a test that fails on a miswiring rather than one that only checks the exit code. `README.md` and the command reference both enumerate the subcommand. Nothing outstanding.

Thanks again for the fix and the quick turnaround.

*(AI-assisted review; approved by a maintainer.)*

Issue #527, opened 2026-07-16
test(events): concurrent rotation test asserts an invariant rotation doesn't guarantee (flake risk)


emmanuelm41:
Follow-up from a post-merge review of #518 (merged as `023162d`).

## What

`test_concurrent_log_append_and_rotate` asserts the tailer observed **every** id `0..=200` (`src/events.rs:1223`). Rotation is explicitly allowed to discard events no poll ever saw, so this asserts an invariant the design does not promise — the tailer is not lossless by contract.

## The margin is thin (measured)

- a serialized `BuildEvent` in this test is **511 bytes**
- the test's cap is **4000 bytes** → only **~7 records** fit under it
- the appender sleeps **3ms** → an event survives roughly **21ms** before rotation discards it
- the tailer polls every **2ms**

So any **>21ms** stall of the tailer thread drops an event and reddens CI for a reason that isn't a bug. kache's Windows/macOS CI runs on shared self-hosted runners, where a 21ms stall is plausible.

## Status: latent, not observed

It passed **25/25** locally unloaded — I could not force a failure, so this is a risk from the arithmetic, not a reproduced flake.

## Corroborating tell

`polled_ids.insert(0)` (`src/events.rs:1154`) pre-seeds id 0, which makes the later id-0 assertion vacuous. That reads as a deliberate guard against exactly this race (the seed event is written before the tailer thread constructs its `EventTailer`, and can be rotated away before the first poll) — i.e. the shaky invariant was already being worked around. **Removing the insert would add flake, not remove it** — the right move is to exclude 0 from the required set or comment why it's pre-seeded.

## Suggested fix

Assert the **real** invariant: no *gaps* in what the tailer observed — which the test already does for the file itself at `src/events.rs:1216` — rather than losslessness. Alternatively widen the cap or slow the rotator so the retention window isn't ~21ms.

## Related: a surviving mutant

Deleting `self.position = 0` from the rename-detection branch **still passes** `test_event_tailer_handles_rename_rotation` (verified by mutation). The replacement file (3 lines) is shorter than the tailer's old position (5 lines), so the pre-existing `file_len < position` fallback resets the position anyway. Only a *lagging* tailer — where the replacement is larger than the old byte offset — would catch it. Worth constructing that case if this test is touched.

[...]

PR #531, opened 2026-07-17
refactor(store,events): extract shared atomic replace helper

### LINKED ISSUE #525: refactor(store,events): extract shared atomic-replace helper (temp + fsync + rename + Windows retry)
## Problem

`src/events.rs` and `src/store.rs` now each carry their own copy of the same atomic-publish machinery. #518 (log rotation) added, inline in `events.rs`:

- a unique temp-file name,
- `sync_all()` on the temp file,
- `fs::rename()` to publish,
- a Windows transient-rename retry keyed on raw OS errors **5 (Access Denied)** and **32 (Sharing Violation)**, with a `(1 << attempt.min(6)).min(64)` backoff over 10 attempts,
- a parent-directory `fsync` on Unix.

`store.rs` already has all five (`fsync_file`, `fsync_dir`, unique temp naming, `commit_blob_rename` / `is_transient_rename_error` / `rename_backoff_ms`) — with the *same* errnos and the *same* backoff formula.

That Windows-errno behaviour was earned through real debugging (#441). It now lives in two places, and they will drift. The next atomic write (a third log type, a config or manifest write) becomes a third copy.

## Two conventions already diverge

1. **Temp naming.** `store.rs` deliberately uses pid **+ nonce** and documents why; `events.rs` uses pid only. It's safe there only because the flock serialises rotators — an implicit invariant nobody will remember.
2. **Temp GC policy.** `store.rs` explicitly never GCs in-progress `.tmp` files (there's a test pinning that). `events.rs` added a janitor that deletes `<log>.tmp.*` older than 5 minutes. Two temp-GC policies in one codebase is the kind of thing that bites during the next "why did my file disappear" incident.

Note the janitor isn't redundant — it does clean temps left by *crashed* processes (a review pass initially called it redundant and that was wrong; under the lock any visible temp is orphaned, but crash debris is real). It's the **5-minute age heuristic** that's arbitrary and inconsistent with the store's policy.

## Proposal

Extract a shared helper (e.g. `src/fsutil.rs` or `src/atomic.rs`):

- `atomic_replace(path, bytes) -> Result<()>` — temp + fsync + rename + retry + dir-fsync,
- the transient-rename predicate and backoff, in one place,
- one temp-naming convention (prefer the store's pid+nonce),
- one documented temp-GC policy.

Then have both `store.rs` and `events.rs` call it. Net effect: the #441 Windows knowledge lives once, and the next atomic write reuses it instead of copy-pasting.

## Context

Fell out of the review of #518 (merged). Non-blocking there — the code is correct and CI-green on all platforms; this is purely about not letting the duplication calcify.

Related: #441 (Windows rename errnos), #510 (store concurrency).

### LINKED ISSUE #526: fix(events): log rotation drops the log file's mode (temp+rename replaces the inode)
Follow-up from a post-merge review of #518 (merged as `023162d`).

## What

`rotate_log_impl` now writes a temp file and renames it over the log. The temp is created with `OpenOptions::new().create(true)` under the **process umask** (`src/events.rs:432`), and `fs::rename` swaps it in — so the original file's mode is discarded. The pre-#518 `fs::write` truncated the *same inode*, which preserved it.

## Evidence

Probed both sides with the same fixture (`chmod 600 events.jsonl`, then `rotate_if_needed`):

| branch | before | after |
|---|---|---|
| `86e7b9c` (pre-#518) | `0600` | `0600` — preserved |
| `023162d` (main) | `0600` | `0644` — **widened** |

## Why it matters

Rotation silently widens permissions on a hand-tightened log, or when writers run under different umasks (e.g. a systemd-started daemon at `022` vs a shell at `077`). On Windows the analogue is an explicit DACL on the log being replaced by the temp file's inherited directory ACL.

## Impact: narrow

Nothing in kache ever sets a non-default mode — `log_event` creates the file with plain append+create, and rotation runs in the same user's wrapper processes under the same umask, so in the normal lifecycle the original and temp modes match. This needs a manual `chmod` or mixed-umask writers to bite. Low urgency, but it's a real behavior change vs pre-#518.

## Fix

Stat the original before the rename and carry its permissions onto the temp file (`fs::set_permissions`) — roughly a one-liner.

Natural home: **#525** (extract the shared atomic-replace helper). The helper should preserve mode for both `store.rs` and `events.rs` so the two can't drift on this.

Ysh204:
## Motivation

`src/store.rs` and `src/events.rs` each implemented their own version of the same atomic file replacement logic:

- unique temporary file creation
- writing and `sync_all()`
- Windows transient rename retries with exponential backoff
- atomic `rename()`
- parent directory `fsync` on Unix

This duplicated the platform-specific behavior introduced for the Windows rename fixes and made the implementations susceptible to drifting over time.

## Solution

Extract the shared implementation into a new `src/atomic.rs` module.

The new helper provides:

- `atomic_write_and_replace(...)`
  - unique `.<file>.<pid>.<nonce>.tmp` naming
  - callback-based writing
  - `sync_all()` before publish
  - destination permission preservation before replacement
  - Windows transient rename retry with exponential backoff
  - optional concurrent-writer handling
  - parent directory `fsync` on Unix

- `atomic_replace(...)` convenience wrapper for byte slices.
- `cleanup_temp_files(...)` for removing stale temporary files using the shared naming convention.

`store.rs` now uses `atomic_write_and_replace(...)` for blob materialization, while `events.rs` uses `atomic_replace(...)` during log rotation. This removes the duplicated atomic publish implementation and centralizes the platform-specific behavior in one place.

## Verification

- [x] `cargo fmt --all --check`
- [x] `cargo test`

Added unit tests covering:

- atomic replacement
- destination permission preservation
- temporary file cleanup

Closes #525.
Closes #526.

### DIFF
--- a/src/atomic.rs
+++ b/src/atomic.rs
@@ -1,0 +1,297 @@
+use anyhow::{Context, Result};
+use std::fs;
+use std::path::Path;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+static ATOMIC_TMP_NONCE: AtomicU64 = AtomicU64::new(0);
+
+/// fsync a directory so a rename/create into it survives a crash. Without it,
+/// a file's contents can be durable while its directory entry is not. No-op on non-Unix:
+/// Windows has no directory-handle fsync and `File::open` on a directory fails.
+#[cfg(unix)]
+fn fsync_dir(dir: &Path) -> std::io::Result<()> {
+    fs::File::open(dir)?.sync_all()
+}
+#[cfg(not(unix))]
+fn fsync_dir(_dir: &Path) -> std::io::Result<()> {
+    Ok(())
+}
+
+/// fsync a file so its bytes are durable on disk.
+/// On Windows, FlushFileBuffers requires write access.
+pub(crate) fn fsync_file(path: &Path) -> std::io::Result<()> {
+    #[cfg(windows)]
+    let file = fs::OpenOptions::new().write(true).open(path)?;
+    #[cfg(not(windows))]
+    let file = fs::File::open(path)?;
+    file.sync_all()
+}
+
+/// Whether a failed rename is a transient Windows state worth retrying.
+/// ERROR_ACCESS_DENIED (5) and ERROR_SHARING_VIOLATION (32) surface when a
+/// concurrent put/remove leaves the destination delete-pending or open; both
+/// clear on their own. No such transient rename error exists off Windows.
+fn is_transient_rename_error(e: &std::io::Error) -> bool {
+    #[cfg(windows)]
+    {
+        matches!(e.raw_os_error(), Some(5) | Some(32))
+    }
+    #[cfg(not(windows))]
+    {
+        let _ = e;
+        false
+    }
+}
+
+/// Backoff before the next rename retry: 1, 2, 4, … ms capped at 64 ms
+/// (≈0.3 s total across the retry budget), bounding how long a wedged Windows
+/// destination can delay a write.
+fn rename_backoff_ms(attempt: u32) -> u64 {
+    (1u64 << attempt.min(6)).min(64)
+}
+
+/// Remove a file, clearing the read-only attribute first to ensure deletion
+/// succeeds even on Windows when the file was marked read-only.
+fn remove_file_robust(path: &Path) -> std::io::Result<()> {
+    if let Ok(meta) = fs::metadata(path) {
+        let mut perms = meta.permissions();
+        if perms.readonly() {
+            perms.set_readonly(false);
+            let _ = fs::set_permissions(path, perms);
+        }
+    }
+    fs::remove_file(path)
+}
+
+/// Perform an atomic file replacement. This creates a temporary file beside the
+/// destination path, executes `write_fn` to populate it, flushes it (`fsync_file`),
+/// copies permissions from the original destination (if it exists) to preserve file
+/// modes, and atomic renames it with transient error retries on Windows.
+///
+/// Returns `Ok(true)` if the rename succeeded.
+/// If `allow_concurrent_winner` is `true`, a concurrent writer winning the race to create
+/// the file is treated as a benign lost race and returns `Ok(false)`. If `allow_concurrent_winner`
+/// is `false`, it retries or errors out on conflicts.
+pub(crate) fn atomic_write_and_replace<F>(
+    dest_path: &Path,
+    allow_concurrent_winner: bool,
+    write_fn: F,
+) -> Result<bool>
+where
+    F: FnOnce(&Path) -> Result<()>,
+{
+    let nonce = ATOMIC_TMP_NONCE.fetch_add(1, Ordering::Relaxed);
+    let pid = std::process::id();
+    let file_name = dest_path
+        .file_name()
+        .and_then(|n| n.to_str())
+        .unwrap_or("file");
+    let temp_path = dest_path.with_file_name(format!(".{file_name}.{pid}.{nonce}.tmp"));
+
+    let res = (|| -> Result<()> {
+        write_fn(&temp_path)?;
+        fsync_file(&temp_path).context("flushing temp file")?;
+        Ok(())
+    })();
+
+    if let Err(e) = res {
+        let _ = remove_file_robust(&temp_path);
+        return Err(e);
+    }
+
+    // Preserve the destination's permissions so atomic replacement does not widen them.
+    // Setting permissions is best-effort (e.g. read-only mounts or filesystem limitations).
+    if let Ok(meta) = fs::metadata(dest_path) {
+        let _ = fs::set_permissions(&temp_path, meta.permissions());
+    }
+
+    const MAX_ATTEMPTS: u32 = 10;
+    let mut last_err = None;
+    for attempt in 0..MAX_ATTEMPTS {
+        match fs::rename(&temp_path, dest_path) {
+            Ok(()) => {
+                if let Some(parent) = dest_path.parent() {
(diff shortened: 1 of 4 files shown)


emmanuelm41:
Thanks for this — the duplication between `store.rs` and `events.rs` was real, and #525/#526 are worth closing. The extraction gets the subtle part right, but two things before it can land: it **doesn't compile on Linux or macOS**, and the second commit belongs to #532 rather than here.

**What I verified and liked.** The blob temp naming survives exactly: `blob_path_in_store_dir` makes `blob.file_name()` *be* the 64-hex hash, so `.{file_name}.{pid}.{nonce}.tmp` is byte-identical to the old `blob_tmp_path`, and the orphan sweep's `is_blob_hash_name` still excludes temps. That's the thing this refactor could most easily have broken silently, and it didn't. The `allow_concurrent_winner` flag also maps faithfully onto both call sites — winner-check-before-`is_dir`-check for the store, straight to the transient check for events. And cleaning the temp when `write_fn` fails is better than the old store path, which leaked a partial temp when `fs::copy` failed.

---

### Blocking: the crate doesn't build off Windows

`src/atomic.rs:2` imports `OpenOptions` unconditionally, but it's only used in the `#[cfg(windows)]` arm of `fsync_file`. `Cargo.toml:160` sets `[lints.rust] warnings = "deny"`, so on any non-Windows target this is a hard compile error, not a lint nit:

```
error: unused import: `OpenOptions`
 --> src/atomic.rs:2:21
```

`cargo clippy --workspace -- -D warnings` exits 101 for me on macOS — the same command that's ticked in the PR body. I suspect it genuinely passed for you because you're on Windows, where the import *is* used; this is invisible from there. I've approved CI on this PR so you'll see it go red on the Linux jobs directly.

The fix is what the original `store.rs` did deliberately — qualify it inside the cfg arm rather than importing it:

```rust
use std::fs::{self};
...
#[cfg(windows)]
let file = fs::OpenOptions::new().write(true).open(path)?;
```

I applied exactly that locally and `cargo clippy --workspace --all-targets -- -D warnings` goes to 0.

### Blocking: the test commit is #532's

Commit 2 here (`test(events): harden concurrent rotation test and cover rename-reset mutation`) is the same change as your #532 — I diffed the resulting test bodies and they're identical apart from `actual_polled.last()` vs indexing and `2 + 10` vs `12`. Since both PRs are open against `main` and both rewrite the same two tests in `src/events.rs`, whichever lands second will conflict.

Could you drop that commit from this PR and keep #531 a pure refactor? The tests then get settled in one place, on #532, where #527 is the context. Nothing wrong with the change itself — it just shouldn't be in two PRs.

### Worth fixing: the new permission copy leaks temps on Windows

`atomic.rs:91` is new behavior neither original had:

```rust
if let Ok(meta) = fs::metadata(dest_path) {
    let _ = fs::set_permissions(&temp_path, meta.permissions());
}
```

On Windows the read-only attribute blocks deletion, which makes this reachable:

1. A and B both pass `blob.is_file()` while the blob is absent.
2. A publishes and `set_blob_readonly` marks the destination read-only.
3. B copies A's now-read-only permissions onto **B's own temp**.
4. B's rename fails; `allow_concurrent_winner` sees the destination and hits `let _ = fs::remove_file(&temp_path)` — which now fails, silently. The temp is orphaned, and nothing sweeps blob temps.

Same shape on the rotation side: a read-only log makes `cleanup_temp_files` unable to delete its own leftovers, since `atomic.rs:155` calls `remove_file` without clearing the attribute. Worth noting the codebase already knows this trap — `unlink_blob` (`store.rs:46`) clears read-only before unlinking, and `Cargo.toml:164` carries `permissions_set_readonly_false = "allow"` for exactly that idiom. Either scope the permission copy to Unix, or clear read-only before the removes.

### Tests: a regression guard went missing

Commit 1 deletes `fsync_file_flushes_a_writable_blob` and `fsync_dir_succeeds_on_a_real_directory` from `store.rs` without porting them to `atomic.rs`. The first one is the named regression guard for #196 — the read-only-handle bug that `fsync_file`'s Windows arm exists to prevent. It should move with the code it guards. `rename_backoff_ms_caps_at_64` and `is_transient_rename_error_*` made the trip, so this looks like an oversight rather than a decision.

Its explanatory comment got left behind too — `store.rs:2072` still carries the "#196 regression guard" block, now sitting above `materialize_blob_errors_when_source_cannot_be_copied`, which it doesn't describe.

### Minor

- Retry exhaustion drops its context: the old blob path surfaced `"atomic rename of blob after retries"`, `atomic.rs:124` now returns the raw OS error.
- Old-style `events.log.tmp.<pid>` temps left by a pre-upgrade crash won't match the new cleanup pattern, so they'll linger forever. Cheap to accept, just worth a deliberate call.

I read the code and reproduced the build failure locally on macOS; I did not test on Windows, so the read-only reasoning above is from the code and the platform's documented semantics, not an observed run.

emmanuelm41:
Everything's addressed — thank you for the quick turnaround. The cfg-scoped import, the ported #196 guards, and reverting the test commit to keep this a pure refactor all look right, and `remove_file_robust` closes the Windows read-only temp leak cleanly (nice touch applying it to `cleanup_temp_files` too, not just the rename paths).

Verified locally on macOS: `fmt`, `clippy -D warnings`, and the full test suite all pass, and both fsync guards run. Re-approving CI now so the Windows and Linux jobs can confirm across platforms. From my side this is clean across all four checks — a maintainer owns the merge call.

emmanuelm41:
@Ysh204 thanks for the PR. Approving and squashing once CI goes green again!

emmanuelm41 (approved):

PR #532, opened 2026-07-17
test(events): harden concurrent rotation tests and cover rename-reset mutation

### LINKED ISSUE #527: test(events): concurrent rotation test asserts an invariant rotation doesn't guarantee (flake risk)

Ysh204:
### 1. Flake risk in concurrent rotation test

`test_concurrent_log_append_and_rotate` asserted that the tailer observed every event ID from `0..=200`. However, the event tailer is intentionally **not lossless**: log rotation may discard events that were never polled. Under scheduler delays, the tailer can legitimately miss events that have already been rotated away, causing the test to assert a stronger guarantee than the implementation provides.

### 2. Surviving mutation in rename rotation test

`test_event_tailer_handles_rename_rotation` wrote only a single event after rotation. Because the replacement log was smaller than the tailer's previous byte offset, the existing `file_len < position` fallback reset the cursor even if the rename-detection reset (`self.position = 0`) was removed, allowing the mutation to survive.

### Harden `test_concurrent_log_append_and_rotate`

- Assert the documented delivery contract by verifying that the events observed by the tailer form a gap-free sequence (excluding the intentionally pre-seeded ID `0`), rather than requiring every emitted event to be observed.
- Increase the log size limit to **65,536 bytes**, the retained-line count to **100**, and the rotation interval to **20 ms** to reduce scheduler-dependent flakiness while preserving the behavior under test.

### Harden `test_event_tailer_handles_rename_rotation`

- After rotation, write enough events (10) so the replacement log is larger than the tailer's previous byte offset.
- This ensures the test exercises the rename-detection reset path even if the `file_len < position` fallback is removed, preventing the previously surviving mutation.

## Verification

Verified locally:

```bash
cargo test events::tests
```

Closes #527.


emmanuelm41 (commented):
Thanks for taking this on — and the rename-test half is genuinely good work, I verified it end to end.

**Blocking on one thing:** the concurrent test no longer fails when the bug it was written to catch comes back. Everything after that is minor.

**What checks out.** Your mutation analysis is exactly right, and I reproduced both halves of it: dropping `self.position = 0` from the rotation-detection branch leaves the *old* `test_event_tailer_handles_rename_rotation` green, and kills it under your version (`left: 7, right: 12` — the tailer resumes at the stale 5-event offset and reads 7 of 12). The 2-kept + 10-new arithmetic holds; `rotate_log_impl`'s size-cap re-check doesn't trim below 2 at 1500 bytes. That half is ready.

**The blocking finding.** With #518's rotation-detection block deleted from `poll()`, `test_concurrent_log_append_and_rotate` **passes** on this branch. On `main` it fails with `EventTailer missed event ID 8` — and that assertion was the test's whole reason to exist as a regression test. The gap-free rewrite doesn't replace it, because gap-freeness is vacuous here: after `fs::rename`, an undetected rotation leaves the tailer's handle on the unlinked inode, whose length is frozen. Appenders reopen by path, so the tailer reads a contiguous prefix and then silence forever — and silence has no gaps. (`file_len < position` doesn't rescue it either; the old inode was never truncated.)

The fix is one line, and it's deterministic rather than another race. The final drain runs after the appender and rotator have stopped, and event 200 is the last line written, so it always survives rotation — a stranded tailer can never reach it:

```rust
assert_eq!(end_id, 200, "EventTailer must observe the final event");
```

Measured: fails `left: 128, right: 200` with the detection reverted (128 is exactly where the first rotation lands), and passes 15/15 with it intact.

**Minor, take or leave:**

- `polled_ids.insert(0)` (`src/events.rs:1159`) is now vestigial — line 1233 filters `id > 0` straight back out. Drop the seed, or keep it and drop the filter.
- The widened cap/interval is the real flake fix here: it moves the tolerated tailer stall from ~21ms to ~90ms. Worth knowing gap-freeness still isn't a contract, just a much wider margin — a good change, not a proof.
- "the documented delivery contract" — it isn't actually documented anywhere I could find. A sentence on `EventTailer::poll` noting rotation may discard unpolled events would make the next reader's life easier. Optional.

**What I checked:** ran both tests and `just check` locally on macOS (green), and ran the mutations above rather than reasoning about them. I didn't test on Windows, where the detection path uses file identity instead of dev/ino.

emmanuelm41 (commented):
That's exactly it — all three addressed, and I re-verified the important one against your actual code, not just the diff.

`assert_eq!(end_id, 200)` restores the regression coverage: with #518's rotation-detection block reverted, the concurrent test now fails `left: 129, right: 200` (the tailer strands on the rotated-away inode and never reaches the final event), and it passes 20/20 with detection intact, so it's deterministic rather than a swapped-in flake. The seed/filter cleanup and the `poll()` doc note are both good. `just check` green locally on macOS.

Nothing left from my side. Thanks for the quick turnaround.

emmanuelm41 (approved):

PR #534, opened 2026-07-17
fix(compile): probe cargo outputs instead of readdir'ing out-dir

Skory:
## Summary

- Replace full `read_dir` of cargo `--out-dir` / `-o` parent directories in `pre_clean_outputs` and `discover_output_files` with O(candidates) path probes.
- Shared `ARTIFACT_SUFFIXES` covers common cargo/rustc products (rlib/rmeta/d/dylib/bin/…) plus Windows import-lib sidecars (`dll.lib` / `dll.exp` / `dll.a`) and special emit kinds (`ll`/`bc`/`mir`/`s`/`asm`).
- Fixes severe **sys CPU** thrash when many parallel `kache rustc` wrappers scan huge monorepo `target/*/deps` trees (~800k entries → tens of seconds of pure readdir each).

Cargo post-compile discovery still prefers rustc JSON artifact notifications; candidate probing is the fallback and the always-on pre-clean path.

## Why

On a large monorepo, sampling showed wrappers stuck in `__getdirentries64` during pre-clean/store, while the daemon itself was nearly idle. Full directory scans do not scale with `split-debuginfo=unpacked` and large deps dirs.

## Test plan

- [x] `just check` (fmt + clippy -D warnings + workspace tests)
- [x] Unit tests for exact candidate set, Windows dll sidecars, emit products, incomplete identity no-op, noise isolation, structural `include_str!` guard that production `compile.rs` has no `read_dir`
- [ ] CI green on this PR

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -60,6 +60,11 @@ const IGNORED_ATTACHED_PREFIXES: &[&str] = &[
 /// Boolean diagnostics / query flags (no value) that must not reach the key.
 const IGNORED_BOOL_FLAGS: &[&str] = &["-v", "--verbose", "-V", "--version", "-h", "--help"];
 
+/// Cargo/rustc artifact basename stem: `{crate_name}{extra_filename}`.
+pub fn format_crate_output_stem(crate_name: &str, extra_filename: &str) -> String {
+    format!("{crate_name}{extra_filename}")
+}
+
 /// Parsed rustc invocation arguments relevant to caching.
 #[derive(Debug, Clone, Default)]
 pub struct RustcArgs {
@@ -512,12 +517,13 @@ impl RustcArgs {
                 .is_some_and(|source| source.starts_with(build_script_out_dir))
     }
 
-    /// Get the output filename stem (crate name + extra filename).
-    #[allow(dead_code)]
+    /// Output filename stem (`crate_name` + optional `extra_filename`).
+    #[cfg_attr(not(test), allow(dead_code))]
     pub fn output_stem(&self) -> Option<String> {
-        let name = self.crate_name.as_ref()?;
-        let extra = self.extra_filename.as_deref().unwrap_or("");
-        Some(format!("{name}{extra}"))
+        Some(format_crate_output_stem(
+            self.crate_name.as_ref()?,
+            self.extra_filename.as_deref().unwrap_or(""),
+        ))
     }
 
     /// Whether this compilation has coverage instrumentation enabled (-C instrument-coverage).
@@ -789,6 +795,13 @@ mod tests {
             parsed.is_user_facing_executable(),
             "--test must count as user-facing"
         );
+    }
+
+    #[test]
+    fn test_format_crate_output_stem() {
+        assert_eq!(format_crate_output_stem("serde", "-9f2a1b"), "serde-9f2a1b");
+        assert_eq!(format_crate_output_stem("serde", ""), "serde");
+        assert_eq!(format_crate_output_stem("app", "-abc"), "app-abc");
     }
 
     #[test]
--- a/src/compile.rs
+++ b/src/compile.rs
@@ -112,7 +112,7 @@ pub fn run_rustc(
     // Discover output files. rustc's own `artifact` JSON notifications
     // are authoritative — when cargo invokes rustc it passes
     // `--error-format=json --json=artifacts`, so rustc reports every
-    // file it writes by exact path. Directory scanning is the fallback
+    // file it writes by exact path. Candidate path probing is the fallback
     // for invocations that don't emit those messages (a bare `rustc`
     // without `--json=artifacts`), where filename guessing is the best
     // available signal.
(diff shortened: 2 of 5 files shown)


emmanuelm41:
Thanks for this — the diagnosis holds up. A full `read_dir` of an 800k-entry `deps/` per wrapper invocation is a real problem, and your profiling reads true.

One blocking issue. `ARTIFACT_SUFFIXES` (`src/compile.rs:258`) only ever probes `{stem}.{suffix}`, but at `codegen-units > 1` with `--emit=obj|llvm-ir|asm|llvm-bc`, rustc writes per-CGU files named `{crate}{extra}.{crate}.{hash}-cgu.N.rcgu.{o,ll,s,bc}` and reports each one through `--json=artifacts` — so kache stores them and restores them 0444. `pre_clean_outputs` can't remove them anymore, which is the exact EACCES the function exists to prevent. The old `starts_with("{prefix}.")` caught them.

It isn't a near-miss. The CGU hash derives from crate name + `-C metadata`, not source content — it's byte-identical across an edited source, so the next miss rewrites precisely the restored filenames. Two tests pinned it: plant read-only `foo-abc.foo.<hash>-cgu.0.rcgu.o` in an out-dir, call `pre_clean_outputs(None, dir, "foo", "-abc")`, assert it's gone. Passes on `main`, fails here. Same root cause, narrower: `--emit=dep-info=out/foo.custom` reports `foo.custom` verbatim — the old code matched it, a closed list can't.

`just check` stays green because `out_dir_candidate_filenames_exact_ordered_set` hardcodes all 45 candidate names — it locks the implementation rather than the invariant. `pre_clean_out_dir_removes_all_common_readonly_candidates` is the natural home for a real per-CGU name.

The good news: your perf win survives untouched. `--emit=link,metadata,dep-info` (cargo build) and `metadata,dep-info` (cargo check) produce no per-CGU files at all, even at `codegen-units=6` — I checked. So the scan you're killing was only ever on the hot path for emit sets that can't produce these. Keeping a directory scan **only** when `--emit` includes `obj`/`llvm-ir`/`asm`/`llvm-bc` costs nothing you're winning. Enumeration isn't an option — the CGU index and hash are unbounded. How you structure it is yours.

Minor, non-blocking: `production_pre_clean_and_discover_do_not_call_read_dir` is fail-open. `split("#[cfg(test)]").next()` truncates at the *first* occurrence, so any earlier `#[cfg(test)]` silently shrinks what's checked — I added a test-only helper plus a real `read_dir` below it and the guard still passed. Also, `dwp` isn't in the list; I couldn't test it (macOS host, no ELF target), so treat as unverified, but packed split-DWARF should produce `{stem}.dwp`, which the old code matched.

What I checked: read the diff and surrounding code, ran the repro against both `main` and this branch, ran `just check` on the branch (green), and reproduced the rustc naming locally with rustc 1.90. I didn't run CI.

Skory:
Thanks for the careful review — agreed this was a real EACCES regression for `--emit=obj|llvm-ir|asm|llvm-bc` with per-CGU names.

Addressed in the latest commit:

1. **Gated prefix scan** — `emit_may_produce_per_cgu_files` detects those emit kinds (including `kind=path` forms). Only then does `for_each_prefix_match_in_dir` run a `starts_with("{crate}{extra}.")` / `lib…` filter. Cargo's usual `link,metadata,dep-info` path stays pure candidate probes (perf win intact).
2. **Regression tests** — plant readonly `foo-abc.foo.<hash>-cgu.0.rcgu.o`; with `emit=[obj]` pre_clean removes it; with cargo-typical emit it is left alone (no readdir).
3. **`dwp`** added to `ARTIFACT_SUFFIXES`.
4. **Structural guard** — split at `\n#[cfg(test)]\nmod tests {` and assert `read_dir` only lives inside `for_each_prefix_match_in_dir`.

`just check` green locally.

emmanuelm41:
Re-reviewed the new commit — this is the right fix, and I verified it end to end. My repro (read-only per-CGU `.rcgu.o` planted, then `pre_clean` with `emit=[obj]`) that failed on the first push now passes, `emit_may_produce_per_cgu_files` correctly catches the `kind=path` forms via the `=` split, and the cargo-typical `link,metadata,dep-info` path stays pure candidate-probes with no `read_dir` — perf win intact. The two new tests lock the *invariant* (removed when emit is per-CGU, left alone otherwise), not just the candidate list, which was the gap. Structural guard splitting at the real `mod tests {` closes the fail-open too. `just check` green here.

One residual, non-blocking and your call whether it's worth covering: `--emit=dep-info=out/foo.custom` still slips through — `dep-info` isn't a per-CGU kind so the gate stays off, and the closed list doesn't probe `.custom`, so a restored read-only `foo.custom` wouldn't be pre-cleaned. It's a hand-rolled bare-rustc shape cargo never emits, so I wouldn't hold the PR for it — noting it for completeness.

emmanuelm41 (approved):

jleni:
thanks @Skory !

PR #542, opened 2026-07-18
Make the daemon socket path configurable

### LINKED ISSUE #539: daemon: socket path can exceed platform limits under long cache paths
## Human context

AI summary pretty much nails it, I was like "why isn't kache helping" and then realized it was failing to use the daemon mode due to long-paths in my environment. 

Will contribute a fix that makes daemon socket path configurable via env var.

### Problem

The daemon socket is derived from `<cache_dir>/daemon.sock`. On platforms with short Unix-domain socket limits, a sufficiently long cache path can make daemon startup fail with an invalid-path error.

### Reproduction

```sh
export KACHE_CACHE_DIR=/a/sufficiently/long/nested/worktree/path/to/kache-cache
kache daemon start
```

### Expected behaviour

The daemon socket path can be configured independently of the cache directory.

scottopell:
### Summary

Make the daemon socket path configurable independently of the cache directory while preserving `<cache_dir>/daemon.sock` as the safe default.

Closes #539.

### Changed

- Added `KACHE_SOCKET_PATH` as an operational configuration override, resolved once during `Config::load` so later ambient environment changes cannot redirect a running configuration.
- Revalidated manually constructed configurations at use time.
- Rejected empty, relative, root, directory, regular-file, symlink, malformed, and inaccessible override targets; unusable values warn and fall back to the default instead of panicking or deleting an operator-named file.
- Kept socket placement independent from cache storage ownership and documented Unix path limits, private-directory requirements, service-manager propagation, and Windows named-pipe behavior.

### Validation

- `just check`
- `cargo test --bin kache config::tests::`
- Independent review of the rebased final diff

### Checklist

- [x] Snapshot and fallback behavior are covered
- [x] Unsafe target types and invalid paths are covered
- [x] Existing default behavior is preserved


src/config.rs
```diff
@@ -854,7 +854,9 @@ impl Config {
     }
 
     pub fn socket_path(&self) -> PathBuf {
-        self.cache_dir.join("daemon.sock")
+        std::env::var_os("KACHE_SOCKET_PATH")
```
jleni: **[bug]** `socket_path()` reads `KACHE_SOCKET_PATH` from the process environment on every call instead of resolving it once when `Config` is built. Other path settings (`cache_dir`, etc.) are snapshotted into the struct; this getter silently overrides even manually constructed `Config` values used throughout daemon unit tests (`test_config(dir)` → `config.socket_path()`). If a developer or CI job has `KACHE_SOCKET_PATH` set, those tests bind/connect on the wrong path and flake or fail. The new unit test’s first assertion (`cache_dir.join("daemon.sock")`) is similarly env-sensitive and is not covered by `config_path_lock()`.

**Suggestion:** Add a `socket_path: PathBuf` field on `Config`, resolve it in `Config::load()` (env override or `cache_dir.join("daemon.sock")`), and make `socket_path()` return that field (cloned or by ref). In tests, set the field explicitly when building configs so ambient env cannot interfere. Optionally clear/lock `KACHE_SOCKET_PATH` in the default-path assertion.

jleni: **[bug]** Empty `KACHE_SOCKET_PATH` (set-but-empty) becomes `PathBuf::from("")`. `Path::parent()` on an empty path is `None`, and daemon startup does `socket_path.parent().unwrap()` in several places (`run_daemon` / `server_main` / `start_daemon_background`). That turns a bad env value into a panic rather than a clean config error. A path of `/` has the same `parent() == None` problem. Relative paths also work only when every client and the daemon share the same cwd (systemd/launchd vs wrapper auto-start often do not).

**Suggestion:** When resolving the override, reject empty values (treat as unset or error), require a normal file path with a parent directory, and prefer requiring an absolute path (or absolutize with `std::env::current_dir()` at resolve time and store the absolute result on `Config`). Surface a clear `anyhow` error from load/start instead of panicking later on `unwrap()`.

jleni: **[suggestion]** Socket path security is now user-controlled, but nothing documents or checks placement. The daemon still derives co-located sidecars from the socket path (`with_extension("run.lock"|"lock"|"log")`, `daemon.state.json`). Pointing `KACHE_SOCKET_PATH` at a world-writable location (e.g. `/tmp/kache.sock`) is an attractive short path for #539 but enables classic multi-user UDS issues: pre-bind DoS (attacker listens first → daemon exits “already running”), client connect to a foreign socket, and writable sidecar files beside the socket. Default under a private cache dir was much safer; configurability reintroduces the footgun without guidance.

**Suggestion:** Document a recommended short private path (e.g. `$XDG_RUNTIME_DIR/kache/daemon.sock` or `~/Library/Caches/kache/daemon.sock` with a shallow cache dir) and warn against `/tmp` and other shared directories. Consider validating that the parent directory is owned by the current user and not world-writable (warn or hard-fail), and that the resolved path length stays under the platform UDS limit so the original #539 failure mode cannot be reproduced with a bad override.

docs/getting-started/configuration.mdx
```diff
@@ -111,7 +112,7 @@ key_salt = "v3-toolchain-abc"
 
 With this, kache ignores the `KACHE_*` overrides for **file-backed settings** (cache dir, max size, key salt, fallback, S3/planner settings, the toggles, etc.) and takes the file value (or built-in default) instead. When an ignored override is actually set, kache logs a warning naming it, so the suppression is never silent.
```
jleni, line 168: **[suggestion]** `KACHE_SOCKET_PATH` is intentionally excluded from `ignore_env` (comment + docs). That matches “no file key,” but it also means the exact problem `ignore_env` exists for—a stray machine-global export—still redirects all daemon IPC, including under a pinned config. Unlike `KACHE_DISABLED`, this is not merely a kill switch: it can point clients at a different endpoint while the rest of the config claims to be locked down. The docs group it with bootstrap vars without calling out that IPC endpoint control is security-sensitive.

**Suggestion:** Either (a) add an optional file key (e.g. `cache.socket_path`) so `ignore_env` can pin it, or (b) keep env-only but document explicitly that `ignore_env` does not stop socket redirection and that operators should treat `KACHE_SOCKET_PATH` like a trust boundary. If left env-only, at least mention it in the configuration table description as operational/security-sensitive, not just “Daemon socket path.”

src/config.rs
```diff
@@ -1575,6 +1577,13 @@ mod tests {
             config.socket_path(),
             PathBuf::from("/tmp/kache/daemon.sock")
         );
+
+        let socket = tempfile::tempdir().unwrap().path().join("kache.sock");
```
jleni: **[nit]** The env-mutation portion of `test_config_socket_path` uses manual `set_var` / `remove_var` without a `Drop` guard. Elsewhere in this module (`ScopedVar`, `TestEnvGuard`, salt guards) cleanup is RAII so a failed assert cannot leak `KACHE_SOCKET_PATH` into later tests. The lock is also taken only after the default-path assert, so concurrent tests that later use the same env var are only partially serialized.

**Suggestion:** Use the existing `ScopedVar`-style guard (or a small `KACHE_SOCKET_PATH` guard) under `config_path_lock()` for the whole test body, including the default assertion (with the var ensured unset). Restore the previous value on drop rather than always `remove_var`.

docs/daemon/overview.mdx
```diff
@@ -37,7 +37,7 @@ You'll still get local cache hits from previous builds, but the remote is effect
 
 ## Communication
 
-The wrapper communicates with the daemon over a Unix socket at `<cache_dir>/daemon.sock` (e.g. `~/.cache/kache/daemon.sock` on Linux, `~/Library/Caches/kache/daemon.sock` on macOS; on Windows this becomes a named pipe). The cache dir can be overridden with `KACHE_CACHE_DIR`. RPC calls are synchronous from the wrapper's perspective but non-blocking — the remote check has a 3-second timeout, so even a hung daemon caps the added latency.
+The wrapper communicates with the daemon over a Unix socket at `<cache_dir>/daemon.sock` (e.g. `~/.cache/kache/daemon.sock` on Linux, `~/Library/Caches/kache/daemon.sock` on macOS; on Windows this becomes a named pipe). The cache dir can be overridden with `KACHE_CACHE_DIR`, and the socket path with `KACHE_SOCKET_PATH`. RPC calls are synchronous from the wrapper's perspective but non-blocking — the remote check has a 3-second timeout, so even a hung daemon caps the added latency.
```
jleni, line 44: **[nit]** Docs mention `KACHE_SOCKET_PATH` but not why it exists (UDS length limits / #539) or what a good value looks like. Readers fixing long-path failures may pick `/tmp/...` without realizing sidecars move with the socket and that Windows still only hashes the path into a named pipe (override still works, but “socket path” semantics differ).

**Suggestion:** One short sentence: use a short path on a user-private directory when `cache_dir` is deep; on Windows the value is hashed into `\\.\pipe\kache-daemon-*`. Link or mention the path-length issue so the knobs’s purpose is obvious.

emmanuelm41 (commented):
Noticed this is still a draft — commenting early since it looks close and the notes are small.

This one reads cleanly. `KACHE_SOCKET_PATH` genuinely is operational rather than file-backed, so leaving it out of `IGNORE_ENV_GATED_VARS` is the right call and matches the contract the surrounding docs describe. You updated both doc surfaces and the doc comment on the constant, which is more than most PRs this size do. The test is also mutation-sensitive — revert the env branch in `socket_path()` and it fails, which is exactly what you want.

Three small things on the test:

- It asserts the default *before* acquiring `config_path_lock()` and before clearing `KACHE_SOCKET_PATH`. If a developer or CI runner already has that variable exported, the first assertion fails for reasons unrelated to the change. Take the lock and clear the var at the top.
- It unconditionally `remove_var`s at the end rather than restoring the previous value, and a panic between set and remove leaks the mutation into other tests. `config.rs` already has `TestEnvGuard` for exactly this — it's currently hardcoded to `KACHE_CONFIG`, but generalising it to hold the key would make it reusable here and give you RAII cleanup.
- `tempfile::tempdir().unwrap().path().join("kache.sock")` drops the `TempDir` at the end of that statement, so the path points into an already-deleted directory. Harmless today because `socket_path()` never touches disk, but it'll bite whoever extends this test. Bind the `TempDir` to a variable.

None of that is behavioural. I read the code but didn't run it.

scottopell:
Addressed the three test notes:

- acquire `config_path_lock()` and clear `KACHE_SOCKET_PATH` before asserting the default;
- generalize `TestEnvGuard` to retain the variable name and restore the previous value on drop;
- keep the `TempDir` bound for the duration of the socket-path assertion.

Validation: `cargo test config::tests::test_config_socket_path`, `cargo clippy --all-targets -- -D warnings`, and `git diff --check`.

jleni (commented):
## Summary
Draft PR #542 adds `KACHE_SOCKET_PATH` so the daemon IPC endpoint can be placed independently of `cache_dir` (motivated by #539: UDS path-length limits under long cache paths). The change is small and correctly plumbed through the single `Config::socket_path()` call site used by daemon bind/connect,

## Issue counts by severity

- bugs: 2
- suggestions: 2
- nits: 2

emmanuelm41:
Hey @scottopell — no rush at all, just poking this one. It's honestly in good shape; you sorted the test notes, so really all it needs is a rebase onto `main` (it's drifted a bit behind). Once you've done that and you're happy with it, kick it out of draft and we'll take a look. Thanks for this 🙏

jleni (approved):
Rebased and verified on current main. Configuration snapshotting, fail-closed target validation, docs, and focused safety coverage address the review findings; full local checks pass.

PR #543, opened 2026-07-18
Reuse known hashes after exact artifact restores

### LINKED ISSUE #540: perf: restored artifacts are rehashed despite a known blob digest
### Problem

On a local cache hit, kache materializes an artifact from a content-addressed blob whose verified BLAKE3 digest is already known. A downstream rustc wrapper can then read that restored `.rlib`, `.rmeta`, or proc-macro artifact in full to calculate the same digest for its cache key.

The second read is redundant when restoration did not transform the file contents. It adds disk and memory traffic directly to warm cross-checkout and cross-worktree builds, which are kache's primary reuse path.

### Opportunity

After an exact restore has completed, associate the restored file's fresh fingerprint with the blob's known digest in the existing file-hash memo. If the file later changes, the fingerprint changes and normal hashing resumes. Files that undergo content, signing, or dependency-info transformations must remain excluded.

### Evidence

Using `kache-scenario` with a real Phoenix Rust workload in two worktrees at different absolute paths, the complete three-patch hash-reuse stack preserved:

- native verdict `ok`;
- 203 hits, 0 misses, and 100% weighted hit rate;
- 180/180 stable cross-clone keys;
- zero path leaks;
- 100% APFS reflink restore.

Across three quiet-host paired runs, the stack reduced total key-hash input from a 225.9 MB median to exactly 171.5 MB and improved warm median wall time from 24s to 23s. A focused earlier A/B of restore-time seeding reduced warm hashing by 13.3% and improved warm wall median by 7.1% in that smaller sample.

scottopell:
### Summary

Seed the existing file-hash memo after an exact cache restore so downstream wrappers reuse the blob's verified digest instead of rereading the restored artifact.

Closes #540.

### Changed

- Associated exactly restored files with their already-known content hash and fresh filesystem fingerprint.
- Limited reuse to artifacts with no post-restore transformation plan.
- Preserved fail-closed behaviour: any fingerprint change returns to normal content hashing.
- Added focused coverage proving the memo returns the known digest after materialization.

### Why

A local hit already verifies and materializes a content-addressed blob. Reading the resulting `.rlib`, `.rmeta`, or proc-macro artifact again to calculate the same digest adds avoidable work to warm cross-worktree builds.

### Evidence

- Focused A/B: 13.3% fewer warm bytes hashed; warm wall median improved 7.1% in that sample.
- Three quiet-host native `kache-scenario` pairs for the complete hash-reuse stack: median warm build 24s → 23s and median key-hash input 225.9 MB → 171.5 MB.
- Native verdict `ok`, 203 hits/0 misses, 180/180 stable keys, zero path leaks, and full APFS reflink restore.

### Validation

- `just check`
- Native cross-clone `kache-scenario` benchmark on a real Rust workspace

### Checklist

- [x] Targeted test covers the new behaviour
- [x] Cross-clone correctness and path-leak checks pass
- [x] No transformed artifact reuses a pre-transform hash

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -657,6 +657,16 @@ impl Store {
         self.file_hasher().record_cached(fingerprint, hash);
     }
 
+    /// Associate a file just materialized from a verified blob with that blob's
+    /// known content hash, avoiding a redundant read when it becomes a compiler input.
+    pub fn record_known_file_hash(&self, path: &Path, hash: &str) {
+        if let crate::cache_key::FileHashLookup::NeedsHash(fingerprint) =
+            self.file_hasher().lookup_cached(path)
+        {
+            self.file_hasher().record_cached(&fingerprint, hash);
+        }
+    }
+
     /// Check if a committed entry exists for this cache key.
     pub fn contains(&self, cache_key: &str) -> bool {
         let entry_dir = self.entry_dir(cache_key);
@@ -2186,6 +2196,21 @@ mod tests {
         let unicode = "wörning: ".repeat(20);
         let capped = cap_diagnostics(&unicode, Some(5));
         assert!(std::str::from_utf8(capped.as_bytes()).is_ok());
+    }
+
+    #[test]
+    fn record_known_file_hash_avoids_reading_materialized_blob() {
+        let dir = tempfile::tempdir().unwrap();
+        let store = Store::open(&test_config(dir.path())).unwrap();
+        let artifact = dir.path().join("artifact.rlib");
+        std::fs::write(&artifact, vec![b'x'; 64 * 1024]).unwrap();
+        let expected = crate::cache_key::hash_file(&artifact).unwrap();
+
+        store.record_known_file_hash(&artifact, &expected);
+        assert!(matches!(
+            store.file_hash_lookup(&artifact),
+            crate::cache_key::FileHashLookup::Hit(hash) if hash == expected
+        ));
     }
 
     #[test]
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -1660,6 +1660,10 @@ fn materialize_cached_artifact(
         }
     }
 
+    if plan.is_empty() {
+        store.record_known_file_hash(target_path, &cached_file.hash);
+    }
+
     Ok(())
 }
 


src/wrapper.rs
```diff
@@ -1660,6 +1660,10 @@ fn materialize_cached_artifact(
         }
     }
 
+    if plan.is_empty() {
```
jleni, line 1663: **[suggestion]** Coverage only exercises `Store::record_known_file_hash` in isolation (`src/store.rs` test). Nothing asserts that `materialize_cached_artifact` actually seeds after an exact library restore, or — more importantly for correctness — that transformed kinds do **not** associate `cached_file.hash` with the on-disk target. A regression that seeds on every restore (e.g. dropping the `plan.is_empty()` guard, or moving the call above `apply`) would mis-memoize expanded dep-info or re-signed binaries under the pre-transform digest and only show up as under-keyed / wrong-key cache behavior.

**Suggestion:** Add a focused `materialize_cached_artifact` test (mirroring the existing dep-info restore test) that: 1. Restores an `ArtifactKind::Library` (≥64 KiB) blob and asserts `store.file_hash_lookup(target) == Hit(blob_hash)`. 2. Restores an `ArtifactKind::DepInfo` with a non-identity expand and asserts lookup is **not** `Hit(blob_hash)` (NeedsHash/Uncacheable/Hit-with-different-hash all fine). Optionally also cover `DynamicLibrary` so the Sign exclusion stays locked.

src/store.rs
```diff
@@ -657,6 +657,16 @@ impl Store {
         self.file_hasher().record_cached(fingerprint, hash);
     }
 
+    /// Associate a file just materialized from a verified blob with that blob's
```
jleni, line 660: **[nit]** The new doc comment says the file was “materialized from a **verified** blob.” Under default `KACHE_VERIFY_RESTORES=off`, `Store::get` only size-checks blobs; full content verification is opt-in (`sampled`/`always`). The optimization still inherits the same trust model restore already has (content-addressed path + optional verify), but “verified” overstates the default.

**Suggestion:** Soften wording to something like “content-addressed blob whose digest is already known” (and note that fingerprint changes still fail closed).

src/wrapper.rs
jleni, line 1663: **[nit]** `plan.is_empty()` intentionally excludes all `Sign` kinds (dylib/proc-macro/exec), including platforms where `ensure_binary_loadable` is a no-op (Linux/Windows/macOS non-arm64). That matches #540’s “signing … must remain excluded” design and is correct; it does mean the PR text’s “proc-macro artifact” win is limited to cases that are classified without a Sign plan (mainly `.rlib`/`.rmeta`). Not a correctness defect.

**Suggestion:** Either adjust the PR/issue wording to emphasize rlib/rmeta (primary extern inputs), or a follow-up could seed when the restore path can prove content equality (e.g. no content transform and Sign did not mutate) — harder on macOS arm64 verify-then-sign, so leave for later unless benchmarks show dylib hashing dominating.

emmanuelm41 (commented):
Noticed this is still a draft — commenting early because there's one premise worth checking before you invest more.

First, credit where it's due: `plan.is_empty()` is the right guard and I don't think it was the obvious one. Guarding on `transforms.is_empty()` would have looked equivalent and been wrong — `PostRestoreAction::Sign` is deliberately *not* a content transform (`src/compiler/mod.rs:457-461`) yet `codesign` rewrites the file in place after linking. You'd have poisoned the memo on every signed arm64 artifact. The conservative choice is correct here.

**The premise I'd check is "verified blob."** The doc comment says the file was materialized from a verified blob, but by default it wasn't. `Store::get` validates a blob by **size only** unless `KACHE_VERIFY_RESTORES` is set — the comment at `src/store.rs:733-741` says so directly: *"`off` (default) relies on the size check above."*

That matters because it changes the failure mode rather than just the performance:

- **Today:** a blob corrupted to same-size different content gets restored, then rehashed normally. The artifact is wrong for this build, but the memo and every downstream cache key hold the *true* hash of the bytes on disk.
- **With this change:** the same corruption records `F(corrupt bytes) → H(original)`. The memo now asserts something false, downstream keys are wrong, and the error persists in the memo rather than staying local to one build.

So the fail-closed property in your checklist holds for the case it describes — a fingerprint change does fall back to hashing — but it doesn't cover this one, because the fingerprint is taken *of the corrupt content* and is perfectly self-consistent.

I don't think that's yours to just fix; it's a design question about what `record_known_file_hash` is entitled to assume. Options I can see: require `verify_content` on the restore path that seeds, carry a verified-content token out of `Store::get` so the seeding is conditional on actual verification, or accept the tradeoff explicitly and change the doc comment so it doesn't claim a guarantee the default doesn't provide. @emmanuelm41, worth your call.

**Separately, the test doesn't cover this PR's behaviour.** `record_known_file_hash_avoids_reading_materialized_blob` calls the helper directly and asserts the memo returns the digest. That tests the helper in `store.rs` — but this PR's actual change is the call site in `materialize_cached_artifact`. Delete that call and the test still passes. It also wouldn't notice if the helper reread and hashed the whole file, which is the "avoids reading" property the name claims. An assertion on hash-invocation count, or a restore-path integration test, would have teeth.

**Heads up on stacking:** #544 adds a byte-identical `record_known_file_hash` to the same spot in `store.rs`, with a near-duplicate test. Whichever lands first, the other conflicts, and keeping both definitions won't compile. Worth noting the dependency in one of the descriptions.

The benchmark evidence is unusually good for a PR this size — thank you for that. I read the code and traced the restore path; I didn't run the benchmarks.

jleni (commented):
## Summary
PR #543 seeds the existing file-hash memo after an **exact** cache restore so a later wrapper can reuse the blob’s known whole-file BLAKE3 digest instead of re-reading large restored artifacts (`.rlib` / `.rmeta` / etc.) when building dependent crate keys (#540).

**Correctness verdict: sound.** The

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 2

emmanuelm41:
Hey @scottopell — gentle nudge 👋 whenever you circle back. Two things on this one: a rebase onto `main` (it's wandered a fair way), and the one open design question — the memo gets seeded assuming the restored blob was content-verified, but restore verification is off by default, so a corrupted same-size blob could quietly poison the hash memo. That's more of a "how do we want to handle this" call than a bug for you to chase — happy to hash out the direction with @jleni if/when you want to pick it back up. No pressure either way.

jleni:
We tried finishing this against current `main`, but are closing the draft rather than merging a narrowed version.

`plan.is_empty()` remains the correct mutation guard. The blocker is the “verified blob” premise: with default `KACHE_VERIFY_RESTORES=off`, a same-size blob is checked only by size. Seeding its restored target could persist `fingerprint(corrupt bytes) -> hash(original bytes)`, while normal downstream hashing would record the actual bytes.

Seeding only hits actually content-verified by `always` or selected `sampled` mode is safe, but provides no default-mode benefit. Unconditionally rehashing before seeding only moves the full read earlier and adds work for leaf artifacts.

#544 covers the related safe, default-useful optimization for fresh compiler outputs: it seeds after final ingest/fingerprint changes and excludes dep-info and independent staging inputs.

We’ll leave #540 open for a future design that can carry trusted blob provenance through local and daemon restore without another full read. Thank you—the benchmark work and conservative `plan.is_empty()` guard were both valuable.

PR #544, opened 2026-07-18
Reuse known hashes after storing compiler outputs

### LINKED ISSUE #541: perf: fresh compiler outputs are rehashed by downstream wrappers
### Problem

On a cache miss, `Store::put` reads and BLAKE3-hashes each fresh compiler output before storing it as a content-addressed blob. When that output becomes an `--extern` input to a downstream crate, another wrapper can read the same file in full to rediscover the digest kache just calculated.

This duplicates I/O and hashing along the normal dependency graph of a cold Rust build.

### Opportunity

After hashing a stable compiler output for storage, associate its current file fingerprint with the known digest in the existing file-hash memo. Any later mutation changes the fingerprint and safely falls back to a normal rehash.

### Evidence

A focused Phoenix A/B with fresh caches reduced cold key-hash input from 239.5 MB to 129.6 MB in every paired run. In the upstream-native `kache-scenario` cross-worktree validation, the complete three-patch hash-reuse stack returned verdict `ok`, 100% hit/key stability, zero path leaks, and reduced total hashing by 24% on a quiet host. Cold wall time remained tied at 33s because rustc compilation dominates that phase; the value is deterministic work and I/O elimination rather than a claimed cold wall-time speedup.

scottopell:
### Summary

Seed the existing file-hash memo after storing fresh, stable Rust compiler outputs so downstream wrappers reuse the digest kache already calculated instead of rereading the artifact.

Closes #541.

### Changed

- Seeded memo entries only after final blob materialization and the store transaction commit, so hardlink/readonly fingerprint changes are captured.
- Limited seeding to normal Rust output ingestion; dep-info and independent C/C++ staging paths remain unseeded because those source paths are mutated or disposable.
- Reused the existing fingerprint validation path and documented the compiler-output stability assumption.
- Added store-level coverage proving a real `put` yields a memo hit with zero bytes rehashed, while dep-info and independent puts do not.

### Why

`Store::put` already reads and hashes every fresh output before content-addressed storage. Repeating that full-file read when the output becomes a downstream `--extern` input is redundant I/O along the normal Rust dependency graph.

### Evidence

- Focused fresh-cache A/B: cold key-hash input fell from 239.5 MB to 129.6 MB in every paired run; aggregate key time fell 12.1% in that sample.
- Three quiet-host native `kache-scenario` pairs for the complete hash-reuse stack: median key-hash input 225.9 MB → 171.5 MB with native verdict `ok` and unchanged correctness.
- Cold wall median remained tied at 33s because rustc compilation dominates; this PR claims deterministic work reduction, not a stable cold wall-time speedup.

### Validation

- `just check`
- `cargo test put_records_known_hash_only_for_stable_outputs`
- Native cross-clone `kache-scenario` benchmark on a real Rust workspace

### Checklist

- [x] Product-level test covers post-ingest memo reuse
- [x] Cross-clone correctness and path-leak checks pass
- [x] Mutable/disposable outputs are excluded

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -1240,7 +1240,9 @@ impl Store {
     }
 
     /// Associate a stable file with its already-known content hash, avoiding a
-    /// redundant read when it becomes a compiler input.
+    /// redundant read when it becomes a compiler input. Call this only after
+    /// every store-side operation that may change the file's fingerprint and
+    /// while the compiler-owned output is stable.
     pub fn record_known_file_hash(&self, path: &Path, hash: &str) {
         if let crate::cache_key::FileHashLookup::NeedsHash(fingerprint) =
             self.file_hasher().lookup_cached(path)
@@ -1620,7 +1622,6 @@ impl Store {
         for (source_path, store_name) in output_files {
             let hash = crate::cache_key::hash_file(source_path)?;
             let metadata = fs::metadata(source_path)?;
-            self.record_known_file_hash(source_path, &hash);
             let size = metadata.len();
             let executable = is_executable(&metadata);
             if size == 0 && !zero_byte_is_valid_output(store_name, crate_types) {
@@ -1713,6 +1714,23 @@ impl Store {
             params![cache_key, crate_name, crate_type_str, profile, num_features, total_size as i64, content_hash, compile_time_ms as i64],
         )?;
         tx.commit()?;
+
+        // Both ingest phases may hardlink a compiler output to the immutable
+        // blob and mark the shared inode read-only, changing ctime. Seed only
+        // after the transaction's final rematerialization. Independent (C/C++)
+        // puts use disposable staging paths and deliberately do not seed them.
+        if allow_source_hardlinks {
+            for (file, (source, _)) in meta.files.iter().zip(&sources) {
+                // The Rust wrapper expands dep-info back to absolute paths
+                // immediately after `put`, so it is not stable here.
+                if !matches!(
+                    crate::compiler::classify_by_filename(&file.name),
+                    crate::compiler::ArtifactKind::DepInfo
+                ) {
+                    self.record_known_file_hash(source, &file.hash);
+                }
+            }
+        }
 
         Ok(put_result)
     }
@@ -3635,17 +3653,72 @@ mod tests {
     }
 
     #[test]
-    fn record_known_file_hash_returns_memoized_digest() {
+    fn put_records_known_hash_only_for_stable_outputs() {
         let dir = tempfile::tempdir().unwrap();
         let store = Store::open(&test_config(dir.path())).unwrap();
         let artifact = dir.path().join("artifact.rlib");
         std::fs::write(&artifact, vec![b'x'; 64 * 1024]).unwrap();
         let expected = crate::cache_key::hash_file(&artifact).unwrap();
 
-        store.record_known_file_hash(&artifact, &expected);
+        store
+            .put(
+                "known-hash-stable",
+                "artifact",
+                &["rlib".to_string()],
+                &[],
+                "host",
+                "dev",
+                &[(artifact.clone(), "libartifact.rlib".to_string())],
+                "",
+                "",
+            )
+            .unwrap();
+        let hasher = store.file_hasher();
+        assert_eq!(hasher.hash(&artifact).unwrap(), expected);
+        let stats = hasher.stats();
+        assert_eq!(stats.cache_hits, 1);
+        assert_eq!(stats.cache_misses, 0);
+        assert_eq!(stats.bytes_hashed, 0);
+
+        let dep_info = dir.path().join("artifact.d");
+        std::fs::write(&dep_info, vec![b'd'; 64 * 1024]).unwrap();
+        store
+            .put(
+                "known-hash-dep-info",
+                "artifact",
+                &[],
+                &[],
+                "host",
+                "dev",
+                &[(dep_info.clone(), "artifact.d".to_string())],
+                "",
+                "",
+            )
+            .unwrap();
         assert!(matches!(
-            store.file_hash_lookup(&artifact),
-            crate::cache_key::FileHashLookup::Hit(hash) if hash == expected
+            store.file_hash_lookup(&dep_info),
+            crate::cache_key::FileHashLookup::NeedsHash(_)
+        ));
+
+        let independent = dir.path().join("independent.o");
+        std::fs::write(&independent, vec![b'o'; 64 * 1024]).unwrap();
+        store
+            .put_with_compile_time_independent(
+                "known-hash-independent",
+                "artifact.c",
+                &[],
+                &[],
+                "host",
+                "dev",
+                &[(independent.clone(), "independent.o".to_string())],
+                "",
+                "",
+                1,
+            )
+            .unwrap();
+        assert!(matches!(
+            store.file_hash_lookup(&independent),
+            crate::cache_key::FileHashLookup::NeedsHash(_)
         ));
     }
 
--- a/src/store.rs
+++ b/src/store.rs
@@ -1239,6 +1239,16 @@ impl Store {
         self.file_hasher().record_cached(fingerprint, hash);
     }
 
+    /// Associate a stable file with its already-known content hash, avoiding a
+    /// redundant read when it becomes a compiler input.
+    pub fn record_known_file_hash(&self, path: &Path, hash: &str) {
+        if let crate::cache_key::FileHashLookup::NeedsHash(fingerprint) =
+            self.file_hasher().lookup_cached(path)
+        {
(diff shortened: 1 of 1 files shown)


jleni:
you are on fire! thanks for the contributions! :)

src/store.rs
```diff
@@ -929,6 +939,7 @@ impl Store {
         for (source_path, store_name) in output_files {
             let hash = crate::cache_key::hash_file(source_path)?;
             let metadata = fs::metadata(source_path)?;
+            self.record_known_file_hash(source_path, &hash);
```
jleni: **[suggestion]** `record_known_file_hash` runs **before** `materialize_blob` in the phase-1 put loop. For hardlink-eligible artifacts (`.rlib` / `.rmeta` / `.o` / … — the exact files that later become `--extern` inputs), `materialize_blob` may: 1. `hard_link` the source into the blob temp (POSIX updates inode **ctime** when link count changes), and 2. apply `set_blob_readonly` on that shared inode (permission change updates **ctime** again). The memo key includes `ctime_ns`. After hardlink ingest, a subsequent `FileHasher::hash` / `lookup_cached` on the build output sees a new ctime → `NeedsHash` → full re-read. On CoW filesystems where reflink wins first (APFS / btrfs / XFS-with-reflink), the source inode is untouched and the optimization works — which matches the PR’s reported A/B numbers. On common Linux CI/dev volumes that fall back to hardlink (e.g. ext4 without reflink), the seed is often dead on arrival: correctness is preserved, but the cold-path bytes-hashed reduction largely evaporates for the primary rlib graph. Phase-2 rematerialize (blob reclaimed between phase 1 and the txn) can also hardlink+RO the source *after* the current record site.

**Suggestion:** Seed the memo from the **post-ingest** fingerprint — ideally after phase-1 `materialize_blob` returns, or once after phase-2 finishes for each `(source_path, hash)` so the stored fingerprint matches the file clients will actually `stat`. Content is unchanged by hardlink/RO, so the blake3 already computed remains valid. Consider a small put-integration assertion on a hardlink-forced path (or a unit test that hardlinks+chmods then checks lookup) so ext4-style ingest doesn’t silently lose the win.

src/store.rs
```diff
@@ -657,6 +657,16 @@ impl Store {
         self.file_hasher().record_cached(fingerprint, hash);
     }
 
+    /// Associate a stable file with its already-known content hash, avoiding a
+    /// redundant read when it becomes a compiler input.
+    pub fn record_known_file_hash(&self, path: &Path, hash: &str) {
```
jleni, line 1246: **[suggestion]** `record_known_file_hash` associates a hash that was computed **before** the fingerprint is taken: ```text hash_file(path) → hashH lookup_cached(path) → NeedsHash(fp_now) // re-stat record_cached(fp_now, hashH) ``` The established `FileHasher::hash` order is fingerprint-first, then hash, then record under the **pre-hash** fingerprint. That fails closed when metadata changes mid-window (lookup misses). Hash-then-fingerprint can fail **open** in a narrow TOCTOU: if the file is replaced with different content *and* new metadata between `hash_file` and `lookup_cached`, the memo stores the old digest under the **new** fingerprint and a later Hit returns a wrong content hash (cache under-keying for dependents). Concurrent mutation of a just-produced compiler output during `put` is pathological (key lock + normal cargo scheduling), so practical risk is low — but the PR text claiming fully fail-closed mutation handling is slightly stronger than the mid-`put` window guarantees, and the order is inconsistent with the rest of the hasher.

**Suggestion:** Prefer aligning with `FileHasher::hash`: build/capture fingerprint (or reuse the `metadata` already read for size/executable), then only `record_cached` if you still trust that the hash belongs to that fingerprint (e.g. fingerprint-first around the existing `hash_file`, or re-stat and skip recording on metadata mismatch with the size used for hashing). At minimum, document that the seed assumes the output is stable for the duration of `put`.

src/store.rs
```diff
@@ -2188,6 +2199,21 @@ mod tests {
         assert!(std::str::from_utf8(capped.as_bytes()).is_ok());
     }
 
+    #[test]
```
jleni, line 3655: **[nit]** `record_known_file_hash_returns_memoized_digest` only exercises the helper in isolation against a hand-written 64KiB file. It does not go through `put` / `put_with_compile_time`, so it cannot catch Issue 1 (hardlink/RO invalidation), path absolute-ization differences vs real artifact paths, or interaction with dep-info rewrite-around-put in the wrapper (relativize → put → expand), which intentionally mutates `.d` after the seed and wastes a doomed memo row.

**Suggestion:** Add a store-level test: `put` a ≥64KiB artifact, then `file_hash_lookup` the same source path and assert `Hit` with the stored blob hash — optionally after forcing the hardlink ingest path if the harness can disable reflink. That locks the product behavior this PR is shipping, not just the private helper.

src/store.rs
jleni: **[nit]** The loop already does `fs::metadata(source_path)?` immediately before `record_known_file_hash`, which re-stats again inside `lookup_cached` → `FileFingerprint::from_path`. Extra syscall per output; harmless at this scale.

**Suggestion:** Optional micro-cleanup later (e.g. a `record_cached` path that accepts an already-built fingerprint) — not worth blocking the PR.

emmanuelm41 (commented):
Noticed this is still a draft — commenting early since there's one concrete thing worth fixing before you take it further.

This is the safer of the two hash-reuse PRs: the digest you seed is computed from the same path on the line immediately above (`hash_file(source_path)`), so unlike #543 there's no question of trusting an unverified blob. Good.

**But dep-info files are rewritten right after you seed them.** `store.put(...)` runs at `src/wrapper.rs:1489`; twelve lines later, `rewrite_depinfo_outputs(&result.artifacts, anchor, DepInfoMode::Expand)` at `:1502` rewrites the same `.d` files on disk from relativized back to absolute paths. So for `.d` outputs you're recording `F(relativized) → H(relativized)` for a file that becomes the expanded form moments later.

In the common case that's self-correcting rather than dangerous — absolute paths are longer, so the size changes, the fingerprint changes, and you fall back to hashing. But it means those entries are pure waste, and the "stable compiler output" precondition in your description isn't actually true for this class of file. The narrow bad case is an expansion that happens to preserve length within one filesystem timestamp tick, where the stale row could survive.

Cheapest fix is to skip the seeding for anything the wrapper post-processes, or move it after the expand step. Either way it's worth being explicit about which output kinds are genuinely stable at the moment you seed, rather than assuming all of them are.

**The test doesn't cover this PR's behaviour.** `record_known_file_hash_returns_memoized_digest` calls the helper directly and asserts the memo returns the digest — but this PR's change is the call inside `Store::put`'s loop. Remove that line and the test still passes. It also can't distinguish "returned from the memo" from "reread and rehashed," which is the property the PR is actually about. Asserting on the number of file reads would give it teeth.

One thing I looked for and didn't find a problem with: I don't think the hash/fingerprint window is a practical concern here, since the compiler has exited by the time `put` runs. It's the dep-info rewrite that's the real ordering issue, not concurrency.

**Heads up on stacking:** #543 adds a byte-identical `record_known_file_hash` to the same location in `store.rs`. Whichever lands first, the other conflicts, and two definitions won't compile.

The measurement discipline here is good — particularly calling out that cold wall time is tied and only claiming the work reduction. I read the code; I didn't reproduce the benchmarks.

jleni (commented):
## Summary
Focused, low-risk perf change for [#541](https://github.com/kunobi-ninja/kache/issues/541): after `Store::put_with_compile_time` hashes a compiler output for content-addressed storage, seed the existing `(path, size, mtime, ctime, inode) → hash` memo so a later wrapper hashing that same path as an `

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 2

jleni:
Follow-up to my review: #548 has merged, so ingest now falls back to hardlinking immutable blobs on non-CoW filesystems and the published blob shares the source inode, including the read-only bit. That makes the first suggestion in my review concrete rather than hypothetical: seeding the memo before materialize_blob stores a ctime that the hardlink and chmod immediately invalidate, so the optimization would be dead on ext4-style volumes on current main. Worth seeding from the post-ingest fingerprint when you rebase over the store.rs changes from #548 and #549.

emmanuelm41:
Hey @scottopell — just flagging something ⚠️ before a rebase catches you off guard. `main` shifted under this one: #548 (hardlinking immutable blobs on non-CoW filesystems) landed, and like @jleni mentioned, that means the seed ends up storing a ctime that the hardlink + read-only chmod immediately invalidate — so on ext4-ish volumes it'd basically be a no-op as written. Rebasing over #548/#549 and seeding from the post-ingest fingerprint should sort it. Totally no rush, just didn't want it to surprise you later.

jleni (approved):
Rebased and verified on current main. Post-ingest seeding, mutable-path exclusions, and product-level coverage address the review findings; full local checks pass.

PR #545, opened 2026-07-19
fix(config): expand bare "~" in shellexpand

Ysh204:
## Summary

Fixes `shellexpand` to correctly expand a bare `~` to the user's home directory.

Previously, `shellexpand` only expanded paths beginning with `~/`, while a standalone `~` was treated as a literal path. As a result, configuration values such as `KACHE_CACHE_DIR=~` or a config path set to `~` would not resolve to the user's home directory.

## Changes

- Handle the special case where the input is exactly `"~"`.
- Preserve the existing behavior for `~/...` paths.
- Add a regression test covering bare tilde expansion.

### Before

```rust
shellexpand("~") -> PathBuf::from("~")
shellexpand("~/cache") -> /home/user/cache
```

### After

```rust
shellexpand("~") -> /home/user
shellexpand("~/cache") -> /home/user/cache
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -454,7 +454,7 @@ impl Config {
             .unwrap_or(false);
 
         let cache_dir = env_or_ignored("KACHE_CACHE_DIR", ignore_env)
-            .map(PathBuf::from)
+            .map(|s| shellexpand(&s))
             .or_else(|_| {
                 file_config
                     .as_ref()
@@ -1012,7 +1012,7 @@ const PROJECT_CONFIG_NAME: &str = ".kache.toml";
 /// Priority: `KACHE_CONFIG` env var > nearest `.kache.toml` > XDG user config.
 pub(crate) fn resolve_config_path() -> PathBuf {
     resolve_config_path_from(
-        std::env::var_os("KACHE_CONFIG").map(PathBuf::from),
+        std::env::var("KACHE_CONFIG").ok().map(|s| shellexpand(&s)),
         std::env::current_dir().ok(),
     )
 }
@@ -1082,10 +1082,13 @@ pub(crate) fn config_file_path() -> PathBuf {
 }
 
 fn shellexpand(s: &str) -> PathBuf {
-    if s.starts_with("~/")
-        && let Some(home) = dirs::home_dir()
-    {
-        return home.join(&s[2..]);
+    if let Some(home) = dirs::home_dir() {
+        if s == "~" {
+            return home;
+        }
+        if let Some(stripped) = s.strip_prefix("~/") {
+            return home.join(stripped);
+        }
     }
     PathBuf::from(s)
 }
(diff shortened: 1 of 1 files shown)


emmanuelm41 (commented):
Thanks for this — the helper change itself is right. `~` → home, `~/x` → home/x, `~user` correctly stays literal, and the `home_dir() == None` fallback survives the restructure. I read the code but didn't check out the branch or run it.

One thing to settle before this lands. The summary says this fixes `KACHE_CACHE_DIR=~` and a config path set to `~`, but neither goes through `shellexpand`: `KACHE_CACHE_DIR` is `.map(PathBuf::from)` at `src/config.rs:361-362`, and `KACHE_CONFIG` is the same at `:903`. Only file-config `cache.local_store` (`:370`) and exclude/`extra_inputs` patterns (`:1069`) reach the function you changed. So the described use case is unchanged by this PR. Either route those env vars through `shellexpand` too, or narrow the description — both are fine, but the PR currently claims the first while doing the second. (Worth knowing if you go the first way: `export KACHE_CACHE_DIR=~` in a shell already expands before kache sees it, so the literal-tilde case only shows up via compose files, CI YAML, or `.env` — real, just not the common path.)

A note on the other call site, which I don't think you'd have hit in testing: `extra_inputs = ["~"]` now expands to an absolute home path, and `normalize_pattern` reshapes a real on-disk directory into `<dir>/**/*` (`src/extra_inputs.rs:444`). That turns a previously-inert pattern into a full `$HOME` recursive walk on every compile, and the `walks_filesystem_root` guard won't warn because the literal prefix isn't `/`. Pathological config, so I don't think it blocks — but worth a conscious "yes, that's acceptable" rather than discovering it later.

On tests: the new one covers the helper directly, which is fine. If you touch this again, a case at one of the two real call sites would be worth more than the helper test — that's where the behavior actually reaches a user.

Ysh204:
Thanks @emmanuelm41 for the review and for pointing out the env var resolution sites!

I've updated and squashed the PR (`3144a05`):

1. **Env Var Resolution (`KACHE_CACHE_DIR` & `KACHE_CONFIG`)**:
   - Updated `KACHE_CACHE_DIR` (`src/config.rs:457`) and `KACHE_CONFIG` (`src/config.rs:1015`) to route through `shellexpand`. Now `KACHE_CACHE_DIR=~` and `KACHE_CONFIG=~` actually resolve to the home directory as described.

2. **Call-Site Integration Tests**:
   - Added `test_kache_cache_dir_env_expands_bare_tilde` and `test_kache_config_env_expands_bare_tilde` verifying `Config::load()` and `resolve_config_path()` under env path overrides.

3. **`extra_inputs` Note**:
   - Agreed on `extra_inputs = ["~"]` — expanding to `$HOME` is consistent with `extra_inputs = ["~/"]` and expected behavior for a user explicitly specifying home as an extra input directory.

emmanuelm41 (approved):
Approving. Both findings from the first pass are resolved: `KACHE_CACHE_DIR` and `KACHE_CONFIG` now route through `shellexpand` (so `=~` actually resolves to home), and the two call-site tests genuinely exercise it — I confirmed they fail if the routing is reverted. Bare-tilde handling itself is unchanged and correct.

One small non-blocking note for a future pass: `KACHE_CONFIG` moved from `var_os` to `var().ok()`, which silently drops a non-UTF-8 path value (mostly a Windows edge case) where it was previously honored. Not worth holding this.

Thanks for the thorough follow-up.

PR #546, opened 2026-07-19
fix(wrapper): avoid following symlinks for warning marker files

Ysh204:
## Summary

Prevent `warn_once_per_session` from following symbolic links when accessing warning marker files.

Warning marker files are stored under `std::env::temp_dir()` and are used to deduplicate warning messages across a session. This change adds defensive checks to ensure that symlinked marker paths are not followed when reading or updating marker files.

## Changes

- Detect symlinked marker paths using `std::fs::symlink_metadata`.
- Skip reading timestamp data from symlinked marker files.
- On Unix, open marker files with `O_NOFOLLOW` to prevent following symbolic links during `open`.
- Add a regression test verifying that symlinked marker paths do not modify the target file.

### Before

```text
warn_once_per_session()
    ↓
Open marker file
    ↓
Follow symlinks if present
    ↓
Update marker timestamp
```

### After

```text
warn_once_per_session()
    ↓
Reject symlinked marker paths
    ↓
Open with O_NOFOLLOW (Unix)
    ↓
Update timestamp only for regular marker files
```

### DIFF
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -2211,10 +2211,17 @@ fn maybe_trigger_prefetch(config: &Config, args: &RustcArgs) {
 /// non-regular files up front, and opens with `O_NOFOLLOW` on Unix to prevent
 /// symlink attacks and arbitrary file truncation in shared temporary directories.
 fn open_marker_for_lock(marker: &Path) -> Option<std::fs::File> {
-    if let Ok(metadata) = std::fs::symlink_metadata(marker)
-        && (metadata.file_type().is_symlink() || !metadata.file_type().is_file())
-    {
-        return None;
+    if let Ok(metadata) = std::fs::symlink_metadata(marker) {
+        if metadata.file_type().is_symlink() || !metadata.file_type().is_file() {
+            return None;
+        }
+        #[cfg(windows)]
+        {
+            use std::os::windows::fs::MetadataExt;
+            if (metadata.file_attributes() & 0x400) != 0 {
+                return None; // Refuse reparse points explicitly
+            }
+        }
     }
 
     let mut options = std::fs::OpenOptions::new();
@@ -2237,10 +2244,16 @@ fn open_marker_for_lock(marker: &Path) -> Option<std::fs::File> {
     let file = options.open(marker).ok()?;
 
     // Post-open verification: ensure the opened file handle itself is a regular file.
-    if let Ok(meta) = file.metadata()
-        && !meta.file_type().is_file()
-    {
+    let meta = file.metadata().ok()?;
+    if !meta.file_type().is_file() {
         return None;
+    }
+    #[cfg(windows)]
+    {
+        use std::os::windows::fs::MetadataExt;
+        if (meta.file_attributes() & 0x400) != 0 {
+            return None; // Refuse reparse points explicitly
+        }
     }
 
     Some(file)
@@ -2519,6 +2532,37 @@ mod tests {
         // Verify target file remains completely untouched
         let content = std::fs::read_to_string(&target).unwrap();
         assert_eq!(content, "sensitive target content");
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn maybe_trigger_prefetch_refuses_symlinked_build_session() {
+        let temp = tempfile::TempDir::new().unwrap();
+        let target = temp.path().join("target_file");
+        std::fs::write(&target, "target content").unwrap();
+
+        let marker = temp.path().join(".build-session");
+        std::os::unix::fs::symlink(&target, &marker).unwrap();
+
+        let mut config = test_config(temp.path().to_path_buf());
+        // Enable remote so prefetch actually triggers its path
+        config.remote = Some(crate::config::RemoteConfig {
+            bucket: "test-bucket".to_string(),
+            endpoint: Some("http://localhost".to_string()),
+            region: "us-east-1".to_string(),
+            prefix: "kache/".to_string(),
+            profile: None,
+        });
+
+        // Use dummy args
+        let args = rustc_args(&["rustc", "foo.rs"]);
+
+        // This must NOT modify the target file
+        super::maybe_trigger_prefetch(&config, &args);
+
+        // Verify target file remains completely untouched
+        let content = std::fs::read_to_string(&target).unwrap();
+        assert_eq!(content, "target content");
     }
 
     // ── Opportunistic size-pressure GC (kunobi-ninja/kache#497) ─────────────
--- a/src/wrapper.rs
+++ b/src/wrapper.rs
@@ -132,16 +132,16 @@ pub(crate) fn warn_marker_path(kind: &str, cache_dir: &Path) -> PathBuf {
 ///
 /// Returns whether this call actually emitted the message (for tests).
 pub(crate) fn warn_once_per_session(marker: &Path, session_secs: u64, message: &str) -> bool {
+    if let Ok(metadata) = std::fs::symlink_metadata(marker)
+        && metadata.file_type().is_symlink()
+    {
+        eprintln!("{message}");
+        return true;
+    }
     if marker_is_fresh(marker, session_secs) {
         return false; // already warned this session
     }
-    let Ok(lock_file) = std::fs::OpenOptions::new()
-        .read(true)
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(marker)
-    else {
+    let Some(lock_file) = open_marker_for_lock(marker) else {
         eprintln!("{message}");
         return true;
     };
@@ -2158,12 +2158,7 @@ fn maybe_trigger_prefetch(config: &Config, args: &RustcArgs) {
     // Marker is stale or missing — try to acquire an exclusive lock so only
     // one process does the (expensive) cargo-metadata + daemon RPC.
     let _ = std::fs::create_dir_all(&config.cache_dir);
-    let Ok(lock_file) = std::fs::OpenOptions::new()
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(&marker)
-    else {
+    let Some(lock_file) = open_marker_for_lock(&marker) else {
         return;
     };
     // std::fs::File::try_lock (1.89+) is cross-platform: flock(2) on Unix,
@@ -2174,7 +2169,7 @@ fn maybe_trigger_prefetch(config: &Config, args: &RustcArgs) {
 
     // Re-check under the lock — another process may have updated the marker
     // between our first check and acquiring the lock.
-    if marker_is_fresh(&marker, session_timeout_secs) {
+    if marker_file_is_fresh(&lock_file, session_timeout_secs) {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


src/wrapper.rs
```diff
@@ -2211,6 +2218,11 @@ fn maybe_trigger_prefetch(config: &Config, args: &RustcArgs) {
 
 /// Check if the marker file contains a timestamp within `timeout_secs` of now.
 fn marker_is_fresh(marker: &std::path::Path, timeout_secs: u64) -> bool {
+    if let Ok(metadata) = std::fs::symlink_metadata(marker) {
```
jleni: **[bug]** `marker_is_fresh` now returns `false` for any symlink. That helper is shared with `maybe_trigger_prefetch` (`config.cache_dir.join(".build-session")`), whose `OpenOptions` open is **not** given `O_NOFOLLOW` and still ends in `write_marker_timestamp` (truncate + rewrite). Before this change, a symlink whose target content parsed as a fresh Unix timestamp would short-circuit and never open/write; after, every such marker is treated as stale and proceeds to an unhardened open that follows the link and can wipe the target. So this shared-helper change is incomplete relative to its callers and is a narrow regression for the prefetch path.

**Suggestion:** Either (a) keep the symlink refusal local to `warn_once_per_session` and do not change `marker_is_fresh`, or (b) also open the build-session marker with `O_NOFOLLOW` (Unix) / a no-follow reparse-point flag (Windows) and refuse to write through a symlink there. Prefer a small internal helper (e.g. `open_marker_for_lock`) used by both sites so the policy cannot drift.

src/wrapper.rs
```diff
@@ -132,16 +132,23 @@ pub(crate) fn warn_marker_path(kind: &str, cache_dir: &Path) -> PathBuf {
...
-        .read(true)
-        .write(true)
-        .create(true)
-        .truncate(false)
-        .open(marker)
-    else {
+    let mut options = std::fs::OpenOptions::new();
+    options.read(true).write(true).create(true).truncate(false);
+    #[cfg(unix)]
+    {
```
jleni: **[suggestion]** `O_NOFOLLOW` is correctly applied only under `#[cfg(unix)]`. On Windows the early `symlink_metadata` check is TOCTOU-prone: a reparse point/symlink planted between the check and `open` is still followed by default, and `write_marker_timestamp` will truncate the target. Temp-dir symlink attacks are harder on Windows (privilege / Developer Mode), but the warn markers intentionally live in a shared temp location, so parity would be nicer if feasible.

**Suggestion:** On Windows, open with a no-follow reparse option (e.g. `FILE_FLAG_OPEN_REPARSE_POINT` via `OpenOptionsExt::custom_flags`) and then reject non-regular files, or document explicitly that Windows is best-effort only. At minimum, keep the early check and treat open-of-reparse as failure (warn, do not write).

src/wrapper.rs
```diff
@@ -2439,6 +2451,26 @@ mod tests {
         let _ = std::fs::remove_file(&marker);
     }
```
jleni, line 2498: **[nit]** The Unix regression test is valuable and checks the critical safety property (target content unchanged). It does not cover `marker_is_fresh` independently, dangling symlinks, or the `O_NOFOLLOW` race (file → symlink swap between check and open). Those are optional; the main gap is coverage for the shared-helper / prefetch interaction in Issue 1 if that behavior is kept.

**Suggestion:** If Issue 1 is fixed by hardening both open sites, add a small Unix test that a `.build-session` (or generic marker open helper) symlink is not followed/truncated. Optionally assert the marker path remains a symlink after `warn_once_per_session`.

emmanuelm41 (commented):
Thanks for this — it's a real bug and the fix is the right shape. I reproduced it: reverting your three production hunks and keeping your test makes it fail with the target file's contents replaced by a Unix timestamp. That's the whole vulnerability in one assertion, and your test catches it. Nice.

Worth stating the threat model explicitly in the PR, because it's narrower and sharper than "avoid following symlinks": the marker name is `temp_dir()/kache-{kind}-warn-{blake3(cache_dir)[..16]}`, derived purely from the cache dir, so it's fully predictable for a default install. Combined with `write_marker_timestamp`'s `set_len(0)`, unpatched kache is an arbitrary-file truncate primitive for anyone who can pre-create that name. It's also effectively **Linux-only** — `temp_dir()` is per-user on macOS (`/var/folders/…`, mode 700) and on Windows, so shared `/tmp` is the exposed case. That's still the platform where CI and containers run, so it matters.

**One thing needs fixing before this lands.** `marker_is_fresh` is shared with `maybe_trigger_prefetch`, whose marker is `<cache_dir>/.build-session` — and that caller's open at `src/wrapper.rs:2158` has no `O_NOFOLLOW`. Your new symlink check changes its behavior for the worse:

- Before: `.build-session` symlinked to a file holding a fresh timestamp → `marker_is_fresh` follows it, returns `true` → `maybe_trigger_prefetch` returns early → **no write**
- After: symlink → `marker_is_fresh` returns `false` → falls through to the unprotected open → follows the link → `write_marker_timestamp` truncates the target

I confirmed the pivot directly: same symlink-to-fresh-timestamp fixture, `marker_is_fresh` returns `true` with your hunk reverted and `false` with it applied. The window is narrow — the target has to contain a *fresh* numeric timestamp, since anything else was already overwritten pre-patch — but it's a case your change makes worse rather than better.

Either scope is fine: keep the symlink check local to `warn_once_per_session` and leave the shared helper alone, or apply the same `O_NOFOLLOW` treatment to the `:2158` open so both callers are consistent. The second is better but larger; your call. If you keep the check in the shared helper, its doc comment should say it returns false for a symlink regardless of content — that's now load-bearing for the second caller.

Two smaller notes, neither blocking:

- `O_NOFOLLOW` is `#[cfg(unix)]`, so Windows keeps only the `symlink_metadata` pre-check, which is a TOCTOU race. Defensible given per-user `%TEMP%`, but worth a comment saying that's deliberate rather than an oversight.
- A symlinked marker now warns on every wrapper invocation instead of once — hundreds per build. That's attack-case only and matches the documented *"a duplicated advisory beats a swallowed one"* policy, so I think it's right; just flagging that it's a conscious choice.

One wording point: the PR says marker files are no longer followed, which isn't quite true yet given `.build-session` and the Windows race. Worth narrowing once the above is settled.

I read the code, checked out the branch, and ran the tests and targeted probes locally on macOS; I have not tested on Linux or Windows.

jleni (commented):
## Summary
PR #546 hardens `warn_once_per_session` against classic `/tmp` symlink attacks on warning marker files (`kache-*-warn-<hash>` under `std::env::temp_dir()`). The intended fix is sound for the warn path on Unix: refuse symlinked markers up front, open with `O_NOFOLLOW`, and fail open (emit the advisor

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 1

Ysh204:
Thanks @emmanuelm41 and @jleni for the detailed review!

I've updated and squashed the PR (`50df549`):

1. **Centralized Safe Marker Open (`open_marker_for_lock`)**:
   - Added `open_marker_for_lock` used by both `warn_once_per_session` and `maybe_trigger_prefetch` (`.build-session`).
   - Checks `symlink_metadata` up front, opens with `O_NOFOLLOW` on Unix and `FILE_FLAG_OPEN_REPARSE_POINT` on Windows, and verifies post-open handle metadata.

2. **Fixed `.build-session` Prefetch Path**:
   - `maybe_trigger_prefetch` now uses `open_marker_for_lock(&marker)` so symlinked `.build-session` markers return early without opening or truncating target files.
   - Updated the re-check under lock in `maybe_trigger_prefetch` to use `marker_file_is_fresh(&lock_file, ...)` to read through the locked handle.

3. **Testing**:
   - Added `open_marker_for_lock_refuses_symlink_target` unit test verifying symlink target contents remain untouched.

emmanuelm41 (commented):
This is the right shape — centralizing `open_marker_for_lock` and routing both `warn_once_per_session` and `maybe_trigger_prefetch` through it fixes the `.build-session` regression cleanly. I confirmed the production behavior: a symlinked `.build-session` now hits the helper, which refuses it, so the earlier truncate-the-target path is closed.

**One gap before this is trustworthy: the `.build-session` fix isn't tested.** I reverted *only* the `maybe_trigger_prefetch` call site back to the old raw `OpenOptions` open — the exact regression this PR fixes — and every marker/symlink/build_session test stayed green (14 of them). Nothing exercises that caller. The new `open_marker_for_lock_refuses_symlink_target` test covers the helper in isolation, but a future edit to the call site would sail through. A test that drives `maybe_trigger_prefetch` against a symlinked `.build-session` and asserts the target is untouched would lock it down.

While you're in there, the layered guards inside the helper are also individually untestable — because the up-front `symlink_metadata` check, `O_NOFOLLOW`, and the post-open check are all defense-in-depth, removing any one still passes. Not a correctness problem, just worth knowing the test can't tell you which layer is load-bearing.

Two smaller hardening points:
- The post-open check fails *open*: `if let Ok(meta) = file.metadata()` accepts the file when `metadata()` errors. For a security guard that should be `let meta = file.metadata().ok()?;` (fail closed).
- On Windows, `is_file()` returning false doesn't reject every reparse point — the comment claims broader coverage than the code gives. An explicit `FILE_ATTRIBUTE_REPARSE_POINT` check would match the comment. (The final-component write primitive is already blocked by `FILE_FLAG_OPEN_REPARSE_POINT`, so this is belt-and-suspenders.)

The residual `marker_is_fresh` read-follows-symlink race is read-only — it can suppress a warning but can't truncate anything — so I don't think it blocks.

I read the code and reproduced the test gap locally (reverting the call site, tests still green); I'm on macOS so I couldn't exercise the Windows reparse path.

emmanuelm41 (approved):
Approving. Both points from the last round are resolved and I verified them:

- The `.build-session` fix is now actually tested — `maybe_trigger_prefetch_refuses_symlinked_build_session` drives the real prefetch path, and it fails if the call site is reverted to the unsafe open (I checked; in the prior revision that revert left every test green). So the regression is locked in, not just fixed.
- `open_marker_for_lock` now fails closed on a metadata error (`let meta = file.metadata().ok()?;`).

Centralizing the safe open through one helper used by both the warn path and `.build-session` is the right shape. The only residual is the read-only `marker_is_fresh` symlink-follow race, which can't truncate or overwrite anything — not worth holding.

Nice work across the iterations on this one.

src/wrapper.rs
emmanuelm41: Addressed in the current revision: both `warn_once_per_session` and `maybe_trigger_prefetch` (`.build-session`) now open through the shared `open_marker_for_lock`, which refuses symlinks and opens `O_NOFOLLOW`. I verified the fix has teeth — reverting the `.build-session` call site to the raw open makes `maybe_trigger_prefetch_refuses_symlinked_build_session` fail (target truncated), where the prior revision left it green. Resolving as done. — AI-assisted; a maintainer owns the merge.

src/wrapper.rs
emmanuelm41: Addressed: on Windows the helper now opens with `FILE_FLAG_OPEN_REPARSE_POINT` and verifies the opened handle metadata (fail-closed on error) before any write, so the reparse point is rejected rather than followed — not just the pre-open `symlink_metadata` check. I could not exercise the Windows path from macOS, so flagging it as code-verified rather than run. Resolving. — AI-assisted.

src/wrapper.rs
emmanuelm41: The new `maybe_trigger_prefetch_refuses_symlinked_build_session` integration test covers the critical safety property on the real caller (target content unchanged). The extras you list (independent `marker_is_fresh`, dangling symlinks, the check→open race) remain optional as you noted. Resolving the nit. — AI-assisted.

PR #547, opened 2026-07-19
fix(remote): enforce max_bytes when Content-Length is unavailable

Ysh204:
## Summary

Ensure `RemoteBackend::get` enforces `max_bytes` even when the response does not include a `Content-Length` header.

Previously, the size limit was only checked against the advertised `Content-Length`. If that header was absent (for example with chunked transfer encoding), the response body would be collected without any size validation, bypassing the existing `max_bytes` check.

## Changes

- Validate the size of the collected response body against `max_bytes`.
- Preserve the existing `Content-Length` pre-check when available.
- Add a regression test covering responses without a `Content-Length` header.

### Before

```text
Content-Length present
    ↓
Validate advertised size
    ↓
Collect body

Content-Length absent
    ↓
Collect body
    ↓
No size validation
```

### After

```text
Content-Length present
    ↓
Validate advertised size
    ↓
Collect body
    ↓
Validate collected size

Content-Length absent
    ↓
Collect body
    ↓
Validate collected size
```

## Testing

```bash
cargo test get_without_content_length_over_the_cap_is_refused
```

Output:

```text
running 1 test
test remote_backend::tests::get_without_content_length_over_the_cap_is_refused ... ok

test result: ok. 1 passed; 0 failed
```

## Notes

This change closes the validation gap when `Content-Length` is unavailable. It does not change the collection strategy—the response is still fully collected before validation. A future streaming implementation could enforce the limit while reading the body.

### DIFF
--- a/src/remote_backend.rs
+++ b/src/remote_backend.rs
@@ -40,8 +40,11 @@ pub trait RemoteBackend: Send + Sync {
 
     /// Fetch `key`, or `None` when it is absent.
     ///
-    /// `max_bytes` checks the object's *advertised* size before the body is
-    /// buffered. A remote that advertises no size is not bounded.
+    /// `max_bytes` checks the object's advertised size before the body is
+    /// buffered when `Content-Length` is present, and enforces the cap against
+    /// collected bytes when `Content-Length` is missing or under-reported.
+    /// Note that when `Content-Length` is unavailable, the body is fully
+    /// buffered before post-collect rejection.
     async fn get(&self, key: &str, max_bytes: Option<u64>) -> Result<Option<GetObject>>;
 
     /// Store `body` at `key`.
@@ -63,6 +66,43 @@ pub struct S3Backend {
 impl S3Backend {
     pub fn new(client: aws_sdk_s3::Client, bucket: String) -> Self {
         Self { client, bucket }
+    }
+
+    pub(crate) async fn collect_get_object(
+        resp: aws_sdk_s3::operation::get_object::GetObjectOutput,
+        max_bytes: Option<u64>,
+        describe: &str,
+        request_ms: u64,
+    ) -> Result<Option<GetObject>> {
+        // Refuse before collecting: the point of the cap is to not buffer the
+        // body at all if an advertised Content-Length is present.
+        if let Some(max) = max_bytes
+            && let Some(len) = resp.content_length()
+            && len as u64 > max
+        {
+            anyhow::bail!("{describe} too large: {len} bytes (max {max})");
+        }
+
+        let body_start = std::time::Instant::now();
+        let body = resp
+            .body
+            .collect()
+            .await
+            .with_context(|| format!("reading body of {describe}"))?;
+        let body_ms = body_start.elapsed().as_millis() as u64;
+
+        let bytes = body.into_bytes();
+        if let Some(max) = max_bytes
+            && bytes.len() as u64 > max
+        {
+            anyhow::bail!("{describe} too large: {} bytes (max {max})", bytes.len());
+        }
+
+        Ok(Some(GetObject {
+            body: bytes,
+            request_ms,
+            body_ms,
+        }))
     }
 }
 
@@ -168,28 +208,7 @@ impl RemoteBackend for S3Backend {
         };
         let request_ms = request_start.elapsed().as_millis() as u64;
 
-        // Refuse before collecting: the point of the cap is to not buffer the
-        // body at all.
-        if let Some(max) = max_bytes
-            && let Some(len) = resp.content_length()
-            && len as u64 > max
-        {
-            anyhow::bail!("{} too large: {len} bytes (max {max})", self.describe(key));
-        }
-
-        let body_start = std::time::Instant::now();
-        let body = resp
-            .body
-            .collect()
-            .await
-            .with_context(|| format!("reading body of {}", self.describe(key)))?;
-        let body_ms = body_start.elapsed().as_millis() as u64;
-
-        Ok(Some(GetObject {
-            body: body.into_bytes(),
-            request_ms,
-            body_ms,
-        }))
+        Self::collect_get_object(resp, max_bytes, &self.describe(key), request_ms).await
     }
 
     async fn put(&self, key: &str, body: Vec<u8>, content_type: Option<&str>) -> Result<()> {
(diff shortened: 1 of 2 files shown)


src/remote_backend.rs
```diff
@@ -322,6 +333,27 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn get_without_content_length_over_the_cap_is_refused() {
```
jleni, line 345: **[suggestion]** `get_without_content_length_over_the_cap_is_refused` claims to cover responses without `Content-Length` (e.g. chunked TE) by using `ReplayedEvent::HttpResponse { status: 200, body: ... }`. In `aws-smithy-http-client`’s wire mock, `with_body` is only a thin constructor for the same `HttpResponse` variant, and `generate_response_event` always builds `http::Response` with `Full::new(body)`. Hyper then serves a known-size body and typically emits `Content-Length`, so this path is equivalent to `get_over_the_cap_is_refused` and can pass via the *pre-check* alone—without exercising the new post-collect guard.

**Suggestion:** Either (a) drive a response that truly omits `Content-Length` / uses chunked encoding (custom wire response if the mock gains header control, or a unit-level hook around the post-collect check), or (b) reframe the test to what it can prove today (e.g. under-reported length: advertised CL ≤ max but collected bytes > max) if the mock cannot omit CL. As written, the name/comments over-claim coverage.

emmanuelm41 (commented):
Thanks for this — the gap you're describing is real. `resp.content_length()` is an `Option`, and an S3-compatible remote using chunked encoding can legitimately omit it, so the cap was skippable. Worth fixing. Your check is also slightly better than advertised: because it measures the collected bytes, it catches a remote that *understates* its `Content-Length`, not just one that omits it.

**The blocker is the test: it passes with your production change reverted.** I checked out the branch, ran `get_without_content_length_over_the_cap_is_refused` (passes), then reverted only the `src/remote_backend.rs` hunk, kept the test, and re-ran it — still passes.

The cause is that `ReplayedEvent::with_body(b)` *is* `ReplayedEvent::HttpResponse { status: 200, body: b }` — same constructor, `aws-smithy-http-client-1.1.13/src/test_util/wire.rs:167`. So the new test builds a byte-identical event to `get_over_the_cap_is_refused` right above it. From there the mock serves the body as `Full<Bytes>`, which reports an exact size hint, so hyper synthesizes a `Content-Length` automatically — and the existing pre-check at `src/remote_backend.rs:173` rejects the request before your new code is reached. The existing test's own comment already spells this out: *"The cap is checked against the advertised Content-Length before the body is buffered."*

`WireMockServer` can't express the case you want — `ReplayedEvent` is `#[non_exhaustive]` with no header control. Two options that can:

- **`StaticReplayClient`** (`aws_smithy_http_client::test_util::replay`) — simplest. It takes a `ReplayEvent` built from a full `http::Response<SdkBody>`, so you set the headers yourself and hyper never gets a chance to add `Content-Length`. Wire it in with `.http_client(...)` exactly as `mock_backend` does.
- A **raw `TcpListener`** writing a literal `Transfer-Encoding: chunked` response. More setup, but it exercises real chunked decoding end to end.

Either way, please confirm the new test fails with the production hunk reverted — that's the property the current one is missing.

Two doc comments the change makes inaccurate, both worth updating in the same PR:

- `src/remote_backend.rs:43-44` — *"`max_bytes` checks the object's advertised size before the body is buffered. A remote that advertises no size is not bounded."* The second sentence is exactly what you're fixing.
- `src/remote.rs:152-153` — *"Guards the honest-`Content-Length` case; a remote that lies about its length is already an integrity problem."* Your check partly handles the lying case now.

Last, a design note rather than a request. Post-collection rejection bounds what `get` *returns*, not what it *buffers* — `collect()` has already materialized the whole body by the time you measure it. Your Notes say this, which I appreciate, but the surrounding code currently reads as though the cap is a memory guard (`:171`, *"the point of the cap is to not buffer the body at all"*, and the `MAX_METADATA_BYTES` rationale about concurrent fetches exhausting memory). Enforcing the limit incrementally while draining the `ByteStream` would make that true; I don't think it needs to happen in this PR, but the comments should stop implying it already does.

I read the code, checked out the branch, and ran the tests locally; I have not exercised a real chunked-encoding remote.

jleni (commented):
## Summary
The production change correctly closes the acceptance gap for `max_bytes` when `Content-Length` is missing: after collecting the body, the implementation rejects payloads larger than the cap with the same error shape as the existing advertised-length pre-check. That also fortuitously rejects under-r

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

## Issues outside the diff

These findings reference lines that are not present in the diff and could not be posted as inline comments:

- **[suggestion]** src/remote_backend.rs:44 — `RemoteBackend::get` docs still say `max_bytes` only checks *advertised* size and that “A remote that advertises no size is not bounded.” After this change, a missing/unavailable length is bounded post-collect, so the trait contract is now wrong. Callers reading only the trait will under-estimate what the cap does (and still won’t learn that unbounded buffering can occur before rejection).
  - **Suggestion:** Update the trait docs to match the new semantics, e.g. pre-check advertised length when present; always enforce against collected body length when `max_bytes` is set; note that without an advertised length the body is still fully buffered before rejection (streaming limit is future work).

Ysh204:
Thanks @emmanuelm41 and @jleni for the thorough review!

I've updated the PR to address all feedback:

1. **Test Verification**:
   - Refactored `S3Backend::get` to extract `S3Backend::collect_get_object`.
   - Updated `get_without_content_length_over_the_cap_is_refused` to construct `GetObjectOutput` directly with `resp.content_length() == None`.
   - Verified that `assert!(resp.content_length().is_none())` passes, ensuring the pre-check is skipped and the post-collection guard is exercised. Reverting the production fix causes this test to fail as expected.

2. **Doc Comments**:
   - Updated `RemoteBackend::get` trait docs in `src/remote_backend.rs` to reflect that `max_bytes` pre-checks `Content-Length` when present and enforces the cap post-collection when `Content-Length` is missing or under-reported.
   - Updated `MAX_METADATA_BYTES` comment in `src/remote.rs`.

emmanuelm41 (commented):
Verified — the refactor does the job. Building `GetObjectOutput` directly with no `Content-Length` is exactly the right move (cleaner than the `StaticReplayClient` route I suggested), and the inline `content_length().is_none()` assert guards the premise nicely. I reverted just the post-collect guard locally and the test now fails as it should, so it's got teeth. Both doc comments read accurately now. This is clean from my side.

emmanuelm41 (approved):
Approving. All four verdicts are clean: the cap gap is real and fixed, the test now genuinely exercises the missing-`Content-Length` path (verified by reverting the guard), and both doc comments are accurate. CI is green across all platforms. Thanks for the careful iteration.

src/remote_backend.rs
emmanuelm41: Addressed in `be14f873`. The test no longer goes through the wire mock — it builds a `GetObjectOutput` directly with `.body(...)` and no `content_length`, plus an inline `assert!(resp.content_length().is_none())` guarding exactly the equivalence you flagged (so a future SDK default would fail loudly rather than pass via the pre-check). I confirmed it has teeth: reverting only the post-collect guard makes the test fail. Good catch on the original — resolving as done.

*(AI-assisted — a maintainer still owns the merge.)*

PR #590, opened 2026-07-23
fix(daemon): bound tmutil addexclusion so startup can't wedge behind an active backup

### LINKED ISSUE #588: daemon: startup wedges behind synchronous tmutil addexclusion during an active Time Machine backup
## Problem

`run_server` calls `crate::store::exclude_from_indexing(&config.cache_dir)` synchronously between binding the socket and spawning the accept loop:

- the socket is bound and "daemon listening" is logged;
- `exclude_from_indexing` runs `tmutil addexclusion <cache_dir>` via `Command::output()` and blocks on it;
- only after that does the accept loop start.

`tmutil addexclusion` on a directory that is not yet excluded can block for a long time while a Time Machine backup session is active. Observed today on macOS 15: with `backupd` mid-session, `tmutil addexclusion <fresh dir>` hung past a 5 s timeout, while running it on an already-excluded directory returned quickly.

Result: a daemon started in that window listens (clients connect via the backlog) but never reads a request. Every RPC times out with no error anywhere: the daemon log shows only "daemon listening", `daemon status` says running, and wrappers silently take their fallback paths. First-run setups (`kache init` creating a fresh cache dir) are the likely victims, since an existing cache dir is usually already excluded and returns fast.

Reproduced against both current main (963872e) and older builds; the code is longstanding, the trigger is environmental.

## Suggested fix

Do not let Spotlight/Time Machine housekeeping gate daemon readiness:

- run `exclude_from_indexing` on a detached thread (fire and forget), or
- give the `tmutil` child a hard timeout and kill it on expiry.

The sentinel file half (`.metadata_never_index`) is cheap and can stay synchronous.

Same call sites worth auditing: anywhere else `exclude_from_indexing` runs on a hot or startup path.

jleni:
Fixes #588.

`tmutil addexclusion` on a not-yet-excluded directory can block indefinitely while a Time Machine backup session is active (reproduced today on macOS 15: hung past 5 s with backupd mid-session, while an already-excluded directory returned instantly). The daemon runs it synchronously between binding its socket and starting the accept loop, so a fresh-cache-dir daemon started in that window listens but never serves: clients connect via the backlog and time out, the log shows only "daemon listening", and `daemon status` reports running. First-run setups are the likely victims; existing cache dirs are normally already excluded and return fast. Plausibly also behind macOS e2e runner hangs when a backup coincides.

Change: spawn the tmutil child with stdio nulled and wait at most 2 s, killing it on expiry (`wait_with_timeout`, std-only polling since std has no `wait_timeout`). The exclusion is best-effort housekeeping and must never gate readiness; a killed exclusion is retried on the next startup. The Spotlight sentinel half is untouched.

Testing: new unit test pins the guarantee (quick child passes, hung `sleep 30` child is killed at the deadline); the existing xattr/sentinel tests still pass, so healthy-tmutil behavior is unchanged. Live check: a fresh-cache-dir daemon now serves stats within 3 s of start.

### DIFF
--- a/src/store.rs
+++ b/src/store.rs
@@ -341,7 +341,11 @@ pub(crate) fn exclude_from_indexing(dir: &Path) {
 /// Wait for a child to exit, killing it when the deadline passes. Returns
 /// true if it exited on its own. Polling is the only std-only option (no
 /// `wait_timeout` in std); 25 ms granularity is plenty for a 2 s bound.
-#[cfg(unix)]
+///
+/// Gated to macOS alongside its only caller, [`exclude_from_indexing`] — a
+/// `cfg(unix)` gate would make this dead code on Linux, which `-D warnings`
+/// rejects.
+#[cfg(target_os = "macos")]
 pub(crate) fn wait_with_timeout(mut child: std::process::Child, dur: Duration) -> bool {
     let deadline = std::time::Instant::now() + dur;
     loop {
@@ -4368,7 +4372,7 @@ mod tests {
     /// exact guarantee daemon startup relies on when `tmutil` stalls behind
     /// an active Time Machine backup.
     #[test]
-    #[cfg(unix)]
+    #[cfg(target_os = "macos")]
     fn wait_with_timeout_kills_hung_child_and_passes_quick_child() {
         let quick = std::process::Command::new("sh")
             .args(["-c", "exit 0"])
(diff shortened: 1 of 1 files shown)


jleni:
Duplicate of #589, which is the better fix — closing this one.

#589 does two things this PR doesn't:
- probes the exclusion xattr first, so an already-excluded cache dir (the common warm case) spawns no subprocess at all; this PR pays a `tmutil` fork on every daemon start.
- runs the exclusion on a detached thread, so startup never waits at all; this PR still blocks up to 2s.

The only thing worth carrying over was gating the tmutil path to macOS (a `cfg(unix)` helper is dead code on Linux and `-D warnings` rejects it), and #589 already does that.

Issue #508, opened 2026-07-12
Show not support Cow, while the drive is devdrive(ReFS)


greenhat616:
## What happened

```
kache: this volume (G:\) has no copy-on-write, so cache hits are restored by COPY — the cache and build tree do not share storage, roughly doubling disk for cached content. For zero-copy dedup put the cache + build dir on a ReFS Dev Drive, or set `[cache] windows_hardlink = true` (only if your build never deletes or rewrites an object output).
         affected output: G:\Programs\Rust\clash-nyanpasu\backend\target\debug\deps\tauri_codegen-dea7728f5db67402.d
```

<img width="3843" height="1675" alt="Image" src="https://github.com/user-attachments/assets/bdb4ad35-9b94-4715-9c87-76099ecc90d1" />

```shell
❯ fsutil devdrv query G:
这是受信任的开发人员卷。

开发人员卷可能不受防病毒筛选器保护。

当前没有筛选器附加到此开发人员卷。

❯ fsutil fsinfo refsinfo G:
REFS 卷序列号:                0x4858027d58026a50
REFS 卷版本:                      3.14
REFS 驱动程序受支持的最大版本 :    3.14
数字扇区 :                           0x00000000773a0000
簇总数 :                           0x000000000ee74000
可用簇  :                           0x0000000006fa0853
预留的总数 :                           0x00000000001096a0
每个扇区字节数  :                        512
每个物理扇区字节数 :                4096
每个簇的字节数 :                        4096
快速层数据填充百分比 :           0.0%
慢速层数据填充百分比 :           0.0%
快速层到慢速层速率(簇/秒) : 0
元数据校验和类型 :                   CHECKSUM_TYPE_CRC64
数据校验和类型 :                       CHECKSUM_TYPE_NONE
```

## Expected behaviour

Should copy as Cow, and donot show this message

## Steps to reproduce

Do any builds on DevDrive

## Environment

- `kache --version`: kache 0.9.0
- OS and version (e.g. macOS 15.4, Ubuntu 24.04): Windows 24H2
- Rust toolchain (`rustc --version`): rustc 1.98.0-nightly (4429659e4 2026-06-22)
- Install method (mise / cargo-binstall / cargo install): cargo binstall
- Remote backend (none / S3 / R2 / MinIO / other): none

## Logs

[out.zip](https://github.com/user-attachments/files/29940121/out.zip)

```
```

saschanaz:
Having the same on 0.9.0:

```
[...]

PR #592, opened 2026-07-24
perf(restore): avoid redundant create_dir_all on restore hot path

### LINKED ISSUE #563: perf(restore): avoid redundant create_dir_all and other restore micro-costs
## Problem

`link_to_target` always `create_dir_all` on the parent of every restored artifact. Under cargo layouts the deps dir already exists; this is small per call but multiplies across files × crates.

Part of #552 (P3). Low priority polish after larger key/store wins.

## Proposed solution

1. Try link/reflink first; on `NotFound` for parent, then `create_dir_all` and retry.
2. Or pass a “parent ensured” flag from `restore_from_cache` after one `create_dir_all(output_dir)`.
3. Audit other per-file syscalls in the restore loop for easy consolidation (single platform detect already done).

### Implementation hints

- `src/link.rs` `link_to_target`; rustc/cc restore in `src/wrapper.rs`.
- Preserve behavior when cargo uses nested output paths or custom `--out-dir`.
- Windows path length / RO attribute paths must remain correct.

## Alternatives considered

- Ignore — likely <1 ms per crate; only worth it after P0/P1.

## Use case

Multi-artifact restores (rlib+rmeta+d) at very high QPS once key cost is gone.

Ysh204:
## Summary

Fixes #563.

This PR optimizes the restore hot path by avoiding unconditional `create_dir_all` calls for every restored artifact.

Previously, both `link_to_target` and `write_restored` eagerly called `create_dir_all` before every restore operation, even though Cargo output directories (such as `target/.../deps`) almost always already exist.

This change switches to an optimistic approach:

- Attempt the restore immediately.
- If it succeeds, return without creating any directories.
- If the operation fails with `std::io::ErrorKind::NotFound`, create the destination parent directory and retry exactly once.

This removes unnecessary filesystem syscalls from the common warm-hit path while preserving behavior when restoring into missing directories.

## Changes

- Refactored `link_to_target` to perform optimistic restore with a single retry on `NotFound`.
- Applied the same retry logic to `write_restored`.
- Preserved all existing reflink, hardlink, copy, and permission handling behavior.
- Added regression tests covering:
  - existing parent directory
  - missing parent directory
  - nested directory creation
  - unrelated I/O failures (`PermissionDenied`) without retry loops

### Before

For every restored artifact:

```text
create_dir_all(parent)
restore artifact
```

### After

For every restored artifact:

```text
restore artifact

if NotFound:
    create_dir_all(parent)
    retry once
```

The common case (existing parent directory) now avoids the redundant create_dir_all call entirely.

### DIFF
--- a/src/link.rs
+++ b/src/link.rs
@@ -90,89 +90,106 @@ pub enum LinkStrategy {
 /// Tries reflink first (zero-copy + write isolation on supported filesystems);
 /// falls back to a strategy-specific path on filesystems without CoW.
 pub fn link_to_target(store_path: &Path, target_path: &Path, strategy: LinkStrategy) -> Result<()> {
-    // Ensure parent directory exists
-    if let Some(parent) = target_path.parent() {
-        fs::create_dir_all(parent)
-            .with_context(|| format!("creating parent dir for {}", target_path.display()))?;
-    }
+    let do_link = || -> Result<()> {
+        // Remove existing file at target (link/clone calls fail if dst exists).
+        clear_target(target_path)?;
 
-    // Remove existing file at target (link/clone calls fail if dst exists).
-    clear_target(target_path)?;
+        // Logical size of the artifact, attributed to whichever restoration
+        // mechanism runs below. Best-effort — a metadata failure here must
+        // not fail the restore.
+        let bytes = fs::metadata(store_path).map(|m| m.len()).unwrap_or(0);
 
-    // Logical size of the artifact, attributed to whichever restoration
-    // mechanism runs below. Best-effort — a metadata failure here must
-    // not fail the restore.
-    let bytes = fs::metadata(store_path).map(|m| m.len()).unwrap_or(0);
-
-    // Try reflink first. CoW gives us zero-copy *and* mutations don't
-    // propagate to the cache blob — strictly better than hardlink when
-    // available (APFS, btrfs, XFS-with-reflink).
-    // Keep the failure reason: on Windows it separates "this volume can't
-    // block-clone" from "this one file couldn't be cloned" (#508).
-    let reflink_err = match try_reflink(store_path, target_path) {
-        Ok(()) => {
-            if matches!(strategy, LinkStrategy::Copy) {
-                // Reflink preserves source mode (0o444 for stored blobs);
-                // executables/dylibs need 0o755 so cargo can run/load them.
-                set_executable_perms(target_path)?;
+        // Try reflink first. CoW gives us zero-copy *and* mutations don't
+        // propagate to the cache blob — strictly better than hardlink when
+        // available (APFS, btrfs, XFS-with-reflink).
+        // Keep the failure reason: on Windows it separates "this volume can't
+        // block-clone" from "this one file couldn't be cloned" (#508).
+        let reflink_err = match try_reflink(store_path, target_path) {
+            Ok(()) => {
+                if matches!(strategy, LinkStrategy::Copy) {
+                    // Reflink preserves source mode (0o444 for stored blobs);
+                    // executables/dylibs need 0o755 so cargo can run/load them.
+                    set_executable_perms(target_path)?;
+                }
+                tracing::debug!(
+                    "reflinked {} -> {}",
+                    store_path.display(),
+                    target_path.display()
+                );
+                crate::opcounts::record_reflinked(bytes);
+                return Ok(());
             }
-            tracing::debug!(
-                "reflinked {} -> {}",
-                store_path.display(),
-                target_path.display()
-            );
-            crate::opcounts::record_reflinked(bytes);
-            return Ok(());
-        }
-        Err(e) => e,
-    };
-    #[cfg(not(windows))]
-    let _ = &reflink_err;
+            Err(e) => e,
+        };
+        #[cfg(not(windows))]
+        let _ = &reflink_err;
 
-    // Reflink unsupported on this filesystem — strategy-specific fallback.
-    match strategy {
-        // Windows has no reflink on NTFS, so the Hardlink strategy would
-        // hardlink the read-only store blob. NTFS stores FILE_ATTRIBUTE_READONLY
-        // in the shared MFT record, so EVERY hardlink to a read-only blob is
-        // itself read-only — and Windows refuses to delete or rewrite a
-        // read-only file (WinError 5). A consumer that owns its output and
-        // deletes/rewrites it — e.g. mozbuild's configure `ar_supports_response_files`
-        // conftest (#429) — then breaks. There is no way on NTFS to give a
-        // hardlink a different read-only state than its blob, so restore via an
-        // independent COPY instead: the output is writable and deletable while
-        // the store blob stays read-only (integrity preserved). This mirrors
-        // `write_restored`, already the proven-safe independent-file path, and
-        // costs only working-tree↔store block sharing (LRU is index-based, not
-        // mtime-based, so eviction is unaffected). gnu/clang restores keep
-        // hardlinking — reflink/hardlink there are writable or CoW-isolated.
-        #[cfg(windows)]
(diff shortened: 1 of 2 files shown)


jleni (changes requested):
Thanks for picking this one up, and for the thorough PR description. The optimistic-restore idea is the right shape for #563, and the test coverage you added is more than the issue asked for.

There is one interaction with the Windows restore path that needs restructuring before this can land. It is not something you could have hit locally, so let me lay out the trace.

## The Windows CoW diagnostics run on the doomed first attempt

The retry currently wraps the whole function body, so when the parent directory is missing, the first attempt runs the entire fallback cascade (including the diagnostics) before the directory gets created.

On a ReFS Dev Drive, restoring an artifact into a parent that does not exist yet:

1. `reflink_windows` fails at `OpenOptions::open(dst)` with `NotFound`, because the parent is missing (`src/link.rs:648`).
2. The fallback calls `warn_no_cow_restore_once(..., &reflink_err, ...)`. This happens for `LinkStrategy::Copy`, and for `Hardlink` without the `windows_hardlink` opt-in.
3. `windows_volume_supports_block_clone` asks `GetVolumeInformationW` about the volume root (`src/link.rs:490`), which answers fine whether or not the directory exists, so it returns `Some(true)`.
4. The artifact is larger than one cluster, so `classify_copy_restore` lands on `UnexpectedOnCowVolume` (`src/link.rs:306`).
5. That variant is deliberately not layout-advisory (`src/link.rs:276`), so `[cache] storage_layout_advice = false` cannot mute it. The user sees "the clone FAILED ... This is unexpected; please report it" with `error: The system cannot find the path specified`.
6. It then writes the `cow-fault` session marker (`src/link.rs:261`, `warn_once_per_session`), which suppresses genuine clone-fault reports across every wrapper process for the rest of the warn window.

The restore itself still succeeds on the retry, so functionally the file lands. But the user gets a spurious "please report this" warning, and real clone faults go quiet for the rest of the session. That warning machinery came out of #508, where kache was emitting hundreds of bogus no-copy-on-write warnings at people running perfectly healthy Dev Drives, so we are quite careful about anything that can fire it without cause.

### Two ways forward

The cleaner one is option 2 from the issue: hoist a single `create_dir_all(output_dir)` into `restore_from_cache` (`src/wrapper.rs:1687`) and let `link_to_target` and `write_restored` assume the parent exists. You get the same warm-path saving, with no error classification and no way to reach the diagnostics on a missing-parent path.

If you would rather keep it local to `link.rs`, the retry needs to sit right at the destination-side `NotFound`, before fallback selection and before any warning is emitted, rather than wrapping the whole function.

## Smaller things

- **CI gates.** `cargo fmt --all -- --check` reports two diffs (trailing whitespace around line 1460, and the long `.join()` chain at 1445), and `cargo clippy --workspace --all-targets -- -D warnings` reports four errors (`collapsible_if` and `unnecessary_map_or`, twice each). `just check` runs both locally and will catch them before the push. Your tests themselves pass, 26/26 in `link::`.
- **`let _ = fs::create_dir_all(parent)`** drops the error. The previous code attached `creating parent dir for {}` as context, and losing that means a permission failure on the mkdir surfaces as a confusing second-attempt error instead.
- **`err.chain().any(NotFound)` is broader than intended.** A missing source blob also matches, which is a real case (a concurrent GC can evict a blob mid-restore, see #182). That path now does a pointless mkdir and a full second cascade before failing.
- **`link_to_target_permission_denied_does_not_loop`** will pass locally but fail anywhere the suite runs as root, since 0o555 does not stop root. Worth a skip-guard, or dropping that one.
- **Worth a number.** The issue estimates under 1 ms per crate and files it as P3, so a quick before/after on a warm build would tell us how much this actually buys. Related: the missing-parent path gets meaningfully more expensive under this change (a full failed cascade plus a redo, versus the two syscalls saved per warm restore). Probably fine given cargo pre-creates `deps/`, but good to say out loud.
- The branch is about 29 commits behind `main` now, so a rebase before the next push will keep CI honest.

Happy to look again once the retry moves below the diagnostics. The core change is worth having.

emmanuelm41 (commented):
Thanks — the optimization is the right idea, and the retry tests have real teeth (removing the retry fails `link_to_target_optimistic_missing_parent` and the nested-dir test). Two correctness issues to fix before it lands, plus some scope to trim.

**1. The `create_dir_all` error is discarded, so a real failure surfaces as a misleading "not found."** Both retry sites do `let _ = fs::create_dir_all(parent);`. If creating the parent fails — e.g. a read-only ancestor — that error is dropped, the retry re-runs, and the caller gets the original `NotFound`. I reproduced it: restoring into a `0o555` grandparent surfaces `copying … : No such file or directory (os error 2)` when the truth is permission-denied on the parent. Pre-PR, `create_dir_all(parent).with_context(…)?` surfaced the real cause. Restore that — `create_dir_all(parent).with_context(…)?` before the retry (this also keeps the safe concurrent-dir behavior, since `create_dir_all` treats "already created" as success).

**2. On Windows, `warn_no_cow_restore_once` runs *before* `copy_file`, so a missing parent can emit a false "restored by COPY" warning even when the retry then reflinks successfully.** Move the warn call to *after* a successful `copy_file`, so it only ever describes a copy that actually happened. (The upfront `output_dir` creation hides this for Rust restores, but `restore_cc_from_cache` → `materialize_cached_artifact` can still hit nested CC output paths.)

**Scope:** two hunks are unrelated to restore perf and should be split out — the `remote_backend.rs` boolean rewrite (a De Morgan's transform of the cache-key validation; I checked, it's logically equivalent, but it doesn't belong here) and the `#[allow(clippy::collapsible_if)]` on `open_marker_for_lock`. Both read as clippy cleanup; dropping them keeps this PR reviewable.

Smaller notes, non-blocking: `NotFound` is a loose proxy for "parent missing" — a missing store blob raises the same kind, so you'd get one wasted retry plus a spurious parent dir before it fails correctly. That's fine as a one-shot recovery hint *once* the `create_dir_all` error is preserved. And `restore_from_cache` now creates `output_dir` before validating cached artifact names, so a rejected entry can still have created the output dir — harmless, just an ordering change worth knowing.

Things I checked that are fine: opcounts don't double-count (counters only fire after a successful op, with nothing fallible after), reflink/hardlink/copy fallback and Windows read-only handling are preserved, and same-dir/different-file concurrency is safe.

I read the code and reproduced #1 locally on macOS; #2 is Windows-only, so I'm going off the ordering in the code, not a run.

emmanuelm41 (commented):
Re-verified `bf40e910` — the blockers from both reviews look resolved:

- **Windows diagnostics on the doomed first attempt (@jleni's main concern):** addressed both ways you suggested — `restore_from_cache` now hoists a single `create_dir_all(output_dir)` up front, *and* `warn_no_cow_restore_once` moved below `copy_file(...)?`, so a missing-parent first attempt short-circuits on the `?` before any warning or `cow-fault` marker can fire. The spurious "please report this" + session-wide suppression path is closed.
- **Discarded `create_dir_all` error:** now `create_dir_all(parent).with_context(…)?` at both sites. I reproduced the fix — restoring into a read-only parent now surfaces `creating parent dir for …: permission denied` instead of the misleading `NotFound`.
- **Scope trimmed:** the unrelated `remote_backend.rs` boolean rewrite and the `#[allow(clippy::collapsible_if)]` are both gone.

Still open from @jleni's smaller list, worth a look but not blockers: the `NotFound` chain match still fires on a missing *source* blob (a GC-evicted blob → one wasted mkdir + cascade before it fails correctly), and `link_to_target_permission_denied_does_not_loop` would still pass-as-root without a skip guard. And a rebase is due (branch is well behind `main`).

@jleni — since you asked to look again once the retry moved below the diagnostics, this is ready for your re-review; the change-request is yours to lift.

I read the code and reproduced the error-propagation fix locally on macOS; the Windows path I'm going off the ordering, not a run.

PR #597, opened 2026-07-25
chore(nix): drop flake-utils input

meskill:
## Summary

Remove `flake-utils` input from flake.nix and replace it with inline function to generate output.

The `flake-utils` is outdated and contains deprecated entry for `x86_64-darwin` that breaks newer nixos builds.

## Test plan

My nixos build passes fine with updated nixpkgs from the new branch

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,23 +1,5 @@
 {
   "nodes": {
-    "flake-utils": {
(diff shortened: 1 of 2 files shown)


jleni:
thanks @meskill !

Issue #565, opened 2026-07-20
perf(daemon): decide lifecycle of experimental daemon-assisted local hits


jleni:
## Problem

Each wrapper process reloads config, opens SQLite, may run dep-info, hashes inputs, and writes events. A long-lived daemon already exists for remote/prefetch/hash batching, but **local hits still redo most of that work in-process**.

Part of #552 (P2–architectural). Can combine with thin wrapper (#559) and dep-info memo (#334).

## Proposed solution

Optional fast path (config/env gated initially):

1. Wrapper sends a compact request: rustc/cc argv, cwd, relevant env subset.
2. Daemon computes cache key (with warm dep-info + file-hash caches), looks up store, returns:
   - **Hit**: list of `(store_blob_path, target_path, strategy)` for the wrapper to reflink/hardlink locally, plus cached stdout/stderr; or
   - **Miss**: “compile yourself” (wrapper runs compiler, then `StorePut` / upload enqueue as today).
3. Fallback: if daemon unreachable, keep today’s fully local path (no hard dependency).

### Implementation hints

- Reuse transport + JSON RPC patterns from `HashFiles` / `RemoteCheck` (`src/daemon.rs`, `src/transport.rs`).
- Security: daemon must only write under cache dir + caller-supplied out-dir; reuse `is_safe_artifact_name` and refuse absolute/`..` targets.
- Concurrency: daemon already multi-connection; key computation should release store locks during rustc dep-info and blake3 (same lesson as #281).
- Start with **lookup-only** RPC (key still in wrapper) as an intermediate milestone: still saves SQLite open if daemon returns meta + blob paths — or daemon-only key as second milestone.
- Measure: warm substrate/firefox with daemon on vs off.

## Alternatives considered

- Only thin binary without daemon — helps load, not shared memo.
- Always require daemon — hurts simple local-only UX; keep graceful degrade.

## Use case

Power users and CI with daemon already running for S3; large graphs where per-process fixed cost dominates.

## Concurrency architecture (design review, 2026-07-23)

A deeper design pass (cross-checked across two model families) confirmed the
direction but sharpened the plan. Full write-up: `notes/design/daemon-local-hit.md`.

Key conclusions:

- The daemon's current `OnceLock<Mutex<Store>>` cannot carry this path. A naive
  `LocalLookup` through `with_store` serializes hundreds of wrappers behind one
  connection and can be *slower* than today's independent WAL connections.
[...]

Issue #559, opened 2026-07-20
perf(wrapper): thin local-only wrapper binary or feature-split to cut process weight


jleni:
## Problem

Release `kache` is ~14 MB and links AWS SDK, tokio (multi-thread), TUI, full CLI, etc. Cargo spawns it **once per compilation unit**. Page cache helps, but cold CI and multi-machine builds still pay process image load for features the local hit path never uses.

Part of #552 (P2). Larger architecture cousin of daemon-owned local path.

## Proposed solution

Phased options:

1. **Cargo features**: `default` includes CLI+remote; `wrapper` feature builds a minimal binary (`kache-wrapper` or `kache` with `--features wrapper-only`) without `aws-sdk-s3`, ratatui, planner client, etc.
2. **Two binaries** in the same package/workspace: `kache` (CLI/daemon/service) and `kache-wrapper` (RUSTC_WRAPPER / CC only).
3. Document `RUSTC_WRAPPER=kache-wrapper` and install packaging (Homebrew/apt/binstall) for both.

Local-only hit path needs: args parse, cache key, store open/get/put, link/reflink, events, config. Does **not** need: S3 client construction, TUI, most of `cli.rs`.

### Implementation hints

- Start by measuring `time kache …` no-op / `--version` and `/usr/bin/time -l` or `hyperfine` spawn cost vs a stub.
- Split carefully around `daemon::send_remote_check` / upload enqueue — stub as no-ops when remote feature off.
- Keep a single version string / CACHE_KEY_VERSION across binaries so stores remain compatible.
- CI matrix: build both artifacts; e2e can use the thin wrapper.

## Alternatives considered

- Always use daemon IPC for local hits (see companion issue) — different tradeoff, can combine.
- Dynamic loading of remote plugin — more complex packaging.

## Use case

CI runners and large local workspaces where wrapper spawn+load is measurable against hit restore time.

jleni:
Tracked under warm-hit overhead epic: #552

jleni:
Measurement suggesting this may be the bigger warm-build lever, from the #334 investigation ([verdict](https://github.com/kunobi-ninja/kache/issues/334#issuecomment-5072090656)):

[...]

Issue #334, opened 2026-06-13
feat(cache): local dep-info memo to skip the pre-pass subprocess on warm rebuilds


jleni:
## Problem

`compute_cache_key` runs a fresh `rustc --emit=dep-info` subprocess (`run_dep_info_pass`, `cache_key.rs:1486`) on **every** invocation to enumerate the source closure — pure warm-rebuild latency paid per cached crate, even when nothing changed.

## Fix

An optional local dep-info memo (direct mode):

1. Compute a cheap `direct_hash` = toolchain id + the cacheability-relevant argv subset + content hash of the crate-root source — **no subprocess**.
2. Look it up in a new `dep_info_manifests` table storing the recorded source-file set + env-dep names.
3. Validate by re-hashing the listed files (near-free via the existing file-hash memo — nearly stat-only on a warm tree).
4. Only on a manifest miss/invalidation run the real pre-pass and back-fill.

Store N manifests per `direct_hash` (one crate-root+args legitimately maps to several dep-set states — cfg-gated `include!`, build-script `OUT_DIR` files appearing/disappearing), with access-time LRU.

Pure latency win: the produced key is **byte-identical** to the always-pre-pass path, so **no `CACHE_KEY_VERSION` bump**; gate `off|on`.

## Hard dependency on #323

The memo must **never** persist a manifest derived from a degraded pass (the crate-root-only fallback) — that would poison every future build into crate-root-only keying. Land #323 first; the manifest writer refuses to persist when the pass degraded.

jleni:
Cross-linking: this is the **P0 highest-ROI** item in the warm-hit overhead epic #552.

Other epic children that interact with this work:
- #553 — cheaper `Store::open` (same build storm)
- #557 — env! scan I/O during key construction (after dep-info list is known)
- #565 — longer-term: daemon-owned key path can host the dep-info memo in one process

### Extra implementation hints from the 2026-07 review

1. **Memo key** (sketch): `(rustc_identity, crate_root fingerprint, stable hash of flags that affect dep discovery)`. Flags to include carefully: `--cfg`, features, `--extern` *names* (not paths), edition, target, unstable options that change name resolution. Exclude `--out-dir`, `-o`, incremental, absolute paths that only affect output placement.
2. **Store** next to `file_hashes` in `index.db` or a sidecar table; value = serialized `DepInfo` (source path list + raw env-deps) or a content hash of that structure.
[...]

Issue #319, opened 2026-06-13
Relocatable macOS debug info for cached executables (store-time .dSYM)


jleni:
## Summary

Cached macOS **executables/dylibs** (the `cache_executables` path, default-off) embed Mach-O `N_OSO` debug-map records pointing at the per-build object files (`/…/target/…/<cgu>.rcgu.o`, rlib members under `~/.rustup/…`). Restored at a different path or on another machine, those `.o` paths are gone, so `lldb` can't resolve source-level symbols for the cached binary.

This is the **Mach-O divergence** half of #78. The SDK-path / cache-key half shipped in #318; splitting this out because it's a distinct concern with its own design and only bites when `cache_executables` is enabled.

Verified: a `rustc -g` binary on macOS carries ~15 `N_OSO` entries (`nm -ap <bin> | grep ' OSO '`), one being the build's own `…rcgu.o` in a temp/objdir absent at the restore location.

## State-of-the-art fix: cache a relocatable `.dSYM` (don't mutate the binary)

The macOS "debug map" model (default `-g`, no dsymutil) leaves DWARF in the `.o`s and is inherently non-relocatable. The portable format is a `.dSYM`, consolidated at build time while the `.o`s still exist:

- `dsymutil <bin>` bakes a self-contained `.dSYM` with no `.o` references left. Its source paths are already portable because kache injects `--remap-path-prefix` (they're `<WORKSPACE>`/`<CARGO_HOME>`/… sentinels).
- **`lldb` prefers an adjacent / UUID-matched `.dSYM` over the executable's debug map**, so the stale `N_OSO` entries become inert — no binary mutation. Cross-machine debugging "just works."

### Why not the alternatives
- **Strip the debug map on restore** — lossy: kills source-level debug of every cached binary.
- **`ld64 -oso_prefix` (relativize OSO at link)** — insufficient alone: relative OSO still needs the `.o`s at the restore location, which an executable-only restore lacks.

## Implementation sketch
- Store-time hook on the exec/dylib cache path: run `dsymutil`, add `<bin>.dSYM` to the cache entry's artifact set (macOS-only, via the `Platform` trait — `platform.rs:30` already earmarks a `source_date_epoch`-style method).
- Restore: materialize the `.dSYM` alongside the binary (its own restore action; the binary needs no mutation).
- Gate to macOS + `cache_executables`; no-op on Linux/Windows.
[...]

PR #603, opened 2026-07-25
feat(cache): cache executables by default on Linux

jleni:
Addresses the default-value half of #319.

## The measurement

Warm rebuild of a 330-crate graph (kache's own), `-j1`, same populated cache:

| | wall | passthrough units | passthrough cost |
|---|---|---|---|
| `cache_executables=false` (old default) | **42.3 s** | 55 | 9.0 s |
| `cache_executables=true` | **35.1 s** | 18 | 0.4 s |

**17% faster.** The single final `kache` binary was **5.7 s on the critical path of every warm build** — more than all cache-key computation on that path combined, and larger than the daemon local-hit path (#565), the dep-info memo (#334), and the thin wrapper split (#559) put together. Details in [#319](https://github.com/kunobi-ninja/kache/issues/319#issuecomment-5073605997).

## Why the old default was wrong *as a global*

The documented reason is macOS-specific: a `-g` Mach-O carries `N_OSO` records pointing at per-build `.o` files, so a restored binary can't resolve source-level symbols in `lldb`. #319's own implementation sketch scopes the fix as "macOS-only; no-op on Linux/Windows".

Linux embeds DWARF in the binary under default `-Cdebuginfo` settings, so a restored executable is self-contained and debugs exactly like a freshly linked one. It was paying a macOS debuggability tax for no benefit.

This change makes the default platform-conditional rather than flipping it globally:

| platform | default | why |
|---|---|---|
| **Linux** | **on** | DWARF is inside the binary |
| macOS | off | `N_OSO` → per-build `.o` files (#319) |
| Windows | off | `.exe` records its `.pdb` path — same class of problem |

I left Windows off deliberately. It has the same external-debug-info shape as macOS and I have not investigated it; flipping it would be a guess.

## Correctness

- Restored executables are `LinkStrategy::Copy`, not hardlinks, so a post-build `strip` or code-signing step cannot write back into a store blob. That disposes of the second documented concern ("may be mutated post-build").
- The verify-then-sign restore path (`PostRestoreAction::Sign`) already handles macOS when the flag is turned on explicitly.
- Verified on macOS that a cache-restored binary runs (`kache --version` → `0.11.0`) and passes `codesign -v`.

## What I could not verify locally

**Linux debuggability is argued, not tested end-to-end.** Docker was unavailable on this machine, so I could not step through a restored binary under `gdb`. The argument is that DWARF is in-binary by default, so a byte-identical copy debugs identically — but CI's Linux e2e is the check that actually matters here, and a reviewer with a Linux box may want to confirm `gdb` on a restored executable before merging.

Split-debuginfo (`-Csplit-debuginfo=unpacked`) moves Linux into the same external-reference shape. The sidecars are themselves cached artifacts (`ArtifactKind::DebugSidecar`), and this already applies to rlibs — which have always been cached — so executables aren't special there.

## Also

- The config TUI's default hint now reflects the platform default instead of a hardcoded `false`.
- Docs updated in three places (configuration reference, quick-start, architecture), including the now-false claim that "most of the compile-time savings come from caching rlibs anyway" — the measurement contradicts it.
- Test pins the platform split so it can't drift silently.

Full suite green (1320 unit + all integration).

### DIFF
--- a/docs/getting-started/configuration.mdx
+++ b/docs/getting-started/configuration.mdx
@@ -80,7 +80,7 @@ for AWS SDK-specific cases that need attention.
 | `KACHE_S3_PROFILE` | `cache.remote.profile` | — | AWS credentials profile |
 | `KACHE_S3_ACCESS_KEY` | — | — | Explicit S3 access key |
 | `KACHE_S3_SECRET_KEY` | — | — | Explicit S3 secret key |
-| `KACHE_CACHE_EXECUTABLES` | `cache.cache_executables` | `false` | Also cache user-facing executables (`bin` crates and `--test` binaries). dylib/cdylib/proc-macro are always cached and are unaffected by this flag |
+| `KACHE_CACHE_EXECUTABLES` | `cache.cache_executables` | `true` on Linux, `false` on macOS/Windows | Also cache user-facing executables (`bin` crates and `--test` binaries). dylib/cdylib/proc-macro are always cached and are unaffected by this flag. Defaults on for Linux only, where DWARF is embedded in the binary so a restored executable debugs like a fresh one; macOS (`N_OSO` records) and Windows (`.pdb` path) reference debug info outside the binary, so they stay off pending [#319](https://github.com/kunobi-ninja/kache/issues/319) |
 | `KACHE_CLEAN_INCREMENTAL` | `cache.clean_incremental` | `true` | Auto-clean tracked incremental dirs during GC; active builds also remove the current crate's incremental dir eagerly. Env is on unless the value is exactly `0` or `false` (case-insensitive) |
 | `KACHE_VERIFY_RESTORES` | — | `off` | Re-hash restored blobs before serving hits: `off`, `sampled` (~1/16 hits), or `always`. `1`/`true` map to `always` |
 | — | `cache.exclude` | `[]` | Source-path glob patterns that bypass kache and compile normally without lookup, store, or upload |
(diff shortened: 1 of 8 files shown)


jleni:
CI caught a real regression on Linux — the exact platform I flagged as untested. Worth recording what it was, because it is not what I expected.

`e2e-rust-fallback` and `e2e-rust-sccache` both failed:

```
max_entries_after   expected <= 0   actual 1
verify              stdout missing substring: `rust-fallback: via fallback`
```

Their own header comments explain why:

> The project is a dependency-free `bin`. **kache never caches a user-facing executable, so building it ALWAYS takes the passthrough path** — which, with `KACHE_FALLBACK` set, hands the compile to the `e2e-fallback` wrapper.

So both fixtures use "executables are never cached" as their *mechanism* for reaching the passthrough path, where the fallback-wrapper feature (#109) lives. With executables cached on Linux, kache compiles the bin itself, the fallback never runs, and the marker it injects never reaches the binary.

The fixtures are testing fallback wiring, not the executable default, so they should not inherit a platform-dependent default for it. Both now pin `KACHE_CACHE_EXECUTABLES = "0"` explicitly, and their stale "kache never caches a user-facing executable" claim is corrected.

**The real interaction worth reviewing, beyond the test fix:** with executables cached, fewer compiles reach the passthrough path, so a configured `KACHE_FALLBACK` delegate sees less traffic on Linux. For the documented use (delegating to another *cache*, e.g. sccache) that is fine or better — kache handles it directly. It would matter for a fallback wrapper that *modifies* the compile, which is what these fixtures do with `--cfg` injection, though that is a test device rather than a documented use of #109.

Verifying on real Linux via docker now rather than relying on CI round-trips.

PR #607, opened 2026-07-25
fix(cc): classify --include=, -fwrapv and --param so aws-lc-sys TUs cache (#580)

### LINKED ISSUE #580: cc: `--include=<file>` unsupported → aws-lc-sys TUs pass through, re-poisoning the rustls subtree cross-clone (post-#471)
> [!NOTE]
> **Disclaimer:** this problem was investigated, analyzed, and written up by Claude Code (Anthropic's AI coding agent) on my machine, at my request. I don't have the expertise to verify the analysis in depth myself, so please treat the reasoning with appropriate skepticism — the raw command outputs are verbatim from my system, and I'm happy to run any follow-up commands to verify or provide more data.

## Summary

`aws-lc-sys 0.43.0` compiles some of its C/asm TUs with `--include=<path>/generated-include/openssl/boringssl_prefix_symbols*.h` (its symbol-prefixing mechanism). kache v0.11.0's cc classifier doesn't recognize `--include=` (nor `-fwrapv` / `--param` in the same invocations), so these TUs **pass through uncached**. The resulting objects differ per checkout, the `.a` content diverges, and the `extern:` content-hash cascade re-keys everything above `aws-lc-sys` — in my workspaces that is the whole rustls/TLS subtree (`rustls` → `aws-lc-rs` → `rig_core`, `aws_sdk_*`, `async_nats`, `sqlx_postgres`, `bollard`, …), which is exactly the set of expensive crates.

This is the same failure class as #471 (native `-sys` `.a` content divergence → weighted-hit-rate collapse), which was closed after #474/#476. Notably #471's SurrealDB bench found `aws_lc_sys` *byte-stable and hitting* — I believe that's why this case wasn't caught: the divergence only appears when the prefix-symbols `--include=` TUs are in play, because those never reach kache's cc cache (and its path normalization) at all.

## Data

Host: Arch Linux x86_64, kache v0.11.0 (AUR), rustc 1.97.1 (8bab26f4f 2026-07-14), local-only mode, btrfs. Several checkouts/worktrees of the same Rust workspaces built concurrently.

`kache stats` / `kache report --since 7d`:

```
Hit rate:   36.2% (local: 660, prefetch: 0, remote: 0, dup: 0, miss: 1162)
Weighted:   6.6% by compile cost
Miss share: 98.1% of wrapper time (~21min)
```

```
passthrough via direct: unsupported|cc unsupported flag(s): --include=/home/…/.cargo/registry/src/index.crates.io-…/aws-lc-sys-0.43.0/generated-include/openssl/boringssl_prefix_symbols_asm.h — not yet (56 total, 0 failed)
passthrough via direct: unsupported|cc unsupported flag(s): -fwrapv --param -fwrapv --include=/home/…/.cargo/registry/src/index.crates.io-…/aws-lc-sys-0.43.0/generated-include/openssl/boringssl_prefix_symbols.h — not yet (6 total, 0 failed)
```

Meanwhile aws-lc TUs *without* that flag do cache and hit (`bcm.c` appears twice in "Expensive cache hits", ~2s avoided each) — so this is specifically the unclassified-flag passthrough, not aws-lc C caching in general.

`kache why-miss aws_lc_sys` — 12 stored entries, single feature set `[prebuilt-nasm]`, same profile/target, all distinct keys, essentially all 0 hits:

```
Diagnosis: key mismatch -- 11 other entries exist but none matched the current build inputs
Differences detected:
  - same config as 37d5a93469ec -- likely source code, dependency, or rustc version change
```

The same "same config, N distinct keys, 0 hits" signature repeats up the cascade (`aws_lc_rs`: 23 entries, `rig_core`: 21, `aws_sdk_sesv2`: 14), and the report's top compiled misses are all in this subtree:

```
Top compiled cache-key misses:
  rig_core — ~11s (139.7 MB)
  aws_sdk_sesv2 — ~10s (180.3 MB)
  ...
  async_nats — ~3s (50.1 MB)
  sqlx_postgres — ~3s (40.1 MB)
```

## Control experiments (isolating the cause)

- A trivial crate copied to a second path gets 100% local hits → cross-clone reuse and rustc path normalization work.
- A crate with a `cc` build script (plain flags) copied to a second path: the C TU gets a **cc local cache hit** and the lib crate hits → kache's cc cache is cross-clone-stable when it can classify the flags.
- Only the `--include=`-flagged aws-lc-sys TUs bypass all of that, so their objects (and the archive) are rebuilt fresh per checkout and the content hash diverges downstream.

## Asks

1. **Classify `--include=<file>`** (and the co-occurring `-fwrapv`, `--param <x>`) in the cc wrapper — `--include` is the `=`-form of GCC/clang `-include`, a preprocessor input whose file content belongs in the key, analogous to flags added in #245 / #428 / #438 / #411 / #424. The paths here are under `$CARGO_HOME/registry/src` and therefore stable across checkouts, so once cached these TUs should converge cross-clone like `bcm.c` already does.
2. (Optional, may deserve its own issue) `why-miss` currently reports the generic "likely source code, dependency, or rustc version change" for these cascades. It would save a lot of digging if it could walk the `extern:` chain and point at the root divergent input — e.g. "dep `aws_lc_sys` artifact content differs; its build contains passthrough (uncached) cc TUs", in the spirit of #471's key-diff work.

Happy to provide `KACHE_LOG=trace` key component dumps or run a patched build if useful.

jleni:
Fixes #580.

aws-lc-sys 0.43 drives BoringSSL symbol prefixing with `--include=<generated-include>/boringssl_prefix_symbols*.h` and compiles jitterentropy with `-fwrapv --param ssp-buffer-size=4`. None of those flags classified, so around 62 TUs passed through uncached. Their objects differed per checkout, the `.a` diverged, and the `extern:` content hash re-keyed everything above `aws_lc_sys`, which in the reporter's workspaces is the whole rustls/TLS subtree.

Upstream source of the flags, from `aws-lc-sys-0.43.0/builder/cc_builder.rs`:

```rust
let force_include_option = if is_cl_like { "/FI" } else { "--include=" };
```

```rust
fn jitter_entropy_dialect_flags(is_cl_like: bool) -> &'static [&'static str] {
    if is_cl_like { &["-Od", "-W4", "-DYNAMICBASE"] }
    else { &["-fwrapv", "--param", "ssp-buffer-size=4", /* … */ "-O0", "-fwrapv", "-Wconversion"] }
}
```

## Classification

Verified against real GNU gcc 15.2.0 and Apple clang `-###`:

* **`--include=<file>` / `--include <file>`** are the long spellings of `-include` and resolve to cc1 `-include <file>`. `PreprocessorCaptured`, same as the short form: the header's bytes are already in the `-E` expansion the key hashes. Matched as `Prefix("--include=")` plus `Exact("--include")` so the unmodeled `--include-*` neighbours (`--include-directory=`, `--include-with-prefix=`, `--include-barrier`, …) keep refusing rather than getting swept in.
* **`-fwrapv`** joins the shared codegen-knob stem list. One stem covers both polarities, which is the structural guard from #422/#426.
* **`--param <name>=<value>` / `--param=<…>`** are `CapturedByProbe`. gcc joins the pair into a single `"--param=<name>=<value>"` cc1 token, so the resolved-token hash separates one value from another. clang drops the option before cc1 and also ignores it, so collapsing those keys is correct rather than lossy.

The separated `--param` value is a bare token the flag classifier treats as inert, which makes two existing mechanisms load-bearing for soundness here:

1. the `CapturedByProbe` row forces the `-###` probe, and key construction bails when it cannot resolve;
2. `config_args()` keeps the value in the probe-memoization key, so two compiles differing only in that value cannot share one resolved-invocation record.

Point 2 was not covered by any test. It is now, with the concrete `ssp-buffer-size=4` vs `=32` canary (the value that decides whether an 8-byte buffer gets a stack canary under gcc). The guard was checked red by adding `--param` to `drops_value`.

Also extends the `*.pch`/`*.gch` forced-include refusal to the long spellings, so modeling them cannot carry a PCH past a refusal the short form catches.

## Before / after

Same invocation, real gcc, two source dirs sharing one forced-include header:

```
# origin/main
cc unsupported flag(s): --include=…/pfx.h -fwrapv --param — not yet — passthrough

# this branch
clone A, cold  -> miss, compiler runs
clone A, warm  -> cc local cache hit (93176c8c056837c6)
clone B        -> cc local cache hit (93176c8c056837c6)   # different source dir, same key
```

`--param 4 -> 8` and `-fwrapv -> -fno-wrapv` each get their own key, so nothing false-hits.

## Tests

* classification and PCH coverage for the real aws-lc-sys flag shapes, plus a negative test pinning the `--include-*` prefix boundary
* `config_args_keeps_separated_param_value_issue_580`, pinning the probe-memo key premise above
* a frozen gcc 15.2.0 `-###` fixture asserting `-fwrapv` and the `--param` value survive into the gnu semantic tokens, and that a changed value changes the token list
* a cross-clone integration test where two source dirs sharing one forced-include header converge on one entry (skipped where `cc` is not on PATH)

Two things deliberately not asserted in the integration test, because both are compiler-dependent and would make it flaky:

* that a changed `--param` value misses. gcc misses; clang correctly hits, since it ignores the option and produces identical objects. The gcc side is pinned on the frozen fixture instead.
* that `-fno-wrapv` produces a *miss*. It gets its own key, but on a trivial TU the object comes out byte-identical, so it records as a `dup`. The test asserts local_hits did not grow, which is the property that actually matters.

## Not addressed

Ask 2 in the issue (having `why-miss` walk the `extern:` chain and name the root divergent input instead of "likely source code, dependency, or rustc version change") is a separate feature. The reporter suggested it may deserve its own issue, and I agree.

## Verification

Full workspace suite green, `cargo fmt --check` and `clippy --all-targets --all-features -D warnings` clean.

Cross-checked with codex (2/2 legs). Both legs independently flagged the same risk: if `config_args()` dropped the inert `--param` value, two values would share a probe-memo record and collide on one key. One leg proposed keeping the separated form fail-closed because of it. Checked against the implementation and against a live gcc run, that collision does not occur, so the row stands, but the gap in test coverage was real and is what `config_args_keeps_separated_param_value_issue_580` closes.


jleni:
Pushed `2086a93`: the Windows failure turned out to be a real pre-existing bug, not a test problem.

MinGW/MSYS gcc spawns `cc1.exe`, and the gnu `-###` extractor matched a bare `cc1` basename (`src/probe/resolve.rs`). The compile line was never found, so the probe reported unresolvable and **every** `CapturedByProbe` flag refused to cache on Windows. It was invisible because refusing is the safe direction: nothing failed, those invocations just never cached.

`-fwrapv` and `--param` are the first probe-keyed flags common enough to expose it, which is why this branch surfaced it. The fix strips a trailing `.exe` before matching, case-insensitively, still anchored on the whole basename so `notcc1.exe` and `cc1obj.exe` stay unmatched.

Issue #617, opened 2026-07-25
planner: the plan path dropped the compile-cost weighting the manifest path has


jleni:
## What

The two prefetch paths apply contradictory prioritization.

`monolithic_manifest_prefetch` does a cost and benefit filter, then sorts most expensive first
(src/daemon.rs:3971):

```rust
let mut worth_prefetching: Vec<_> = manifest.entries.iter()
    .filter(|e| e.compile_time_ms >= min_compile_ms)   // default 1000ms
    .collect();
// Most expensive crates first, maximizes value of limited S3 concurrency slots
worth_prefetching.sort_by_key(|entry| std::cmp::Reverse(entry.compile_time_ms));
```

The plan path applies no cost weighting at all. `order_candidates_by_crate_order` sorts purely by
the crate's position in `intent.crate_names`, which the wrapper fills from a guppy traversal.

## Impact

The newer path dropped a signal the older path already had. Under a bounded prefetch queue, order
decides what arrives before demand, so spending early slots on crates that take 40ms to compile
while a crate that takes 30s waits behind them is a direct loss.

Guppy order is also only an approximation of demand order. Cargo reorders work for parallelism,
build scripts, proc macros, feature resolution, and job availability.

Note that `compile_time_ms` is already indexed in the store as of #598, so the local fallback
planner has access to this signal today.

## Fix sketch

Rank candidates by expected value rather than graph position: roughly probability of use times
avoided compile time, divided by download cost. The exact `< 1000ms` cutoff need not be copied,
but cheap compiles and large low-value artifacts should not displace expensive likely-needed
ones.

This needs candidate size and compile cost on the wire, so it lands naturally with the budget
work.

Related: #583, #598.

Found during a cross-model design review of the prefetch subsystem; verified against main.

Issue #616, opened 2026-07-25
daemon: prefetch plans have no key, byte, or time budget


jleni:
## What

A prefetch plan has no cap on key count, no cap on bytes, and no deadline. `handle_prefetch`
filters candidates for validity and local presence, then downloads all of them.

The candidate sources feeding that plan are not equally trustworthy, and the weakest one is
unbounded. `key_cache_keys_for_crate` resolves a crate name against the in-memory index built
from the periodic full-bucket LIST, and returns every cache key in the bucket for that name:

```rust
async fn key_cache_keys_for_crate(&self, crate_name: &str) -> Result<Vec<String>> {
    let keys = self.daemon.key_cache_keys_for_crate(crate_name).await;
    ...
}
```

There is no filtering by target triple, toolchain, profile, features, or age. A crate name is
not a build identity. If a bucket holds `m` historical variants of a crate, this path offers all
`m` and the plan downloads all `m`, of which at most one is the entry this build wants.

`PrefetchCandidate` carries only `cache_key` and `crate_name`, so the daemon cannot weigh a
candidate by size or confidence even if it wanted to.

## Impact

Plan size is bounded only by how many variants happen to exist in the bucket. A long-lived
shared bucket makes the weakest candidate source the largest contributor to the plan, and there
is no ceiling to stop it. Wasted bandwidth is the visible cost; the cost that matters more is
that the wasted downloads occupy the prefetch queue ahead of candidates the build actually
needs.

Adaptive cancellation does not contain this. `should_cancel_prefetch` counts
`downloaded_not_demanded` as potential hits, so a plan that downloads a lot of unused entries
raises its own upper-bound hit rate.

## Fix sketch

- Add per-plan budgets: maximum candidate keys, maximum compressed bytes, and a wall-clock
  deadline after which remaining candidates are dropped.
- Extend `PrefetchCandidate` with compressed size and a source/confidence tag, so the daemon can
  spend the budget on high-confidence candidates first.
- Restrict undimensioned key-cache expansion. Either filter by build shape once the intent
  carries it, or place that source behind a strict sub-budget, or make it shadow-only until
  precision is measurable.
- Do not pick the first key returned for a crate as if the ordering meant anything.

Related: #583, #581.

Found during a cross-model design review of the prefetch subsystem; verified against main.

PR #622, opened 2026-07-25
fix(prefetch): merge planner sources; stop inferring whole-bucket warms (#614, #615)

### LINKED ISSUE #614: planner: one matching shard suppresses the history and key-cache fallback
## What

`build_prefetch_plan` returns as soon as the shard lookup produces any candidates
(crates/kache-core/src/lib.rs:75):

```rust
match source.shard_candidates(namespace, &intent.cargo_lock_deps).await {
    Ok(candidates) if !candidates.is_empty() => {
        return Ok(execute_plan(planner_name,
            order_candidates_by_crate_order(candidates, intent)));
    }
    Ok(_) | Err(_) => {}
}
```

Shard matching is exact per bucket. `compute_shards` assigns each crate to one of
`NUM_SHARDS = 64` buckets by `blake3(crate_name) % 64`, then hashes that bucket's sorted
`name@version` list. A shard hits only if the exact same set of pairs landed in that
bucket in the run that wrote it, so a single dependency bump invalidates one bucket while
the other 63 still match.

The early return fires on any non-empty result. History and remote key cache lookup are
then skipped for every crate whose shard missed, even though those sources could have
resolved them.

## Impact

A partial shard hit silently produces a partial plan and suppresses the recovery paths
that exist precisely for that case. The crates most likely to be missing from the plan
are the ones in buckets touched by a recent dependency change, which is also where a cold
build is most likely to need a fresh download.

The failure is invisible: the planner reports an `Execute` plan with candidates, and
nothing distinguishes a full shard hit from a 1-of-64 hit.

## Fix sketch

Merge candidate sources instead of returning early. Shard results seed `seen` and
`resolved_crates`, then history fills the crates still unresolved, then the key cache
fills what remains. The existing dedup and ordering logic already supports this; only the
early return needs to go.

Worth pairing with a plan field recording how many shards matched, so a degraded plan is
observable rather than inferred.

Tests to add: partial shard coverage, workspace crates absent from shards, and dedup that
preserves source priority.

Related: #583.

Found during a cross-model design review of the prefetch subsystem; verified against main.

### LINKED ISSUE #615: daemon: an empty PrefetchRequest means "download the whole bucket"
## What

In `handle_prefetch`, an empty key list does not mean "nothing to prefetch". It means
"list the entire bucket and download everything missing locally" (src/daemon.rs:2551):

```rust
// If empty keys were sent, fetch all remote keys missing locally.
if req.keys.is_empty()
    && let Ok(backend) = self.get_remote_backend().await
    && let Ok(s3_keys) = crate::remote_plan::RemotePlanner::new(&self.config)
        .plan(crate::remote_plan::RemoteWorkload::KeyDiscovery)
        .layout(backend.as_ref(), remote)
        .list_keys()
        .await
{
    for (key, crate_name) in s3_keys {
        ...
        if !entry_dir.exists() { keys_to_fetch.push((key, crate_name, entry_dir)); }
    }
}
```

There is no budget on the result. Every key in the bucket that is not already local gets
queued for download.

## Impact

Any caller that naturally encodes "no candidates" as an empty request starts an unbounded
whole-bucket download. The `BuildStarted` path currently guards against this by checking
`candidates.is_empty()` before building the request, but that is a property of one call
site rather than of the API.

On a shared bucket serving many repos, targets, and toolchains, the blast radius is the
entire remote cache, on a cold machine with an empty local store.

## Fix sketch

Make the sentinel explicit. "Warm everything" should be a distinct request variant or a
dedicated CLI operation with a required budget, and an empty candidate list should mean
exactly what it says.

Should land with the plan budget work so the explicit warm path is bounded too.

Related: #583.

Found during a cross-model design review of the prefetch subsystem; verified against main.

jleni:
Fixes #614. Fixes #615.

Two independent defects in how a prefetch plan is built and dispatched. Both are small and in
the planning path, so they travel together.

## #614: a partial shard hit suppressed every fallback

`build_prefetch_plan` returned as soon as the shard lookup produced any candidates. Shard
matching is exact per bucket: `blake3(crate_name) % NUM_SHARDS` picks the bucket and the shard
hash covers that bucket's sorted `name@version` list, so one dependency bump invalidates one
bucket while the other 63 still match.

A single matching shard was therefore enough to short-circuit history and key-cache recovery
for every crate in every bucket that missed. The result was indistinguishable from a full hit:
an `Execute` plan with candidates, silently missing the crates most likely to need downloading,
since the missed buckets are exactly the ones a recent dependency change touched.

Sources are now merged in descending order of confidence, each filling only the crates the ones
before it left unresolved: shards, then local history, then the remote key cache. Crates already
resolved by shards are no longer re-queried from the lower-confidence sources, which also drops
a store lookup per resolved crate.

Candidates stay grouped by source rather than globally re-sorted by crate order. Ranking is
#617's problem, and the pre-existing history/key-cache path already concatenated this way.

## #615: an empty request meant "download the whole bucket"

An empty `PrefetchRequest.keys` did not mean "nothing to prefetch". It meant "LIST the whole
bucket and download everything missing locally", with no budget on the result. Any caller that
encoded no-candidates the obvious way started a download proportional to the entire remote.

The `BuildStarted` path happened to guard against it by checking `candidates.is_empty()` before
building the request, but that was a property of one call site rather than of the API.

Whole-remote warming now has to be asked for: a `warm_all` field, `#[serde(default)]` so an old
client's `{"keys": []}` becomes a no-op rather than a bucket-sized download. Making it a required
field in Rust also made the compiler enumerate every construction site. It is still unbounded by
keys, bytes, or time, which is #616.

Nothing in-tree relied on the old sentinel: the only caller that could reach it, `send_prefetch`,
is `#[allow(dead_code)]`.

## Tests

Four added. Two were confirmed red before the fix and green after:

- `test_build_prefetch_plan_fills_crates_a_partial_shard_hit_missed` — with the old logic the
  plan is `["dep-shard-key"]` instead of all three crates
- `test_empty_prefetch_request_does_not_warm_the_bucket` — with the old condition restored it
  fails on "an empty prefetch request must not download anything". The fixture deliberately
  seeds both the manifest and pack objects so the key IS discoverable by listing; otherwise the
  test would pass vacuously

The other two are regression guards rather than reproductions:
`test_build_prefetch_plan_does_not_requery_shard_resolved_crates` (passes on old code too, since
the early return produced the same result) and `test_warm_all_prefetch_request_downloads_missing_keys`.

Verified on the commit in an isolated worktree, so unrelated work in my tree could not affect it:
1360 bin tests and 11 kache-core tests pass, `cargo clippy --all-targets --all-features` clean,
`cargo fmt --check` clean.

### DIFF
--- a/crates/kache-core/src/lib.rs
+++ b/crates/kache-core/src/lib.rs
@@ -80,19 +80,17 @@ where
     if let Some(namespace) = intent.namespace.as_deref()
         && !intent.cargo_lock_deps.is_empty()
     {
-        match source
+        // A failed shard lookup is not fatal: the sources below still run.
+        if let Ok(shard_candidates) = source
             .shard_candidates(namespace, &intent.cargo_lock_deps)
             .await
         {
-            Ok(shard_candidates) => {
-                for candidate in order_candidates_by_crate_order(shard_candidates, intent) {
-                    resolved_crates.insert(candidate.crate_name.clone());
-                    if seen.insert(candidate.cache_key.clone()) {
-                        candidates.push(candidate);
-                    }
+            for candidate in order_candidates_by_crate_order(shard_candidates, intent) {
+                resolved_crates.insert(candidate.crate_name.clone());
+                if seen.insert(candidate.cache_key.clone()) {
+                    candidates.push(candidate);
                 }
             }
-            Err(_) => {}
         }
     }
 
--- a/crates/kache-core/src/lib.rs
+++ b/crates/kache-core/src/lib.rs
@@ -65,6 +65,18 @@ pub async fn build_prefetch_plan<T>(
 where
     T: PlannerDataSource + Sync + ?Sized,
 {
+    let crate_order = crate_query_order(intent);
+    let mut seen = HashSet::new();
+    let mut resolved_crates = HashSet::new();
+    let mut candidates = Vec::new();
+
+    // Sources are merged in descending order of confidence, each one filling
+    // only the crates the ones before it left unresolved. Shard lookups used
+    // to RETURN as soon as they produced anything (kunobi-ninja/kache#614),
+    // but a shard hit is exact per bucket: one dependency bump invalidates one
+    // of `NUM_SHARDS` buckets while the rest still match, so a single matching
+    // shard was enough to short-circuit history and key-cache recovery for
+    // every crate in every bucket that missed.
     if let Some(namespace) = intent.namespace.as_deref()
         && !intent.cargo_lock_deps.is_empty()
     {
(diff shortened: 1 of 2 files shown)


jleni:
CI caught a real one: \`clippy::single_match\` on the shard lookup, failing `Check (Linux)`, `Test (macOS)` and `Test (Windows)` — all three were the same lint, not three problems.

My local `cargo clippy --all-targets --all-features` only warned because I omitted `-D warnings`; the gate is `just lint`, i.e. `cargo clippy --workspace --all-targets -- -D warnings`. Fixed in 35ac0d4 by taking clippy's own suggestion (`if let Ok(...)` instead of a one-armed match), which reads better anyway since the error arm was empty.

Re-verified with the exact CI command plus `cargo fmt --all --check`: both clean, 1360 bin tests and 11 kache-core tests pass. Note the earlier failures killed the build before `just ci` reached the coverage step, so the 88% gate is running for the first time on this push.

PR #636, opened 2026-07-27
Improve the monitor tui scroll state management

AzureZee:
## Summary
Fix the issue in BuildTab’s live build where pressing the down key scrolls up and pressing the up key scrolls down.

Use Viewport abstraction to improve scroll state management, adding a maximum limit of (item_count - visible_rows) to the scroll offset.

## Checklist

- [ ] User-visible changes (CLI, config, output) are reflected in the README or relevant docs

### DIFF
--- a/src/tui.rs
+++ b/src/tui.rs
@@ -78,12 +78,6 @@ impl Viewport {
     /// Update this viewport from the rows and height that will actually render,
     /// clamp stale state, and return the corresponding range in logical order.
     fn visible_range(&mut self, item_count: usize, visible_rows: usize) -> std::ops::Range<usize> {
-        if visible_rows == 0 {
-            self.offset = 0;
-            self.max_offset = 0;
-            return 0..0;
-        }
-
         let visible_rows = visible_rows.min(item_count);
         self.max_offset = item_count.saturating_sub(visible_rows);
         self.offset = self.offset.min(self.max_offset);
--- a/src/tui.rs
+++ b/src/tui.rs
@@ -2186,7 +2186,7 @@ mod tests {
         s.active_tab = Tab::Build;
         s.build_scroll.max_offset = 10;
         handle_key(&mut s, KeyCode::Up);
-        handle_key(&mut s, KeyCode::Up);
+        assert_eq!(s.build_scroll.offset, 1);
         handle_key(&mut s, KeyCode::Char('c'));
         assert!(s.events.is_empty());
         assert_eq!(s.build_scroll.offset, 0);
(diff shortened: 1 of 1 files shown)


emmanuelm41 (commented):
Welcome, and thanks — this is a really clean first PR. The scroll math is careful in exactly the places it usually goes wrong: `saturating_sub`/`saturating_add` throughout, so no `usize` underflow when the list is shorter than the viewport, and you clamp stale offsets in `visible_range` rather than trusting them. I traced the boundaries (empty, `item_count < visible_rows`, `== visible_rows`, top/bottom) and they all hold, with no wrap and nothing getting stuck. Nice touch routing every tab through one `Viewport` so the behavior stays consistent instead of fixing only BuildTab — and the bottom-anchor handling for Build is the right model for a tail view.

Two small things, neither blocking:

- **Tiny-terminal resize drops your scroll position.** In `visible_range`, the `visible_rows == 0` branch sets `offset = 0` (and `max_offset = 0`), so if the panel momentarily has no room to draw (a quick shrink), the offset is lost and Build snaps back to tail. Since `max_offset` at zero rows is really `item_count`, preserving and clamping — `self.max_offset = item_count; self.offset = self.offset.min(self.max_offset);` — keeps the position across a shrink/restore. Minor, but it's the one spot the clamp changes behavior rather than just bounding it.
- **One test could have more teeth.** `handle_key_clear_resets_build_offset` presses Up twice then asserts `offset == 0` after clear — but it never asserts the Ups moved the offset first, so if Build's Up/Down dispatch were reversed the test would still pass. A quick `assert_eq!(offset, 2)` before the clear (or a small "Up moves Build toward older rows" test) would lock the direction you're actually fixing.

Everything else checks out — no new panic paths, tailing still follows at offset 0, and the resets on filter/clear/sort are all justified by the underlying list changing. I read the code and traced the math; I didn't run it in a live terminal.

AzureZee:
@emmanuelm41 Thanks!

https://github.com/user-attachments/assets/d927d76d-3e5a-4864-ad77-7904c3baada2

emmanuelm41 (approved):
Approving. The tiny-terminal resize regression is fixed — I checked that `visible_range` now preserves the offset when no rows fit (and restores the position on resize back), so scroll no longer snaps to tail. Combined with the earlier pass — careful saturating scroll math, no underflow, consistent `Viewport` across all tabs, correct bottom-anchor tailing for Build — the change is solid, and all the required checks are green.

Two non-blocking notes for later if you feel like it: the `Mutation testing` job (not a required gate) still shows surviving mutants in the `draw_*` functions — those are genuinely awkward to unit-test, but a couple of assertions on the Build Up/Down direction and the `visible_range` boundaries would kill the scroll-relevant ones. Neither is needed for this to land.

Really nice first contribution — thanks for it, and welcome. 🎉

I read the code and reproduced the resize fix locally; I didn't drive it in a live terminal.

Issue #618, opened 2026-07-25
prefetch: P0.5 telemetry cannot compute byte precision or blocking time


jleni:
## What

P0.5's stated exit criterion is that telemetry can compute per-plan precision by bytes and
cold-checkout blocking time. `BuildSummaryEvent` cannot do either.

Byte precision is unavailable because `used_bytes` is a lower bound by construction, as the
field docs already acknowledge. A completed prefetch is imported into the local store and
consumed by the wrapper as a `LocalHit` without contacting the daemon, so daemon-side demand
never sees it. In the limit a perfect plan looks unused: 100 keys downloaded, 100 consumed
locally, `used_keys` zero. The ratio `used_bytes / downloaded_bytes` can therefore sit anywhere
between 0 and the true value, and nothing in the record says where.

Blocking time is unavailable because there is no field for it. `started_at_ms` and
`last_activity_ms` bracket an inferred session, not compiler wait time. The record has no demand
timestamp, no download completion timestamp, and no way to distinguish "arrived before it was
needed" from "arrived after the compiler had already waited for it".

Also missing: 404 and download failure counts, and which candidate source and rank produced a
given download.

## Impact

Timeliness, not coverage, is what prefetch is for. A candidate that was downloaded and later
compiled counts as a success under the current record even if the compiler blocked on it the
whole time.

The direct consequence is for the roadmap. P1-min makes these aggregates durable and P2a scores
a shadow planner against them. Both inherit the bias. A shadow planner scored against
daemon-visible demand systematically excludes its own best outcomes, which are the prefetches
that landed in time and were consumed locally.

## Fix sketch

Add wrapper-visible attribution so every demanded key is represented, including local hits,
without putting a synchronous call on the compiler hot path. Per key, record at minimum:

- first demand timestamp, and prefetch completion timestamp
- outcome: local hit, prefetched local hit, remote hit, or miss
- plan id and candidate rank where applicable
- milliseconds blocked waiting on an in-flight download or direct GET
- attempted GETs, 404s, errors, cancellations
- LIST result count alongside the existing request count and duration

Define a useful prefetch as one that completed before first demand or measurably reduced demand
wait, not merely one that was eventually compiled.

Collect from real cold CI jobs with a control group that has automatic prefetch disabled, and
measure total job time rather than cargo time, so an explicit warm step cannot look like an
improvement by moving the same wait outside the timer.

Suggest holding P1-min until this lands.

Related: #583, #584.

[...]

PR #640, opened 2026-07-28
feat(prefetch): rank candidates by cost and urgency, cap weak sources (#616, #617)

### LINKED ISSUE #616: daemon: prefetch plans have no key, byte, or time budget

### LINKED ISSUE #617: planner: the plan path dropped the compile-cost weighting the manifest path has

jleni:
Closes #616. Closes #617.

Stage 1 (#633) bounded a plan's **resource** use: keys, bytes, deadline. This bounds its
**composition**, and decides what order the survivors are fetched in.

## #617: ranking

The plan path applied no cost weighting at all, while the older monolithic-manifest path
sorted by `compile_time_ms` and filtered crates cheaper to rebuild than to fetch. The newer
path had simply dropped a signal the older one already had.

Candidates now carry what ranking needs, and every source tags what it produced:

- `PrefetchCandidate` gains `compile_time_ms`, `size_bytes`, `source`, `demand_index`. All
  `Option`/defaulted, so an old planner's candidates still parse and a new plan still reaches
  an old daemon.
- `ShardEntry` gains `compile_time_ms` and `artifact_size`. `save-manifest` already had both
  from `ManifestEntry` and was discarding them at write time. Old shards in S3 parse as `None`.
- `keys_for_crates` selects the two columns it was ignoring and returns a named
  `CrateHistoryEntry` instead of a bare `(String, String, PathBuf)`.

**Unknown is never zero.** Both store columns default to `0` for rows written before their
migrations. Read as a measurement, a `0` would rank an un-backfilled entry as free to fetch
and worthless to have, so `0` maps to `None`.

Dispatch order is **lexicographic, not a weighted score**: a weighted sum needs coefficients,
and nothing can calibrate them until #618 makes "arrived before it was demanded" measurable.

1. **Urgency bucket.** Prefetch races the build, so a costly artifact needed at minute eight
   loses to a cheap one needed at second five.
2. **Source confidence.**
3. **Value descending**, unknown cost last (after known cost, but still ahead of anything in a
   later bucket).

Demand position is *bucketed* (width 16, about one download wave) rather than used exactly. It
comes from a guppy traversal, which only approximates when cargo actually asks, so treating
position 40 and 45 as different is false precision while 40 versus 400 is real.

## #616: composition caps

`PlanLimits` bounds what each source may contribute. That is a different job from stage 1's
byte/key/time budgets: those bound resource use and are the trust boundary, these stop the
weakest source crowding out better candidates before the budget is even reached.

The key cache maps a crate **name** to every key in the bucket, with no target, toolchain,
profile, or feature dimension, so of `n` variants at most one can be right. Defaults: 2 per
crate, 64 per plan, plus 2 history entries per crate. Shards get no per-crate cap, because one
crate legitimately has several compile units.

Capping is not a fix for that source; dimensioning the remote layout is, and that needs its own
index version and migration. The design note says so rather than implying this closes it.

Drops are counted per class and logged, so a trimmed plan stays distinguishable from one that
had nothing more to offer.

## Why this is one PR, not three

The design note staged this as 2, 3, 4. I revised that in the note. All three rewrite
`build_prefetch_plan`, the `PlannerDataSource` surface, and `fallback_planner.rs`, so splitting
meant changing the same functions three times over two rebases. And the metadata stage alone
would have landed wire fields no code reads, which a reviewer cannot judge without seeing the
ranking that consumes them. The separation that carried its weight was enforcement from
selection; going finer than that did not.

## The mutation gate found real gaps

`kache-core` is under complete mutation coverage as of #634. My first test pass looked fine at
24 green, but **9 mutants survived**, including the per-crate history cap, which had **no test
at all**.

The instructive one: `cap_to`'s `-` mutated to `/` survived because the test used 4 items
capped to 2, where subtraction and division both give 2. The test asserted the right value for
the wrong reason.

After targeted tests: `cargo-mutants` reports **52 caught, 0 missed** for `kache-core` (was
43 caught, 9 missed). Run locally with cargo-mutants 27.0.0; CI pins 27.1.0.

## Verification

- `cargo clippy --workspace --all-targets -- -D warnings` (the real `just lint` gate): clean
- `cargo fmt --all --check`: clean
- 1409 bin tests + the rest of the workspace: pass

One local e2e failure, `clone_worktrees_pull_makes_both_commits_checkoutable`, is my own git
commit-signing config breaking that test's temp-repo commits. It passes with
`GIT_CONFIG_GLOBAL=/dev/null`, touches no prefetch code, and CI runners have no signing key.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2916,7 +2916,7 @@ dependencies = [
 
 [[package]]
 name = "kache"
-version = "0.12.0"
+version = "0.13.0"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -2964,7 +2964,7 @@ dependencies = [
 
 [[package]]
 name = "kache-core"
-version = "0.12.0"
+version = "0.13.0"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -2976,7 +2976,7 @@ dependencies = [
 
 [[package]]
 name = "kache-e2e"
-version = "0.12.0"
+version = "0.13.0"
 dependencies = [
  "anyhow",
  "chrono",
@@ -2990,7 +2990,7 @@ dependencies = [
 
 [[package]]
 name = "kache-service"
-version = "0.12.0"
+version = "0.13.0"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kache"
-version = "0.12.0"
+version = "0.13.0"
 edition = "2024"
 description = "Zero-copy, content-addressed build cache for Rust, C/C++ and more, with S3 and shared-filesystem remotes."
 license = "Apache-2.0"
@@ -36,7 +36,7 @@ reqsign-aws-v4 = { version = "=3.0.2", default-features = false }
 reqsign-core = { version = "=3.1.0", default-features = false }
 # Direct dep so OpenDAL/reqwest can use ring without reintroducing aws-lc.
 rustls = { version = "0.23", default-features = false, features = ["ring"] }
-kache-core = { version = "0.12.0", path = "crates/kache-core", default-features = false, features = ["planning"] }
+kache-core = { version = "0.13.0", path = "crates/kache-core", default-features = false, features = ["planning"] }
 async-trait = "0.1"
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "macros", "process", "fs", "io-util", "net", "time", "signal", "sync"] }
 serde = { version = "1", features = ["derive"] }
--- a/crates/kache-core/Cargo.toml
+++ b/crates/kache-core/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "kache-core"
-version = "0.12.0"
+version = "0.13.0"
 edition = "2024"
 description = "Core planner data structures and algorithms for kache."
 license = "Apache-2.0"
--- a/crates/kache-core/src/lib.rs
+++ b/crates/kache-core/src/lib.rs
@@ -17,10 +17,79 @@ pub struct BuildIntent {
     pub cargo_lock_deps: Vec<(String, String)>,
 }
 
+/// Which source produced a candidate, i.e. how much to trust it
+/// (kunobi-ninja/kache#617).
+///
+/// `Unknown` is the `#[serde(other)]` arm so a newer planner naming a source
+/// this build has never heard of degrades to "untrusted" instead of failing
+/// the whole plan.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq, Eq, Default)]
+#[serde(rename_all = "snake_case")]
+pub enum CandidateSource {
+    /// Exact lockfile-shard match: this build's dependency set produced it.
+    Shard,
+    /// This machine built this crate before.
+    History,
+    /// A crate NAME matched something in the remote listing. A crate name is
+    /// not a build identity, so this is a guess.
+    KeyCache,
+    #[default]
+    #[serde(other)]
+    Unknown,
+}
+
+impl CandidateSource {
+    /// Confidence rank, lower is better. Only the ORDER matters; these are not
+    /// probabilities and must not be presented as any.
+    pub fn confidence_rank(self) -> u8 {
+        match self {
+            CandidateSource::Shard => 0,
+            CandidateSource::History => 1,
+            CandidateSource::KeyCache => 2,
+            CandidateSource::Unknown => 3,
+        }
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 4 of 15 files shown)


jleni:
Pushed f9f6686: workspace bump to v0.13.0.

`Check (Linux)` failed on the `cargo package -p kache --locked` step. Nothing else in that job failed: `just ci` (fmt-check, lint, coverage run) passed, and the coverage *threshold* step plus the whole `Mutation testing` job were skipped as downstream of it.

Packaging resolves `kache-core` from the registry rather than the workspace path, so the bin gets compiled against published `kache-core` 0.12.0, which has none of `CandidateSource`, `PlanLimits`, `build_prefetch_plan_with_limits`, or the new `PrefetchCandidate` fields. Reproduced locally: five E0433/E0425/E0560 errors.

That is not restructurable. The ranking metadata originates in the bin's data sources (store index rows, shard entries) and has to reach the planner through the shared `PrefetchCandidate`, so the bin genuinely needs the newer kache-core.

0.13.0 rather than 0.12.1 because kache-core gains public API. With 0.13.0 unpublished the packaging step takes its documented "not on crates.io yet" branch and skips, which is the state the tree normally sits in between a bump and its release. All four manifests plus the dep pin move together so `check-version-consistency.sh` stays green.

Flagging one thing for review: this folds a release-number bump into a feature PR, whereas the convention has been standalone `chore: bump` PRs (#570, #612). Happy to split it back out if you would rather keep that boundary.

PR #646, opened 2026-07-30
fix: close build coalescing race

alecthomas:
## Summary

- Re-check the local store after acquiring the per-key build lock.
- Restore a peer-committed artifact instead of compiling again when another process commits between the initial miss and lock acquisition.
- Update the architecture docs to show the post-lock re-check.

## Why

kache already coalesces same-host in-flight Rust builds with a per-key lock, but there was a lookup-to-lock race: if project A committed an artifact after project B saw a miss but before B acquired the lock, B would acquire the now-free lock and compile the same artifact again. The new claim flow closes that window while preserving the existing wait-for-winner behavior under contention.

## Verification

- cargo test --workspace
- cargo fmt --all -- --check
- git diff --check

A normal push was blocked by three unrelated pre-existing clippy warnings in remote_backend.rs, tui.rs, and wrapper.rs; the changed code was previously checked with those existing lints allowed.

### DIFF
--- a/docs/how-it-works/architecture.mdx
+++ b/docs/how-it-works/architecture.mdx
@@ -26,6 +26,8 @@ check remote (via daemon) ── hit → restore via reflink → done
     ↓ miss
 acquire per-key build lock
     ↓
+re-check local store ── peer committed → restore via reflink → done
+    ↓ still absent
 run rustc
     ↓
 store output files in content-addressed blobs
@@ -35,7 +37,7 @@ send upload job to daemon (async)
 release lock
 ```
 
-The lock prevents duplicate compilation when cargo spawns multiple parallel rustc processes for the same crate (which can happen in certain workspace configurations). A process that loses the lock waits for the winner's result and restores from the cache instead of compiling.
+The lock prevents duplicate compilation when cargo spawns multiple parallel rustc processes for the same crate (which can happen in certain workspace configurations). A process that loses the lock waits for the winner's result and restores from the cache instead of compiling. A process that acquires the lock re-checks the store before compiling, covering a peer that committed between the initial lookup and lock acquisition.
 
 Build outcomes are separated into three cacheable cases:
 
--- a/src/store.rs
+++ b/src/store.rs
@@ -1335,6 +1335,11 @@ impl Store {
             return Ok(BuildClaim::Contended);
         };
         match self.get(cache_key)? {
+            Some(meta) if meta.files.is_empty() => {
+                tracing::warn!("cache entry {cache_key} has no files, evicting before build");
+                self.remove_entry(cache_key)?;
+                Ok(BuildClaim::Acquired(lock))
+            }
             Some(meta) => Ok(BuildClaim::Committed(Box::new(meta))),
             None => Ok(BuildClaim::Acquired(lock)),
         }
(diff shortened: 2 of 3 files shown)


emmanuelm41 (commented):
Welcome, and nice PR — this is a real race and the fix is the right shape. Re-checking the store *under* the build lock is exactly how to close the window: once you hold the lock, no cooperating wrapper can commit before you compile, and if the previous owner committed just before releasing, your `get()` sees it and you restore instead of recompiling. The lock lifecycle is clean too — `Committed` holds the temporary guard and drops it exactly once, `Acquired` hands the guard to the wrapper, `Contended` never owns one — and the wrapper refactor that unifies the "committed → restore" path (shared by both the claim re-check and the post-contention wait) is a genuine simplification. `Box<EntryMeta>` is the right call for the enum size, and the diagnostics replay / `LocalHit` logging / `clean_incremental_dir` are all preserved.

One blocker before it lands: **the `Committed` path doesn't guard against poisoned empty entries.** An entry with `files.is_empty()` is a known bad state (from earlier bugs) that every other hit path defends against by evicting and recompiling — `wrapper.rs:1184` ("Safety: skip entries with no cached files"), the cc path at `:561`, the daemon-hit path at `:1881`, and `store.rs:366`. But `claim_build` returns any committed entry as `Committed(meta)` without that check, and `restore_from_cache` on an empty `files` list iterates nothing, succeeds vacuously, and the wrapper returns `Ok(0)` — a "successful" build that produced **no output**, which then breaks the actual compile (cargo expects the `.rlib`/`.o` to exist). Before this change, that window would have re-acquired the released lock and compiled, so it's a new way to serve the poison.

The fix mirrors the existing guard, in `claim_build` while you still hold the lock:

```rust
match self.get(cache_key)? {
    Some(meta) if meta.files.is_empty() => {
        tracing::warn!("cache entry {} has no files, evicting before build", cache_key);
        self.remove_entry(cache_key)?;
        Ok(BuildClaim::Acquired(lock))   // compile instead of serving the poison
    }
    Some(meta) => Ok(BuildClaim::Committed(Box::new(meta))),
    None => Ok(BuildClaim::Acquired(lock)),
}
```

Two smaller, non-blocking notes:
- The regression test proves the re-check and the release, but it commits directly rather than modeling the real sequence (waiter misses → a peer holds the lock, commits, and releases → waiter claims). Driving it through a second `Store` handle's `claim_build`/`drop` makes it exercise the actual production path deterministically, no threads needed. And an `empty_committed_entry` test would lock in the fix above.
- Now that `Err` also covers a re-check `get()` failure after the lock was acquired, the `"acquiring build lock ... failed"` message is slightly off for that case — worth a word tweak, though the passthrough behavior is correct and safe.

Worth noting the scope this closes: it covers cooperating local wrapper writers. Daemon remote imports don't take the build lock, and stale-lock reclaim can still reap a lock older than an hour — both pre-existing, not this PR's job, just so the boundary is clear.

I read the code and traced the lock/restore paths; I didn't run a multi-process repro. Thanks again — this is a strong first contribution, just needs the empty-entry guard.

alecthomas:
Addressed the blocking feedback in `68c01cb`:

- `claim_build()` now evicts committed entries with an empty `files` list while holding the build lock and returns `Acquired`, so the wrapper compiles instead of serving a poisoned empty entry.
- The race regression now models the real peer/waiter sequence with two `Store` handles: waiter contends, peer commits and releases, waiter reclaims and restores.
- Added an empty-entry regression test.
- Tweaked the wrapper warning from lock acquisition failure to build claim failure.

Verification:
- `cargo test --workspace claim_build`
- `cargo test --workspace`
- `git diff --check`
- `cargo clippy --workspace --all-targets -- -D warnings -A clippy::nonminimal-bool -A clippy::collapsible-match -A clippy::collapsible-if`

Plain clippy still fails on the three unrelated existing warnings noted earlier.

alecthomas:
Thanks for reviewing!

emmanuelm41 (approved):
Approving. The coalescing-race fix is correct — re-checking the store under the build lock closes the TOCTOU window, and the follow-up added exactly the missing piece: `claim_build` now evicts a poisoned empty (`files.is_empty()`) entry and compiles instead of serving it, matching the guard every other hit path already had, with a regression test to lock it in. Lock lifecycle is sound, the wrapper restore path is unified cleanly, and all required checks are green (the only red is the non-required Mutation testing job). Nicely done.

Issue #429, opened 2026-06-23
Cached compiler check breaks Firefox configure step


saschanaz:
## What happened

```
> KACHE_LOG=kache=debug ./mach build
 0:00.71 W Warnings in third-party code are being suppressed from the terminal output. Use --show-all-warnings or --verbose to see them.
 0:00.71 W All warnings will still be dumped to D:/firefox/obj-aarch64-pc-windows-msvc\.mozbuild\logs\build\warnings_*.json at the end of the build.
Streaming resource usage profile to: D:/firefox/obj-aarch64-pc-windows-msvc\.mozbuild\logs\build\profile_log_20260623_181340.json
 0:01.40 Clobber not needed.
 0:01.40 D:/firefox/obj-aarch64-pc-windows-msvc/config.status is out of date with respect to D:/firefox/build/moz.configure/toolchain.configure
 0:01.77 Using Python 3.14.6 from C:\Users\sasch\.mozbuild\srcdirs\firefox-b46365449af4\_virtualenvs\build\Scripts\python.exe
 0:01.77 Adding configure options from D:\firefox\mozconfig
 0:01.77   --disable-optimize
 0:01.77   --target=aarch64-pc-windows-msvc
 0:01.77   --with-compiler-wrapper=kache
 0:01.77 checking for vcs source checkout... jj
 0:01.79 checking for a shell... C:/mozilla-build/msys2/usr/bin/sh.exe
 0:01.79 checking for host system type... aarch64-pc-windows-msvc
 0:01.91 checking for target system type... aarch64-pc-windows-msvc
 0:02.06 checking whether cross compiling... no
 0:02.10 checking if configuration file confvars.sh exists... D:/firefox/browser/confvars.sh
 0:02.10 checking if configuration file configure.sh exists... D:/firefox/browser/branding/unofficial/configure.sh
 0:02.11 checking for Python 3... C:/Users/sasch/.mozbuild/srcdirs/firefox-b46365449af4/_virtualenvs/build/Scripts/python.exe (3.14.6)
 0:02.11 checking for git... C:/PROGRA~1/Git/cmd/git.exe
 0:02.11 checking for jj... C:/Users/sasch/.cargo/bin/jj.exe
 0:02.13 checking for Git version... 2.54.0.windows.1
 0:02.13 checking for MozillaBuild directory... C:\mozilla-build
 0:02.13 checking for MozillaBuild version... 4.2.1
 0:03.46 checking for ccache... not found
[...]

PR #648, opened 2026-07-31
fix(restore): keep the executable bit on custom-harness test binaries

glebpom:
A `[[test]] harness = false` target supplies its own `main`, so cargo
invokes rustc with neither `--test` nor `--crate-type`. Its extensionless
output therefore classifies as `Other("rustc:unknown")`, whose link
strategy is `Hardlink`, and the restore path derived the output's
permissions from that strategy alone — so no `0o755` was applied.

On a CoW filesystem the bit is lost twice over: the reflink ingest writes
the blob into a fresh temp file that `set_blob_readonly` leaves at `0o444`,
and the reflink restore then creates the output at the process umask and
never chmods it. The restored test binary comes back non-executable and
cargo fails the run with "Permission denied (os error 13)". On tmpfs both
paths fall back to `fs::copy`, which preserves the source mode and hides
the whole thing.

The executable bit recorded at insert time is the reliable signal, and the
insert side already trusts it over the filename (`hardlink_eligible`
refuses to hardlink anything carrying a mode bit). Restore now trusts it
the same way, which also keeps executables on the independent-inode path
so a post-build `strip` or codesign cannot reach back into the shared blob.

Covered by unit tests in `wrapper::tests`, which reproduce without
depending on filesystem capabilities, and by a new end-to-end
`custom_harness_test` that builds the `test-projects/custom-harness`
fixture through the real wrapper, drops the target dir, and lets cargo
execute the restored binary. The e2e test runs on the repository's
filesystem rather than TMPDIR, since tmpfs masks the failure.


emmanuelm41 (commented):
Welcome, and this is an excellent first PR — the diagnosis is spot-on and the fix is right. A `harness = false` target compiling to `Other("rustc:unknown")` → `Hardlink` → no `0o755` over a `0o444` blob is exactly the kind of misclassification that's easy to miss, and trusting the recorded `executable` bit over the (absent) kind signal is the correct call. The rationale for `Copy` — independent inode so a post-build `strip`/codesign can't reach the shared blob — is the same reasoning behind #429/#508, and it's good that you spelled it out.

Worth stating for reassurance, since it's the natural worry: routing executables to `Copy` does **not** cost you zero-copy. `link_to_target` tries reflink first regardless of strategy, so on a CoW filesystem an executable still reflinks (block-shared, independent inode, then `0o755`); `Copy` only changes the *non-CoW* fallback from hardlink to byte-copy — which is precisely what's needed to restore `+x` and keep the blob safe. And the `0o755` is set on all three paths (reflink-success, copy fallback, and the transformed `write_restored`). Windows is unaffected (`is_executable` is always false there, and `.exe` already classifies as `Copy`). The direct unit test is a solid, mutation-sensitive guard — it forces the blob to `0o444` and fails without the patch on every Unix restore path.

One thing worth tightening (not blocking, since the unit test already guards the fix): **the integration test isn't a deterministic regression guard.** Two gaps:

- It never asserts the warm run was a cache **hit** — a warm miss that recompiled would satisfy the "binary is executable" check without exercising restore at all.
- It assumes a same-filesystem reflink drops `+x`, but on macOS `clonefile` copies the source's mode. If the store blob kept its execute bits (a copy-ingested blob is `0o555`), the reflink restore reproduces `+x` and the **old, unfixed** code passes — so the test can green even without your change, depending on platform and ingest path.

To make it a real guard on every platform: after the cold build, explicitly clear `+x` from the store blobs, and assert the warm run recorded a `local_hit` for the `harness` crate (via `kache report --format json`). Then a non-executable restore is the only way through, and it fails without the fix. Happy to point at the exact spots if useful.

I read the code and traced the link/permission paths; the production side is correct and I have no blocking concern there — just the E2E test hardening.

emmanuelm41 (approved):
The E2E hardening is exactly right — and you went further than I asked. The test is now a real regression guard:

- The warm run must be a `local_hit` with `compiler_runs == 0`, and the cold run is asserted to have compiled `harness` — so a warm *miss* that relinks the binary can no longer satisfy the mode check without ever entering the restore path.
- Every `store/` blob is stripped to `0o444` before the warm run, so no materialization path — reflink, hardlink, or macOS `clonefile` — can reproduce `+x` for free. `stripped > 0` guarding against a store-layout change is a nice touch.
- Scoping the strip to `store/` (leaving `index.db` and `events.jsonl` at cache-root writable) is what keeps the warm restore and `kache report` working — I confirmed the layout to be sure.

All four verdicts are clean:

- **Safety:** CLEAR — test-only delta, no workflow or production change.
- **Purpose:** ACCOMPLISHES — closes the deterministic-guard gap from the last round.
- **Tests:** COVERED — the guard now fails without the production fix on every Unix restore path.
- **Docs:** N/A — test-only.

The production fix was already right in the first pass. I approved the CI run; green there and this is ready to squash from my side. Nice work, and thanks for the thorough follow-through.

PR #649, opened 2026-08-03
feat(prefetch): make speculative remote indexing optional

jmagar:
## Summary

- add cache.prefetch_enabled / KACHE_PREFETCH_ENABLED so long-lived warm runners can disable speculative manifest, advisory, and fallback prefetch without disabling exact remote hits or background uploads
- add cache.remote_key_cache_refresh_secs / KACHE_REMOTE_KEY_CACHE_REFRESH_SECS; zero performs one startup population and disables periodic full-remote refreshes
- preserve these advanced settings through the config TUI, expose the policy in kache stats, and document runner-lifetime guidance

Exact-key RemoteCheck and upload paths are intentionally unchanged.

## Test plan

- [x] cargo check --all-targets --all-features
- [x] cargo clippy --workspace --all-targets --all-features -- -D warnings
- [x] new config, TUI, stats, disabled-planning, exact-remote-hit, and upload regression tests
- [x] full suite: 1,459 passed; two existing ZFS FIEMAP/reflink failures in untouched sharing.rs reproduce on clean main
- [x] isolated MinIO producer/consumer canary: one upload, one clean-cache exact remote hit, zero speculation or remote-key-cache/List activity after 65 seconds
- [x] repository-native just check: fmt and Clippy pass; tests report 1,459 passed and the same two untouched ZFS FIEMAP/reflink failures reproduced on clean main
- [x] refresh-zero live canary: one startup population, refresh-disabled marker, zero periodic refreshes after 65 seconds

## Checklist

- [ ] just check is fully green locally; blocked only by the two clean-main ZFS sharing-test failures described above
- [x] New behavior is covered by tests
- [x] Commit message follows conventional commits
- [x] User-visible config, CLI output, daemon behavior, and CI guidance are documented


jmagar:
GitHub Actions created run 30781186422 with action_required. This is a first-time fork contribution, so a maintainer needs to approve the workflow run before CI can execute.

jleni (commented):
Thanks for this. The shape is right: the two knobs default to exactly the current behavior (60s refresh reproduces the old 300s authoritative window), the env parsing follows the `clean_incremental` default-true precedent, `IGNORE_ENV_GATED_VARS` is updated, and the TUI round-trip is covered by a test. Three things before it goes in.

### 1. The env var will not reach an already-running daemon, and `kache stats` will claim otherwise

`docs/remote-cache/ci.mdx` tells long-lived self-hosted runners to set `KACHE_PREFETCH_ENABLED=0`. But the daemon captures its env once at startup, and the only auto-restart triggers are the config *file* fingerprint watcher in `server_main` and the client-binary-newer check. With `DEFAULT_DAEMON_IDLE_TIMEOUT_SECS = 600`, a runner with back-to-back jobs keeps the same daemon, so the env var does nothing until an idle gap happens to reap it. That is the exact deployment this PR is written for, and the failure is silent and intermittent.

On top of that, `render_stats` reads the *client's* config, so `kache stats` prints "Prefetch: disabled" while the daemon is still LISTing and planning.

Please lead with the file-based `prefetch_enabled = false` in the CI doc, since a file change triggers the fingerprint restart on its own, and note that changing the env var needs `kache daemon restart`. Sourcing the stats line from the daemon snapshot rather than client config would be better still, but the doc fix is the part that blocks.

### 2. `remote_key_cache_refresh_secs` now also controls how long a stale negative index is trusted

`KEY_CACHE_AUTHORITATIVE_MULTIPLIER = 5` makes the trust window `refresh * 5`. The configuration doc suggests raising the interval as the cheaper middle ground, but at `900` a key uploaded by a peer just after a LIST is answered as a miss with no HEAD probe for up to 15 minutes, and up to 75 minutes if refreshes are failing. On a shared remote that is a real hit-rate loss, and it sits badly with "exact remote lookup remains enabled", since the lookup can then be served entirely from a stale snapshot.

The old coupling was implicit and fixed at 60/300. Now it is user-tunable, so I would rather it were bounded: either cap it at the old value (`min(refresh * 5, 300s)`) or give the authoritative window its own constant. Either way the tradeoff should be written down next to the advice to raise the interval.

### 3. The warming barrier is signalled by accident, and nothing pins it

With a remote configured and prefetch off, `manifest_handle` falls into the existing `else` arm, which does call `signal_warming_complete()`, so the behavior is correct today. But that arm still reads "No remote configured, nothing to warm, unblock immediately", which is no longer true. If anyone later tightens the branch to match its comment, every `handle_remote_check` pays the full 750ms `REMOTE_CHECK_WARMING_GRACE`, because `wait_for_warming` only short-circuits once the flag is set. Please fix the comment and add a test asserting the barrier is released when a remote is configured and prefetch is disabled.

### Smaller things

- `handle_prefetch` checks `prefetch_enabled` before the remote check and `handle_build_started` checks it after, so the disabled-plus-no-remote combination answers differently in each. Cosmetic, but the new tests pin both orderings.
- `refresh_secs = 0` gives a startup LIST whose negative half is then never trusted, because of the `> 0` gate. Positive entries and fallback planning still benefit, so it is not wasted, but the docs describe only the refresh half of that.
- Renaming `PreservedEnvLists` to `PreservedAdvancedConfig` dropped the original comment's rationale, which was that the two `Option<Vec<String>>` lists are positionally swappable and a swap would silently corrupt a user's config. The new members are differently typed, so the replacement comment no longer covers that hazard. Worth keeping the original paragraph alongside the new one.
- `warm_all: true` is only constructed in tests today, so the early return cannot silently no-op a user-facing warm command. Worth remembering if such a command lands later.

Test coverage is good otherwise: config precedence, `ignore_env` pinning, TUI round-trip, both daemon disabled paths, and both stats renderings. Nothing covers the `server_main` task spawning, which is where item 3 lives.

The two ZFS FIEMAP/reflink failures you hit do reproduce on clean main, so those are not yours.

jmagar:
Mutation-test follow-up pushed in b312f05c692fe7ea88b36fcaf150db3b5280f2ab.

Local validation on the rebased branch:

- workspace check: passed
- strict all-target/all-feature Clippy: passed
- full workspace test suite: passed, with only the two unchanged ZFS allocation/reflink tests filtered on this ZFS host
- mutation candidates: 37 total
- caught: 32
- unviable: 5
- missed: 0
- timed out: 0

The original mutation failure exposed real test-coverage debt around configuration predicates and daemon lifecycle branches. The follow-up extracts those predicates into pure helpers with truth-table tests and adds a real server_main bind/shutdown lifecycle test. Upstream CI remains the authoritative full baseline on its normal filesystem.

jmagar:
@jleni Thank you for the detailed review. Addressed all three blockers and the TUI rationale in a3255ec7d6f64a5dd429457bef878fadd62d4b6a:

1. Persistent-daemon configuration
   - CI guidance now leads with cache.prefetch_enabled = false in the watched config file.
   - Both CI and configuration docs state that KACHE_PREFETCH_ENABLED changes require kache daemon restart.
   - The docs explicitly warn that client-side stats can otherwise reflect the new env while an existing daemon keeps its old policy.

2. Stale negative-index authority
   - Negative entries are now authoritative for min(refresh interval × 5, 300 seconds).
   - Increasing the LIST interval can no longer suppress exact remote HEAD checks beyond the original five-minute maximum.
   - refresh_secs = 0 is documented and tested as one initial positive/fallback snapshot, no periodic refresh, and no authoritative negative entries.

3. Warming barrier
   - Manifest startup now goes through one start_manifest_warming helper used by server_main.
   - No remote or prefetch disabled both release the warming barrier immediately.
   - Added a direct remote-configured + prefetch-disabled barrier test, plus the existing real server_main bind/shutdown lifecycle test.

Also restored the positional-swap rationale for the two Option<Vec<String>> TUI-preserved fields.

Validation on the rebased branch:

- workspace check: passed
- strict all-target/all-feature Clippy: passed
- full workspace test suite: passed; only the two unchanged ZFS FIEMAP/reflink tests were filtered on this ZFS host
- focused review-fix mutation gate: 9 total, 5 caught, 4 unviable, 0 missed, 0 timed out
- cumulative PR mutation work: no missed mutants

The new Actions run will require maintainer approval again because this is still a first-time fork contribution.

emmanuelm41 (commented):
Really strong PR, and a great first contribution — the gating is careful, the tests are mutation-sensitive, and the docs are thorough. I read the daemon gating, config precedence, TUI round-trip, and stats paths, and cross-checked with an independent review pass.

**Safety:** CLEAR — no workflow/CI changes; no new attack surface.

**Purpose:** ACCOMPLISHES — `prefetch_enabled = false` skips manifest warming, build-start planning, direct prefetch, and the whole-remote key-cache LIST, while exact-key remote checks and uploads stay on separate paths (verified `handle_upload` and the exact-check path are ungated). `refresh_secs = 0` does one population and never treats negatives as authoritative. The warming barrier is released synchronously when speculation is off, so exact checks never pay the grace wait.

**Tests:** COVERED — truth tables for the gating predicates, plus disabled-path tests for prefetch requests, build-started, and the warming barrier. Two optional depth adds if you feel like it: (1) assert `KACHE_REMOTE_KEY_CACHE_REFRESH_SECS=abc` falls through to the file/default (you test the bool helper's `maybe`, but not the numeric fall-through), and (2) a server-startup test with `remote = Some` **and** `prefetch_enabled = false` to exercise the `cache_handle` skip end-to-end (right now `server_main_binds_...` uses `remote = None`, so the changed branch isn't observed).

**Docs:** REQUIRED & PRESENT — comprehensive, and the refresh semantics match the code. One tiny precision note, non-blocking: "exact misses fall through to a remote HEAD probe" is true except during the *pre-existing* remote-HEAD circuit-breaker window (after 3 consecutive HEAD failures, the daemon deliberately treats keys as misses for 45s to avoid hammering a down remote). That's existing daemon behavior unrelated to your change, and a miss there is just a safe recompile — so please **don't** touch the circuit breaker; at most you might add a half-sentence caveat if you want the docs airtight.

Nothing blocking on my side. Nice work.

emmanuelm41 (approved):
CI is green across all 8 required checks (including Mutation testing). Verdicts are clean — Safety CLEAR, Purpose ACCOMPLISHES, Tests COVERED, Docs PRESENT — and jleni's three blockers are all addressed in a3255ec7. Approving. Thanks for the thorough work, and welcome!

PR #650, opened 2026-08-03
fix(sharing): never report a partial or unvalidated probe as authoritative

jleni:
Follow-up to #641, which merged before this review landed. These are three violations of the
one safety property `src/sharing.rs` documents for itself:

> Every probe failure degrades to "not shared, all bytes private" — exactly what the
> `nlink`-only code returned.

All three occur on a **successful** syscall whose reply the code then over-trusts, which is
why none of them show up as an error path.

## 1. FIEMAP batch-cap exhaustion reported a prefix as the whole map

The loop runs at most 64 batches of 32 extents. If a file has more than ~2048 extents and no
processed extent carries `FIEMAP_EXTENT_LAST`, the loop simply expired and returned the
accumulated prefix as authoritative.

- `private_bytes` is understated, so `clean` promises less reclaimable space than a delete
  actually frees.
- A genuinely reflinked file whose shared extents sit past the cap is reported unshared.

The cap is a sensible runaway guard; exhausting it is just not permission to answer from a
prefix. It now returns `unknown_for(size)`.

## 2. Nothing validated the shape of a batch

- A zero-length extent, or one starting behind the walk, leaves the window offset unmoved, so
  the next batch re-reads the same region and counts it twice. `saturating_add` then hid the
  double-count behind a plausible total.
- `fm_mapped_extents > fm_extent_count` was clamped with `.min()`, silently accepting a reply
  the code does not understand.
- Accumulators could saturate.

Now: forward progress is required per extent, an over-large batch count is rejected, and both
accumulators use `checked_add`.

## 3. `getattrlist`'s reply length was never read

`Buf::length` was declared and never checked. A short-but-successful reply left the zeroed
tail to be parsed as real values — and a zeroed `private_size` reads as "deleting this frees
nothing", which is the one direction that costs a user bytes they could have reclaimed.

## Testing

`extent_is_usable` is extracted next to the existing `extent_is_shared` / `extent_is_last`
predicates, for the reason the file already gives for those: a real kernel will not produce
these shapes, so synthetic values are the only way to cover them. The new test pins each
rejected shape (zero length, backwards, wrapping) and the boundary where an extent starts
exactly where the walk is.

**Linux verified this time, not reasoned about.** `cargo zigbuild --target
aarch64-unknown-linux-gnu` under `RUSTFLAGS="-D warnings"` builds clean — the same gate that
caught the dead-code errors on #641, now reproducible locally rather than only in CI. macOS:
`fmt --check`, `clippy --workspace --all-targets -- -D warnings`, 1458 tests.

## Deliberately not fixed here

A cross-model review also argued the macOS parser should be rewritten to walk the payload by
cursor, on the theory that `getattrlist` packs a partial attribute set contiguously — so a
reply carrying `EXT_FLAGS` but not `PRIVATESIZE` would be read at the wrong offset.

That rests on a claim about Apple's ABI I could not confirm, and rewriting the parser against
an unverified layout would trade a possible bug for a certain one. The length check in (3)
blocks the short-reply case; settling the subset case needs a real partial response from a
filesystem that returns one. Worth its own issue.

Same review flagged a hardlink concern — `nlink > 1` sets `shared` but leaves `private_bytes`
at full size. I checked: `artifact_sharing`'s two call sites both read only `.shared`, and the
sole `private_bytes` consumer sits in the `nlink == 1` branch, so it does not bite today. It is
a latent trap if anyone starts consuming `private_bytes` from that path.

### DIFF
--- a/src/sharing.rs
+++ b/src/sharing.rs
@@ -269,6 +269,27 @@ fn extent_is_last(fe_flags: u32) -> bool {
     fe_flags & FIEMAP_EXTENT_LAST != 0
 }
 
+/// Did the kernel report more extents than it was given room for?
+///
+/// Extracted for the same reason as its siblings: a real kernel cannot produce
+/// this, so a synthetic count is the only way to cover it. Clamping with `min`
+/// instead — the original shape — silently accepts a reply we do not
+/// understand and reports it as authoritative.
+#[cfg(target_os = "linux")]
+fn batch_count_is_valid(mapped: usize, capacity: usize) -> bool {
+    mapped <= capacity
+}
+
+/// Did this batch advance the walk past where it started?
+///
+/// A batch that ends where it began means the next window re-reads the same
+/// region and counts it twice. The boundary is the whole point: equal offsets
+/// are NOT progress.
+#[cfg(target_os = "linux")]
+fn batch_made_progress(offset: u64, batch_start: u64) -> bool {
+    offset > batch_start
+}
+
 /// Is this extent a usable, forward-progressing piece of the map?
 ///
 /// Its own function for the same reason as the flag tests: the failure shapes
@@ -372,7 +393,7 @@ fn probe_linux(path: &Path, size: u64) -> Sharing {
         }
         // The kernel must never report more extents than it was given room for.
         // Clamping would silently accept a reply we do not understand.
-        if mapped > FIEMAP_MAX_EXTENTS {
+        if !batch_count_is_valid(mapped, FIEMAP_MAX_EXTENTS) {
             return Sharing::unknown_for(size);
         }
 
@@ -410,7 +431,7 @@ fn probe_linux(path: &Path, size: u64) -> Sharing {
             complete = true;
             break;
         }
-        if offset <= batch_start {
+        if !batch_made_progress(offset, batch_start) {
             return Sharing::unknown_for(size);
         }
     }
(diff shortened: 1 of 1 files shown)


jleni:
11/11 green except `Mutation testing`, which is blocked on something structural rather than on
this diff. Recording the analysis rather than pushing a fourth attempt.

Three rounds, and the survivor kept moving:

| round | survivor | outcome |
|---|---|---|
| 1 | `>` / `>=` at the batch-count guard, `<=` at the progress guard | extracted to predicates, killed |
| 2 | `delete !` in `if !complete` | moved the branch into `fiemap_verdict`, killed |
| 3 | `delete !` in `if !batch_made_progress(…)` | created by round 2's extraction |

Extracting `batch_made_progress` killed the mutants inside the predicate but converted an
inline comparison into a new `!` at the call site. The call site is in `probe_linux`, which no
test can reach: it needs a real FIEMAP filesystem and the lane runs on ext4. So the mutant was
relocated, not removed.

The one that genuinely died is instructive. `!complete` died because the *branch* moved into a
tested function, not just the predicate. Everything still surviving needs an `if` somewhere
inside `probe_linux`, and nothing inside `probe_linux` is observable, so extraction cannot
finish this job.

Options, in order of how much they actually fix:

1. **Inject the ioctl** behind a trait or closure so the whole walk runs against a fake kernel.
   Kills the class permanently and makes every guard testable. It is a real refactor of code
   that merged two days ago.
2. **Narrow the mutants.toml exclusion** to cover the `probe_linux` driver loop. #641 excluded
   `probe_macos` and `probe_unsupported` on exactly the "this lane cannot run it" argument; the
   driver loop has the same property in practice, though it does compile, which is why #641
   deliberately kept it in scope. This partially reverses that decision.
3. **Give the Linux lane a btrfs loopback** so `probe_linux` is genuinely exercised. Best
   coverage, and it would also close the "Linux path untested at runtime" gap this PR inherited
   from #641. Clearly its own piece of work.

The correctness fixes in this PR stand on their own and are verified independently: Linux
`-D warnings` clean for both the binary and `--tests` via cargo-zigbuild, plus macOS fmt,
clippy, and 1458 tests. The mutation gate is asking a fair question that this diff cannot
answer alone.

Issue #330, opened 2026-06-13
research(cache): path-invariant rmeta digest for cross-clone/worktree key convergence


jleni:
## Problem

Across worktrees / checkouts at different absolute paths, an upstream crate's `.rmeta` embeds absolute source paths, so its bytes differ even when nothing semantically changed. kache content-hashes the **verbatim** `.rmeta` into every dependent's `extern:<name>=<hash>`, so the divergence cascades up the dependency graph and keys do not converge across worktrees. Input-flag normalization cannot reach this — the path lives **inside** the artifact.

On the substrate bench, ~246/943 crates diverge cross-clone, capping warm speedup at ~1.06×. Related: #260.

## Direction (research)

- Derive the extern digest from a **path-masked view** of the `.rmeta` (canonicalize/strip embedded source-path strings) rather than the verbatim bytes.
- Gate strictly: rustc-version-pinned normalization + **mandatory validate-on-hit** (compare produced outputs) so a too-coarse digest can only cause a recompute, never a false hit.
- A merge-on-read manifest then absorbs residual divergence.
- Prototype behind a default-off flag with a **zero-false-hit harness** (build at path A, restore at path B, diff) on the substrate / Firefox benches before any default-on.

The C/C++ path already converges via prefix-map sentinels; this is the Rust frontier. Marked research: needs a rustc-version-pinned, conservatively-validated normalization before it can be trusted.

AndreasKarg:
I think I hit the mirror image of this bug, and it might be useful evidence for why **validate-on-hit** needs to be mandatory rather than optional.

**Setup:** `RUSTC_WRAPPER = kache`, local store on a separate drive, shared across several independent (non-workspace) Cargo projects on the same machine. I originally hit this via real projects — registry crates like `khronos_api`, `serde_core`, `serde`, `thiserror`, `glutin_egl_sys`, `glutin_wgl_sys`, `rustversion`, `rspolib`, `rav1e`, and `i_slint_compiler`, all pulled in transitively via `slint` — but I've reduced it to a minimal, fully synthetic repro below so I'm not posting internal project paths.

[...]

Issue #167, opened 2026-05-27
CARGO_MANIFEST_DIR env!() restores stale rlib across worktrees


DaniPopes:
`kache` can restore an rlib from another checkout when a dependency embeds `env!("CARGO_MANIFEST_DIR")`. The restored artifact keeps the first checkout's manifest path as a runtime string.

This is a concrete repro for the bug class described in #79, but it is easier to trigger because the stale value crosses a normal workspace dependency boundary.

Minimal workspace:

```toml
# Cargo.toml
[workspace]
members = ["helper", "consumer"]
resolver = "3"
```

```toml
# helper/Cargo.toml
[package]
name = "helper"
version = "0.1.0"
edition = "2024"

[lib]
path = "src/lib.rs"
```

```rust
// helper/src/lib.rs
#[inline(never)]
pub fn manifest_dir() -> &'static str {
    env!("CARGO_MANIFEST_DIR")
}
```

```toml
# consumer/Cargo.toml
[package]
name = "consumer"
version = "0.1.0"
edition = "2024"

[dependencies]
helper = { path = "../helper" }
```

```rust
// consumer/src/lib.rs
pub fn helper_manifest_dir() -> &'static str {
    helper::manifest_dir()
}
```

```rust
// consumer/tests/manifest.rs
use std::path::Path;

#[test]
fn helper_manifest_dir_matches_this_checkout() {
    let embedded = Path::new(consumer::helper_manifest_dir()).canonicalize().unwrap();
    let expected = std::env::current_dir()
        .unwrap()
        .parent()
        .unwrap()
        .join("helper")
        .canonicalize()
        .unwrap();
    assert_eq!(embedded, expected);
}
```

Repro:

```sh
a=$(mktemp -d /tmp/kache-manifest-ws-a.XXXXXX)
b=$(mktemp -d /tmp/kache-manifest-ws-b.XXXXXX)
cache=$(mktemp -d /tmp/kache-cache-ws.XXXXXX)

# Put the workspace above in "$a", then:
cp -a "$a/." "$b/"

KACHE_CACHE_DIR=$cache CARGO_TARGET_DIR=$a/target RUSTC_WRAPPER=kache \
  cargo test -q -p consumer --manifest-path "$a/Cargo.toml"

KACHE_CACHE_DIR=$cache CARGO_TARGET_DIR=$b/target RUSTC_WRAPPER=kache \
  cargo test -q -p consumer --manifest-path "$b/Cargo.toml"
```

[...]

PR #688, opened 2026-08-08
fix(key): crate-scoped force entries in path_only_env_vars (#687)

### LINKED ISSUE #687: OUT_DIR env-dep kept absolute in key for sys crate whose build script emits link-search into OUT_DIR — one permanent miss cascades through extern-content hashes
## Summary

On a 383-crate Tauri workspace (kache 0.13.0, macOS arm64, rustc nightly), one third-party unit — `cef-dll-sys`, whose CMake build script emits `cargo:rustc-link-search` into its own OUT_DIR — gets `env_dep:OUT_DIR=<abs path> (kept absolute runtime path)` in its cache key. Trace excerpt:

```
[key:cef_dll_sys] env_dep:OUT_DIR=/private/tmp/kache-w6/target/debug/build/cef-dll-sys-931509eee85f0f82/out (kept absolute runtime path)
[key:cef_dll_sys] link_search:<WORKSPACE>/target/debug/build/cef-dll-sys-931509eee85f0f82/out/build/libcef_dll_wrapper
[key:cef_dll_sys] link_lib_content:static=cef_dll_wrapper=<hash>
```

Note `link_search` is normalized to `<WORKSPACE>`, but the OUT_DIR env value is kept raw — so the key differs per checkout path and the unit misses across worktrees forever. `[cache] path_only_env_vars = ["OUT_DIR"]` did not change this (tried `~/.config/kache/config.toml` and `KACHE_CONFIG=`; the `(kept absolute runtime path)` annotation persists).

## Why one unit matters: the cascade

We made everything else path-stable (removed all `env!("CARGO_MANIFEST_DIR")` from lib code; `-ffile-prefix-map` + `ZERO_AR_DATE=1` make the CMake staticlib SHA-256-identical across worktrees — verified). With that, the *outputs* of `cef_dll_sys` are byte-identical across worktrees — kache itself records `dup` events for it (same bytes, different key).

But because the key differs, each worktree compiles it locally, and the locally-compiled rlib differs from the restored one in exactly ~79 bytes (SVH region — plausibly because the injected `--remap-path-prefix=<worktree>=<sentinel>` value differs per worktree and rustc folds remap args into the SVH). Downstream crates hash extern *content* raw, so this single unit re-poisons the entire dependent tree: measured 229-230 first-party misses (~1,900s of compile work) on every fresh worktree, with per-crate `why-miss` showing "same config … likely source code, dependency, or rustc version change" all the way up.

## Ask

Either normalize path-valued env deps under `<WORKSPACE>` consistently with `link_search` (or honor `path_only_env_vars` for OUT_DIR in this shape), or use the existing dup detection to serve content-equal artifacts across key variants for the env-only-differing case.

## Environment

kache 0.13.0 (brew), macOS 15 arm64 (M4 Max), rustc 1.97.0-nightly, cargo workspace with 383 members, `RUSTFLAGS='--cfg tokio_unstable'` via config.

Happy to run instrumented builds or test a patch — the reproduction is deterministic on our repo.

valorkin:
Closes #687 (or rather: offers the escape hatch discussed there).

A `crate_name:VAR` entry in `path_only_env_vars` bypasses the include-proof and runtime-value-use source scans for exactly that (crate, var) pair — an explicit user assertion that the compiled artifact does not observably embed the value. Plain entries keep the existing scan-gated semantics unchanged.

Motivating case (full evidence in #687): cef-dll-sys uses `env!("OUT_DIR")` in a fallback branch of `get_cef_dir()` that deployments with `CEF_PATH` set never execute. The conservative cfg-blind scan keeps its OUT_DIR absolute in the key, so the unit re-keys per checkout path; its locally-compiled rlib then differs in the SVH bytes (the per-worktree remap flag value), and extern-content hashing cascades the miss through every dependent — measured ~230 first-party crates (~30 min compile work) per fresh worktree on a 383-crate workspace.

Live result with this patch + `cef_dll_sys:OUT_DIR`: a fresh worktree at a never-seen path went from ~15 min (230 misses) to 4m10s with 1,677 of 1,682 units hitting.

Forced values normalize through the same #330 OUT_DIR sentinel / prefix rules as scan-approved values; the trace decision reads `forced path-only (user-asserted)` for observability. Two policy tests added mirroring the existing shapes (force overrides the scan; crate scope does not leak to other crates).

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -1015,8 +1015,13 @@ pub fn compute_cache_key(
 
         hasher.set_group("env_deps");
         for (var, val) in &dep_info.env_deps {
-            let normalized_env_dep =
-                normalize_env_dep_value(var, val, &dep_info.source_files, path_normalizer);
+            let normalized_env_dep = normalize_env_dep_value(
+                crate_name,
+                var,
+                val,
+                &dep_info.source_files,
+                path_normalizer,
+            );
             fold_field(&mut hasher, b"env_dep_var:", var.as_bytes());
             fold_field(
                 &mut hasher,
@@ -1493,6 +1498,9 @@ enum EnvDepNormalizationDecision {
     Unchanged,
     NormalizedPathOnly,
     KeptAbsoluteRuntimePath,
+    /// Normalized because the var (optionally crate-scoped) is in the
+    /// user-asserted force list, bypassing the source scans.
+    ForcedPathOnly,
 }
 
 impl EnvDepNormalizationDecision {
@@ -1501,6 +1509,7 @@ impl EnvDepNormalizationDecision {
             Self::Unchanged => "unchanged",
             Self::NormalizedPathOnly => "normalized path-only",
             Self::KeptAbsoluteRuntimePath => "kept absolute runtime path",
+            Self::ForcedPathOnly => "forced path-only (user-asserted)",
         }
     }
 }
@@ -1658,7 +1667,30 @@ fn clean_static_lib_name(spec: &str) -> Option<&str> {
     Some(name)
 }
 
+/// The normalized key value for a path-only env dep: the `<OUT_DIR:unit>`
+/// sentinel form when the value lives under the build's own OUT_DIR
+/// (kunobi-ninja/kache#330 — the unit-hash component stays observable), else
+/// the generic prefix-rule normalization.
+fn sentinelized_env_dep_value(resolved: &str, normalized: &str) -> String {
+    if let Some(rel) = out_dir_relative_suffix(resolved) {
+        let unit = std::env::var_os("OUT_DIR")
+            .map(std::path::PathBuf::from)
+            .as_deref()
+            .and_then(|p| p.parent().and_then(|d| d.file_name().map(|n| n.to_owned())))
+            .map(|n| n.to_string_lossy().into_owned())
+            .unwrap_or_default();
+        if rel.is_empty() {
+            format!("<OUT_DIR:{unit}>")
+        } else {
+            format!("<OUT_DIR:{unit}>/{}", rel.trim_start_matches('/'))
+        }
+    } else {
+        normalized.to_string()
+    }
+}
+
 fn normalize_env_dep_value(
+    crate_name: &str,
     var: &str,
     val: &str,
     source_files: &[std::path::PathBuf],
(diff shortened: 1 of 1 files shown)


jleni:
Thanks for this, and for the writeup in #687. The keytrace excerpts and the observation that kache already records `dup` events for the unit made the cascade easy to follow. A 230-crate blast radius from one env dep is worth having on record on its own.

The design is right: scoped, opt-in, off by default, and forced values route through the same #330 sentinel rules. `sentinelized_env_dep_value` is a clean extraction. I would like to land this.

One thing needed to get CI green: `cargo fmt`. A few of the updated test call sites are over the line limit, so `Check (Linux)` fails before anything else runs.

Everything else I noticed is follow-up material and I am happy to take it in a separate PR rather than bounce it back to you:

- The force branch bypasses all three conditions of `env_dep_is_safe_to_normalize`, not only the two source scans, so `somecrate:CARGO_MANIFEST_DIR` would re-open #167 (stale rlib across worktrees). Harmless unless someone force-lists that specific var, but worth refusing outright.
- A crate-name typo (`cef-dll-sys` vs `cef_dll_sys`) is a silent no-op, which is a rough failure mode for an escape hatch people reach for mid-debugging.
- The comment above the force check is duplicated, and its first line says plain `VAR` forces, which the code does not do.
- Worth a line in the config reference, plus a forced variant of `e2e-out-dir-runtime` as an end-to-end guard.

On CI: the run is approved and going now. Both Windows jobs are pinned to our self-hosted runner with no fork fallback, so they cannot pick up a fork PR and will sit unclaimed; macOS falls back to a hosted runner, so Linux and macOS results will be real. Windows coverage is ours to sort out, not yours.

Locally the two new tests pass and clippy is clean. `cargo fmt --check` is the only gate that fails.

jleni:
CI came back and it is a good board: Linux and macOS e2e, macOS tests, Nix, audit and version consistency all pass. The only genuine red is `Check (Linux)` on `cargo fmt`; `Mutation testing` failing is just the cascade from that, not something to chase.

So a single `cargo fmt` commit should give a clean run, minus the two Windows jobs which cannot pick up a fork PR at all. Those are on us.

Note that each push re-triggers the fork approval gate, so there may be a short wait before the new run starts.

valorkin:
Pushed `b0f24d2`: `cargo fmt` across the touched files, plus the duplicated force-gate comment removed — you were right that its first line described a plain-`VAR` force the code doesn't have; the stale block was a leftover from an earlier design. Policy tests still green locally (9/9 in the env_dep_policy group).

On the follow-ups: happy to send them as a separate PR so this one stays reviewable — concretely (1) refuse `*:CARGO_MANIFEST_DIR` force entries outright (the #167 hazard you called out), (2) warn on force entries that match no crate all run (the silent-typo failure mode), (3) the config-reference line + a forced variant of `e2e-out-dir-runtime`. Say the word if you'd rather take them yourself; otherwise I'll open it once this lands so it can build on the merged shape.

And thanks for the note on the Windows runners — understood that those two are unclaimable for fork PRs.

jleni:
I rebased this contribution onto current `main` and added the mutation-sensitive trace-label coverage plus deterministic environment-test isolation.

GitHub accepted the maintainer rebase but denied direct pushes to this organization-owned fork despite maintainer edits being enabled, so I published the same commits—with @valorkin's authorship preserved—as maintainer delivery PR #714.

I will keep this PR open until #714 reaches terminal CI and merges, then close it as superseded.

valorkin:
@jleni hey, sent you an invite with write access

jleni:
Merged through maintainer delivery PR #714 as `737638ac`, with both original commits preserved in the review history and Dmitriy credited in the squash commit's `Co-authored-by` trailer.

Closing this PR as superseded. Thank you, @valorkin.

PR #693, opened 2026-08-08
perf(key): memoize the rustc dep-info pre-pass (avg key 177ms → 1.8ms on a 288-unit graph)

valorkin:
## Problem

Per-hit overhead is dominated by key computation, and profiling shows it is almost entirely the `rustc --emit=dep-info` subprocess run per unit: with new phase instrumentation, `dep_info_ms` = 26.2 of 27.4ms `key_ms` (96%) on trivial crates, up to 2.5s on proc-macro-heavy units. (The file-content hashing itself was already fine — the existing `(path,size,mtime_ns,ctime_ns,inode)` SQLite memo and daemon HashFiles path measure ≈0ms.)

## Fix

A ccache-direct-mode-style memo of the pre-pass *result* (`src/dep_info_memo.rs`): one content-addressed JSON record per invocation shape under `<cache_dir>/depinfo/` (your probe-cache pattern, atomic temp+rename), digest = blake3(schema, `rustc -vV`, normalized cwd, normalized filtered dep-info argv minus cosmetic flags).

**Reuse is validation-gated, never trusted**: every recorded source file and extern artifact (proc-macro dylibs included) must match its recorded content hash via the existing FileHasher memo, and every env-dep value must compare equal (path-normalized, so OUT_DIR-style values port across checkouts). Any mismatch or IO failure falls back to the real pre-pass — a stale record can cost one reuse, never produce a wrong key. Records store paths in key-sentinel form with a new `PathNormalizer::denormalize_candidates`, so they port across worktrees exactly as far as keys already do. Key-invariance is pinned by two rustc-gated e2e tests (memo-on byte-equals memo-off keys, both recording and reuse passes).

Config: `[cache] dep_info_memo` / `KACHE_DEP_INFO_MEMO`; lazy GC prunes records idle >30 days. It currently defaults ON in our fork — happy to flip to default-OFF to match your convention for new features.

## Measured (kache's own report JSON, back-to-back, same machine)

| | before | after |
|---|---|---|
| avg key_ms (real 288-unit graph incl. proc macros) | 177.2 | **1.8** |
| avg hit overhead ms | 182.6 | 6.5 |
| total key ms per restore | 51,031 | 513 |
| p50 / p90 per event | 30.5 / 104ms | 1.0 / 5ms |

Synthetic 100-leaf workspace: 25.5 → 0.5ms avg key. Suite: 1,760 passed, 0 failed; 13 memo unit tests + 3 denormalize + 2 e2e invariance tests. Benchmark scripts committed (`scripts/bench-hit-overhead*.sh`).

## Known soundness boundary (shared with ccache direct mode and with existing key computation)

Inputs rustc does not report — e.g. files a proc macro reads without `tracked_path` — are invisible to validation, exactly as they are invisible to the key today. Documented in the module.


jleni (changes requested):
Thanks for the substantial investigation. I cannot approve this implementation, and I am closing the current PR rather than asking for a rebase.

This revives the memo design closed as not planned in #334, defaults it on, and omits the safety envelope that discussion required. A concrete false hit: memoize `mod foo;` with `foo.rs`, then add `foo/mod.rs`. The recorded files still hash equally, so this patch can restore the old artifact; a real dep-info pre-pass now fails E0761.

The same closure problem exists for proc-macro filesystem/env probes, target JSON changes, pathless `--extern` plus `-L`, inner-rustc/env identity, and unset-versus-empty env state. Dropping diagnostic flags can also mask invalid rustc arguments. The record format is unauthenticated/unbounded, accepts record-supplied paths/env names, persists plaintext env values, and is not cleared/accounted by `kache clear`.

This branch also bundles old #687 work now landed safely through #714 and conflicts with it. Its stale workflow will not be approved.

Smallest viable restart: an instrumentation-only benchmark PR using corrected methodology; then reopen #334 only if the measured value justifies a default-off, no-proc-macro/no-pathless-extern design with negative probes, directory/symlink identity, sampled verification, hostile-record bounds, and a soak period.

valorkin:
@jleni ok, let me dig into it

PR #694, opened 2026-08-08
feat(policy): incremental_crates force-list — always-passthrough with rustc incremental for named crates

valorkin:
## Problem

For a huge hot crate, `KACHE_ADAPTIVE_INCREMENTAL` cannot engage in a real edit loop. From `src/incremental_policy.rs` (LEARNING_WINDOW_SECS=60, ACTIVE_IDLE_SECS=30, MAX_ACTIVE_LEASES=8) and the wrapper call sites: observations are stamped at compile *end*, so a ~98s crate compile plus link plus human edit time always overruns the 60s learning window — the policy re-learns forever and every iteration pays the full non-incremental compile. Even a scripted `--lib`-only loop that stays inside the windows re-probes every ≤10 iterations because `active_leases` increments per lease and never decrements. Measured on a 383-crate Tauri workspace whose top unit is ~312k LOC: 97.9s per edit iteration under kache vs a 6.4s incremental floor.

## Fix

`[cache] incremental_crates` / `KACHE_INCREMENTAL_CRATES` (comma/whitespace separated, env replaces file list, `-`/`_` normalized): listed crates ALWAYS pass through with `-C incremental` preserved — no strip, no lookup, no store — regardless of the adaptive heuristic, with a preserved per-crate incremental dir so the passthrough compiles are genuinely incremental across iterations. The read-only-restore interaction is handled (a force-listed crate never restores, so plain incremental rustc never collides with reflinked read-only artifacts).

## Measured (live, same workspace)

- Forced `tap_lib`: edit-loop iterations settle at the **6.4s incremental floor vs 97.9s** unforced (15x per iteration).
- Cold restores of everything else unaffected: eviction test after deleting an unrelated unit's artifacts → 98 local_hits + 1 forced passthrough, dependent cascade all hits.
- Suite: 1,624 unit tests + all 12 integration suites, 0 failures; `incremental_passthrough_test` 13/13 including 3 new force-list tests.

Rationale: kache shouldn't try to beat incremental rustc on the crate being edited — it should cache the 1,681 crates you didn't touch and hand the one you did to incremental. This config makes that policy expressible.


jleni (changes requested):
This needs redesign before a rebase/CI run:

- The branch includes old #688 commits; after #714 it must be restacked to retain only the incremental-policy work.
- A crate-name match currently bypasses Kache even when `--incremental` is absent, disabling cache and fallback for no benefit.
- Unsafe/non-sibling incremental paths fail open to Cargo's shared profile directory.
- The early return bypasses `AdaptiveUnit` identity, isolation checks, leases, exclusions, and hidden-input guards. Reuse the guarded per-unit `AdaptiveUnit`/Lease machinery; never run unisolated state.
- Docs must say matching uses rustc `--crate-name`, not necessarily the Cargo package name.

I will handle the eventual restack, so no mechanical rebase is needed from you. Please first narrow the design to the guarded adaptive lane and add no-incremental, unsafe-path, hidden-input, and concurrent-lease tests. The stale fork run will not be approved.

jleni:
Maintainer adoption is now tracked in https://github.com/kunobi-ninja/kache/pull/716. It preserves Dmitriy Shekhovtsov's contributor authorship and carries the reviewed fixes on a repository-owned branch. This original PR remains open while the replacement completes hosted CI and review.

jleni:
Adopted and merged via #716 at 14cd3dc221d595b816b31ce26e76c178608b3d8f after full GitHub-hosted CI. Contributor authorship is preserved in the replacement history and merge metadata. Thank you for the contribution; closing this superseded source PR.

PR #701, opened 2026-08-08
fix(native): parse BSD/Darwin archives in portable_static_archive_hash (#691)

### LINKED ISSUE #691: portable_static_archive_hash: BSD/macOS archives always fall back to raw bytes, re-keying cc-built staticlibs per checkout
## Summary

`portable_static_archive_hash` (src/native_archive.rs) parses GNU archives only: any archive with `#1/N` extended names or a `__.SYMDEF`/`__.SYMDEF SORTED` member returns `None`, and `compute_static_lib_hash` falls back to whole-file blake3. On macOS, `ar`/libtool always produce BSD archives, so every `link_lib_content:static=...` key component hashes raw archive bytes — the member-name portability the GNU arm provides is unreachable on the platform that needs it most.

## Measured failure

`rquickjs-sys 0.11.0` copies its quickjs C sources into `OUT_DIR` and compiles them there. cc-rs (1.2.62, `objects_from_files` in src/command_helpers.rs) names each object `{hash(dirname)}-{stem}.o`, where `dirname` is the absolute `OUT_DIR` — not relativizable by cc's `CARGO_MANIFEST_DIR` strip, because `OUT_DIR` lives under the CONSUMER's target dir. Two checkouts of the same commit on one macOS machine produce `libquickjs.a` archives whose five member objects are **byte-identical** while every member NAME carries a different 16-hex prefix (measured: `b3f4cc7072c29bc2-cutils.o` vs `21592bc2ccb88103-cutils.o`, etc.). BSD inline names live in the member data area, so the raw-byte fallback hashes differ and the crate misses cross-checkout forever:

```
[key:rquickjs_sys] link_lib_content:static=quickjs=6e80acf8db6803fa (<checkout-a>/target/debug/build/rquickjs-sys-62f715df837919cc/out/libquickjs.a)
[key:rquickjs_sys] link_lib_content:static=quickjs=2aec79be9a2de23e (<checkout-b>/target/debug/build/rquickjs-sys-62f715df837919cc/out/libquickjs.a)
```

Any macOS crate in this shape — cc-compiled sources staged under `OUT_DIR` — has the same permanent per-checkout miss. `ZERO_AR_DATE=1` does not help: timestamps are already zero; the member names are the sole divergence.

## Suggested fix

Add a BSD/Darwin arm to `portable_static_archive_hash`:

- parse `#1/N` extended names; the inline name occupies the first N bytes of the member data area — exclude those bytes from the hashed content (they are the path-derived name);
- treat `__.SYMDEF` / `__.SYMDEF SORTED` / `__.SYMDEF_64` the way the GNU arm treats the `//` long-name table: ignore its content but fold its length (its ranlib entries reference member offsets; with equal-length name prefixes the offsets converge, and the length fold keeps crafted-archive collisions out, mirroring the reasoning already applied to the GNU long-name table);
- keep the strict parse-else-fallback contract.

The existing GNU-arm test `identical_content_different_member_names_hash_equal` states exactly the property macOS needs.

valorkin:
## Problem

`portable_static_archive_hash` parses GNU archives only, and on macOS `ar`/libtool always emit BSD archives — `#1/N` inline member names in the data area plus a `__.SYMDEF[ SORTED]` ranlib member — so every macOS staticlib fell through to the whole-file blake3. cc-rs (1.2.62) names each object `{hash(absolute OUT_DIR)}-{stem}.o`, and for crates that stage their C sources under `OUT_DIR` (rquickjs-sys) that hash is not relativizable, so two checkouts produce archives with byte-identical member content and divergent inline names. Because BSD names live inside the member data, the raw-byte fallback re-keys `link_lib_content:static=` per checkout forever, missing the crate and everything downstream of it (#691 — the platform half of #471 that v19 deferred).

## Fix

Add a BSD/Darwin arm beside the GNU one in `src/native_archive.rs`, tried after it (`gnu_archive_hash(bytes).or_else(|| bsd_archive_hash(bytes))`), with the same strict parse-else-`None` contract:

- **`#1/N` inline names are excluded from the hashed member content** — they are the path-derived cc names — but each member **folds its inline-name LENGTH**. That is the BSD analog of the GNU `//`-length fold: the ranlib payload stores absolute member offsets that shift with those lengths, so a length change must re-key or a stale/crafted ranlib could point at a different member yet collide. cc names are a fixed-width hash prefix + stem, so the lengths (and the stored offsets) converge across clones.
- **The ranlib member (`__.SYMDEF[_64][ SORTED]`) is hashed as-is**, tagged with its trimmed variant name (`_64` reads identical bytes at a different word width; ` SORTED` changes the lookup contract — the `/` vs `/SYM64/` reasoning). This deliberately mirrors the GNU **symtab** treatment rather than the length-only `//` treatment the issue sketched: measured across two checkouts of rquickjs-sys, the 8,568-byte ranlib payload is **byte-identical** (fixed-width cc names keep every stored offset equal), so hashing it raw is equally portable *and* keeps a stale/crafted ranlib from colliding with one that links differently — the same property the GNU arm documents for its symbol table. The ranlib member must be the first member and appear at most once (where Darwin `ranlib` writes it), mirroring the GNU symtab strictness.
- **Fail-closed exactly like the GNU arm**: mixed GNU/BSD layouts (`/`-led or `/`-terminated names next to BSD members), misplaced/duplicate ranlib members, non-decimal or oversized `#1/N`, truncated data, trailing bytes → `None` → whole-file hash. GNU is tried first, so **every pre-existing GNU digest is unchanged**.
- **Distinct domain** (`kache.native-ar.bsd.member-content.v1`, textual prefix `bsd-ar-v1:`): a BSD-parsed archive never hashes equal to a GNU-parsed archive of identical members — rustc bundles the archive BYTES into its output, so the format difference is an output difference, and the two arms frame different symbol-table semantics. Documented on the entry function.
- **`CACHE_KEY_VERSION` 23 → 24**: BSD static-lib key bytes change; this is the same one-time re-key v19 applied when the GNU arm landed (GNU keys are byte-identical before/after, so the bump is the only cross-platform cost — flagging it explicitly since it invalidates all platforms once).
- **Static-lib digest-cache scheme `static-ar-v1` → `static-ar-v2`**: a v1 row can hold a pre-#691 whole-file digest for a BSD archive, and the fingerprint (size/mtime/inode) of an unchanged file would keep serving that unportable digest forever. Without this bump the fix would silently not apply to already-fingerprinted archives.

## Measured before/after

Probe: two checkouts of a crate depending on `rquickjs-sys = "0.11.0"` at `/tmp/kache-probe-a-wa` and `/tmp/kache-probe-b-wa`, one shared fresh `KACHE_CACHE_DIR`, `CFLAGS=-g0` so the C members are byte-identical across trees and the member names are the *only* divergence (verified with `ar t` + `cmp` per extracted member: 5/5 members identical, names `762023313f39ea17-*.o` vs `d30efccc898219da-*.o`). Same instrument (`KACHE_LOG_FILE="kache=trace"`) for both runs.

**Unpatched (upstream/main, 4961602):**

```
[key:rquickjs_sys] link_lib_content:static=quickjs=293ccba2e5e8ccdd (…/kache-probe-a-wa/…/out/libquickjs.a)
[key:rquickjs_sys] link_lib_content:static=quickjs=4adc1cbf94180964 (…/kache-probe-b-wa/…/out/libquickjs.a)
cache miss for rquickjs_sys, compiling (cf738ae077552200)
```

Tree B: 4 hits / 2 misses — `rquickjs_sys` and its dependent bin re-keyed purely by the member-name divergence.

**Patched (this branch), store wiped, same probes:**

```
[key:rquickjs_sys] link_lib_content:static=quickjs=bsd-ar-v1:3d0abb (…/kache-probe-a-wa/…)
[key:rquickjs_sys] link_lib_content:static=quickjs=bsd-ar-v1:3d0abb (…/kache-probe-b-wa/…)
local cache hit for rquickjs_sys (96bfaabaf6d1b11d)
```

Tree B: **6 hits / 0 misses**. A same-tree wipe-and-rebuild of tree A also hits 6/6 (no warm-path regression), and the restored probe binary runs.

## Tests

- 16 new BSD unit tests in `native_archive.rs`, in the existing byte-exact fixture style: equal-content/different-names equality, content/order/ranlib-content/ranlib-variant/name-length re-keys, short-header-name and odd-size parsing, domain tagging, ranlib placement strictness, malformed/truncated/trailing-garbage fallbacks — plus one macOS-only test that drives the **system `ar`** end-to-end over the `#1/N` path.
- All pre-existing GNU tests pass unchanged. Two mixed-layout tests were renamed with their assertions intact (`bsd_symdef_falls_back` → `bsd_symdef_with_gnu_member_falls_back`, `darwin64_symdef_falls_back` → `darwin64_symdef_with_gnu_member_falls_back` — they always exercised BSD-marker-next-to-GNU-name inputs, which still fall back). The old `bsd_archive_falls_back` test asserted the limitation this PR removes; its input now hashes portably and is covered by the positive BSD tests.
- `cargo test --workspace`: **1921 passed / 0 failed** (26 targets; `kache` bin target 1659 passed incl. the 30 native_archive tests).
- e2e gate suite (`kache-scenario --select suite:e2e --select tier:gate`, the macOS CI gate, run locally on Apple Silicon): **40 pass / 0 fail / 2 platform skips** (`e2e-cc-bench-flags-gnu` Linux-only, `e2e-cc-cl-debug` Windows-only), harness exit 0.
- `cargo fmt --check` clean. `cargo clippy --workspace --all-targets -- -D warnings` with the pinned 1.95 toolchain reports 3 pre-existing errors (`remote_backend.rs`, `tui.rs`, `wrapper.rs`) that reproduce identically on pristine upstream/main with the same clippy 0.1.95 — none are in this diff (verified by running the same command on an upstream/main worktree: identical error set).

## Scope / follow-ups

- This fixes the archive-**name** divergence axis only. With C debug info enabled and the C side not routed through kache's cc remapping, the members themselves embed the absolute `OUT_DIR` in DWARF (`comp_dir`/`.debug_line`) and legitimately differ across checkouts — the portable hash then honestly re-keys on content. That is a cc-side remapping concern, not an archive-parsing one, and is out of scope here.
- `__.SYMDEF_64[ SORTED]` coverage is synthetic (byte-exact fixtures); the system-`ar` test exercises the 32-bit `SORTED` shape macOS emits today.
- Windows COFF `.lib` and thin archives keep the whole-file fallback, unchanged.

Fork PR: CI needs maintainer approval to run; per project convention the Windows CI jobs are the maintainer's to shepherd for fork PRs.

Fixes #691

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -179,7 +179,17 @@ use std::path::{Path, PathBuf};
 // generated file can observe its own remapped path via `file!()`, so units
 // differing only by unit hash must not collide. Bump so v22 entries with the
 // old spelling invalidate cleanly.
-pub(crate) const CACHE_KEY_VERSION: u32 = 23;
+//
+// v24 (kunobi-ninja/kache#691): `-l static=` BSD / Darwin archives now get the
+// build-path-portable member-content hash ([`crate::native_archive`]) instead
+// of the whole-file fallback. macOS `ar` stores the `cc` path-derived member
+// names INLINE in the member data (`#1/N`), so the whole-file hash re-keyed
+// every cc-built staticlib per checkout on macOS (rquickjs-sys: byte-identical
+// members, names differing only in the 16-hex `cc` prefix — the platform half
+// of #471 that v19 deferred). GNU digests are unchanged, but the BSD static-lib
+// key bytes change, so bump to invalidate v23 entries cleanly (the same
+// one-time re-key v19 applied when the GNU arm landed).
+pub(crate) const CACHE_KEY_VERSION: u32 = 24;
 const MIN_PERSISTED_HASH_BYTES: i64 = 64 * 1024;
 
 /// Collapse runs of ASCII whitespace into single spaces and trim
@@ -2075,10 +2085,10 @@ pub fn hash_file(path: &Path) -> Result<String> {
 }
 
 /// Compute a linked `static=` archive's cache-key digest: the build-path-portable
-/// member hash for a GNU archive ([`crate::native_archive`]), else the exact
-/// whole-file blake3 [`hash_file`] produces (so non-GNU archives keep their prior
-/// key, modulo the `CACHE_KEY_VERSION` bump). Reads the file once; caching is the
-/// caller's ([`FileHasher::hash_static_lib`]) concern.
+/// member hash for a GNU or BSD archive ([`crate::native_archive`]), else the
+/// exact whole-file blake3 [`hash_file`] produces (so unparseable archives keep
+/// their prior key, modulo the `CACHE_KEY_VERSION` bump). Reads the file once;
+/// caching is the caller's ([`FileHasher::hash_static_lib`]) concern.
 fn compute_static_lib_hash(path: &Path) -> Result<String> {
     let bytes = std::fs::read(path).with_context(|| format!("reading {}", path.display()))?;
     if let Some(portable) = crate::native_archive::portable_static_archive_hash(&bytes) {
@@ -2417,9 +2427,9 @@ impl<'db> FileHasher<'db> {
 
     /// Hash a linked `-l static=` archive for the cache key. Uses a build-path-
     /// PORTABLE digest that ignores the `cc`-derived archive member names
-    /// (kunobi-ninja/kache#471) when the file is a cleanly-parseable GNU static
-    /// archive; otherwise falls back to the whole-file content hash (BSD / thin /
-    /// COFF / non-archive — correct, just not yet cross-clone portable). The
+    /// (kunobi-ninja/kache#471, #691) when the file is a cleanly-parseable GNU
+    /// or BSD static archive; otherwise falls back to the whole-file content
+    /// hash (thin / COFF / non-archive — correct, just not cross-clone portable). The
     /// too-new guard (#324) is applied either way, and the portable digest is
     /// domain-tagged ([`crate::native_archive`]) so it cannot collide with a
     /// whole-file hash. Scoped to `static=` (this method) on purpose — `.rlib`s
@@ -2459,9 +2469,13 @@ impl<'db> FileHasher<'db> {
         // things — `hash` stores the whole-file blake3, this stores the portable
         // member digest). This restores the warm-build fast path the whole-file
         // hasher had: an unchanged large `static=` archive (e.g. rocksdb) is not
-        // re-read on every incremental build.
+        // re-read on every incremental build. `v2`: a `static-ar-v1` row can
+        // hold a pre-#691 whole-file digest for a BSD archive; the fingerprint
+        // (size/mtime/inode) of an unchanged file would keep serving that
+        // unportable digest forever, so the digest-definition change bumps the
+        // scheme namespace too.
         let key = FileFingerprint {
-            path: format!("static-ar-v1\0{}", fingerprint.path),
+            path: format!("static-ar-v2\0{}", fingerprint.path),
             size: fingerprint.size,
             mtime_ns: fingerprint.mtime_ns,
             ctime_ns: fingerprint.ctime_ns,
--- a/src/native_archive.rs
+++ b/src/native_archive.rs
@@ -12,11 +12,12 @@
 //! cross-clone-missing the lib and everything downstream of it.
 //!
 //! [`portable_static_archive_hash`] hashes the archive's link-relevant content
-//! while ignoring those path-derived member names, for the GNU / SysV `ar`
-//! format. It is deliberately **fail-closed**: anything it cannot parse as a
-//! plain (non-thin) GNU archive returns `None` and the caller falls back to the
-//! whole-file hash — so it can never produce a *wrong* (colliding) hash, only a
-//! less-portable one.
+//! while ignoring those path-derived member names, for the two `ar` flavors
+//! rustc links on kache's Unix targets: GNU / SysV (Linux) and BSD / Darwin
+//! (macOS, #691). It is deliberately **fail-closed**: anything it cannot parse
+//! as a plain (non-thin) GNU or BSD archive returns `None` and the caller falls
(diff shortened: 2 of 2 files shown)


jleni (changes requested):
Thanks—this is a sound, fail-safe direction for #691, and the cache-key version bumps are correct.

Two gaps need closing before it lands:

- The real macOS test uses `ar cqS`; `S` disables the symbol table, so it never exercises the new `__.SYMDEF[_64][ SORTED]` path. Please build real Mach-O objects with equal-length differing names, use `ar crs`/ranlib, and assert both archives are recognized and hash equally.
- Valid BSD archives may contain ordinary slash-terminated short members (`name/`). The current BSD parser rejects all of them as GNU. Accept those while still rejecting `/`, `//`, `/N`, and `/SYM64/`, or narrow the documented support.

Do not approve or rerun the existing stale workflow; it predates hosted-runner routing. I will handle the rebase and maintainer delivery after #714, so no rebase work is required from you.

jleni:
Maintainer adoption is now tracked in https://github.com/kunobi-ninja/kache/pull/717. It preserves Dmitriy Shekhovtsov's contributor authorship and carries the reviewed, fail-closed archive-identity fixes on a repository-owned branch. Issue #691 intentionally remains open because this replacement does not claim cross-checkout portability. This original PR remains open while the replacement completes hosted CI and review.

jleni:
Thank you, Dmitriy. We adopted the contribution into maintainer-owned [#717](https://github.com/kunobi-ninja/kache/pull/717), preserved your authorship, and added the fail-closed archive/linker hardening and mutation coverage required for landing. It landed as signed squash [`af3e19c`](https://github.com/kunobi-ninja/kache/commit/af3e19c0981fcdcc7e0bdb3e5a11085e1aea11f9).

Closing this source PR in favor of the replacement. We reopened issue #691 because the landed change intentionally does not claim cross-checkout portability for path-derived Darwin member names.

Issue #327, opened 2026-06-13
feat(remote): weight S3 permits by pack size so large restores can't saturate the pool


jleni:
## Status

Most of this issue landed in #719 (c3bac93). One item is left, so the scope below is trimmed to it.

## Remaining work: weighted permits

Every S3 operation acquires exactly one permit from `s3_semaphore` regardless of how much it transfers. The pool is built as `Semaphore::new(config.s3_concurrency)` (`src/daemon.rs:2047`) and every acquisition site calls `.acquire()` rather than `acquire_many` (2149, 2767, 2840, 3181, 3325, 3578, 5321, 5497, 5596).

A pool sized for N concurrent operations therefore admits N large pack restores at once. Permits should scale with pack size, capped so that one pack can never require more than the pool holds and deadlock against itself.

### Acceptance criteria

- A restore acquires permits proportional to its expected pack size, capped strictly below the pool size.
- A pack above the cap still runs, holding the capped number of permits rather than waiting forever.
- Small operations (HEAD, key-cache LIST) keep taking a single permit and are not starved behind one large restore.
- The existing `available_permits` accounting test is extended rather than replaced.

## Landed in #719

- Error classification into `{Miss, Transient, Timeout}`, with exponential backoff and per-attempt jitter (`src/remote_resilience.rs`).
- `RemoteBreaker`, generalizing `RemoteHealth` to every remote op: HEAD, GET, PUT, and key-cache LIST all feed it and consult it. While degraded a restore answers miss immediately and rustc recompiles locally.
- Deadline-bounded restore via `remote_restore_timeout_secs`.
- `KeyedSingleflight`, the named per-key lock that replaced the busy-poll download dedup.
- `NegativeKeyCache`, the persistent negative-lookup cache, which also closed #564.

## Out of scope

A detached signature binding a cache key to its manifest, tracked separately in #179 as a distinct trust layer.

jleni:
Line refs refreshed for v0.19.0. Large restores are now partly bounded by the 512 MiB download-memory budget in src/remote_backend/download_memory.rs, which reserves by size with acquire_many_owned. The request pool still takes one permit per operation, so the remaining question is whether that budget leaves a real saturation case. Measure that before building weighted permits.

PR #702, opened 2026-08-08
feat(remote): error classification, download-gating breaker, deadline-bounded restore, negative-result TTL (#327, #564)

### LINKED ISSUE #564: perf(remote): negative-result TTL so parallel wrappers don’t all miss the same key on S3
## Problem

On local miss with remote configured, each wrapper may `send_remote_check` (up to ~3s timeout). Prefetch helps the warm remote path, but when a key is **absent** from S3, many parallel wrappers can still stampede the daemon/S3 for the same missing key.

Part of #552 (P2).

## Proposed solution

Daemon-side **negative cache**:

- After a definitive remote miss, remember `(cache_key → missed_at)` for a short TTL (e.g. 30–120s, configurable).
- Subsequent `RemoteCheck` for that key return miss immediately without S3.
- Invalidate negative entry on successful upload of that key (or any put of that key).

### Implementation hints

- `src/daemon.rs` `RemoteCheck` / restore check path; shared in-memory map with TTL (dashmap / mutex HashMap).
- Don’t negative-cache soft failures (timeouts, 5xx, credential errors) — only definitive “not found”.
- Prefetch path should populate negative or positive coherently.
- Metrics: count negative hits vs S3 round-trips in daemon stats.
- Tests: two checks for missing key → one S3 get; after upload → check hits.

## Alternatives considered

- Client-side negative cache per wrapper process — doesn’t help across processes.
- Longer remote check timeout — makes stampede worse.

## Use case

Cold CI with partial remote fill; local miss storms for crates not yet uploaded.

valorkin:
## Problem

The remote path collapses every non-404 error into a single `Err` (only `NotFound` is special-cased on the HEAD/GET paths), so the daemon cannot tell "the key is absent" from "the remote is down" from "the remote is stalling". The health breaker (`RemoteHealth`) gates only the HEAD probe: a key-cache positive goes straight to GET with no breaker consultation and no deadline, and the upload worker burns a full transport retry budget per queued entry against a dead endpoint — serialized through the same S3 semaphore interactive checks need. Separately (#564), when a key is absent from S3, parallel wrappers stampede the daemon/S3 with repeated checks for the same missing key.

## Fix

New module `src/remote_resilience.rs`, composed by the daemon around every remote op:

- **Error classification `{Miss, Transient, Timeout}`** (`classify_remote_error`): walks the error chain for the typed causes the transports actually produce (`EntryNotFound`, `opendal NotFound`, tokio `Elapsed`, `io TimedOut`, reqwest timeouts), with a message sniff as the fallback for timeouts that arrive as text. `retry_transient` retries **only** `Transient` failures, with exponential backoff + per-attempt jitter (replacing the unconditional single HEAD retry with a static 150 ms sleep — static delays make parallel wrappers retry in near-lockstep). A `Timeout` is never retried at the daemon level: the transport's retry layer already spent its budget getting there, so it goes straight to the breaker.
- **`RemoteBreaker`** generalizes `RemoteHealth` from HEAD-only to every remote op: HEAD probes, restores (GET), uploads (PUT), and key-cache LISTs all feed it and consult it. While degraded, a restore returns not-found immediately and rustc recompiles locally — the cache is an optimization, never a hard dependency; uploads are dropped (the entry stays local; a later build re-queues it), and key-cache refreshes are skipped for the cooldown. Threshold/window semantics are unchanged (3 consecutive failures, 45 s).
- **Deadline-bounded restore**: `remote_restore_timeout_secs` (default **300 s**, `0` = off) wraps the GET+extract in `tokio::time::timeout`; on elapse the future is dropped (cancelling the transfer) and the wrapper gets a miss. The default matches `prefetch_deadline_secs`: the transport's 3.1 s connect and 30 s read-inactivity timeouts already bound dead endpoints, so 300 s only catches slow-drip transfers (at 2 MiB/s that is ~600 MiB — far above any sane pack), i.e. no existing user's legitimate transfer changes behavior.
- **Daemon-side negative-result TTL cache** (#564): after a definitive remote miss (**404 only** — never timeouts, 5xx, or credential errors), `cache_key → missed_at` is remembered for `remote_negative_ttl_secs` (default **60 s**, `0` = off); subsequent checks answer miss without touching S3. A successful upload of the key (or any observation it exists: HEAD hit, fresh LIST) clears the entry immediately. The 60 s default matches `remote_key_cache_refresh_secs`: the daemon already treats LIST data of that age as authoritative for misses, so this introduces no new staleness class. Bounded at 16 K entries (expired pruned first, then oldest evicted).
- **Stats**: `kache stats` gains a `Resilience:` line and the daemon `StatsResponse` gains `remote_check_roundtrips`, `negative_hits`, `negative_entries`, `downloads_suppressed`, `uploads_suppressed`, `remote_degraded` (all `#[serde(default)]`, so old/new client-daemon mixes keep working).

Both knobs follow the existing config conventions (`KACHE_*` env / `[cache]` file key, env wins, preserved by the config TUI's advanced-settings passthrough) and are documented in `docs/getting-started/configuration.mdx`.

## Measured evidence (macOS, M4 Max, APFS)

Instrument: a 31-unit cold probe build (25 small crates.io deps + root) through `RUSTC_WRAPPER=<binary>`, isolated `KACHE_CACHE_DIR`/`KACHE_SOCKET_PATH`/`KACHE_CONFIG`, `cargo build --offline` from a wiped store and target; counters read from the daemon's `Stats` response. Same instrument for baseline (unpatched `upstream/main` @ 4961602) and patched.

**Blackhole endpoint** (`KACHE_S3_ENDPOINT=https://10.255.255.1`, connect timeout):
- baseline: build wall 14.3 s; upload queue **still 21-deep when the 304 s drain cap expired** (10 uploads processed, all failed, ~30 s of HEAD+PUT timeouts each — projected ~13 min to drain).
- patched: build wall 16.3 s (wrapper-side 3 s check timeouts dominate both runs until the breaker trips); queue drained in **12.4 s** with **31/31 uploads suppressed, 0 failed**, `remote_degraded=true`.

**Connection refused** (`KACHE_S3_ENDPOINT=http://127.0.0.1:1`):
- baseline: build 5.5 s; drain **63.5 s**, 31 uploads failed.
- patched: build 5.0 s; drain **0.1 s**, 31/31 suppressed, 0 failed.

**Stalling endpoint** (local Python server: HEAD → 200, GET → accepts and never responds):
- baseline: the restore hung **121.1 s** (30 s read-inactivity × 4 transport attempts), then answered an error.
- patched (`KACHE_REMOTE_RESTORE_TIMEOUT_SECS=5`): answered `found=false` at **5.01 s**, `downloads_failed=1`, key **not** negative-cached (timeouts never are).

**Online regression (real R2 bucket, real credentials, dedicated `wb327-probe` prefix)** with the patched build:
- cold build: 31 HEAD misses → 31 negative entries; 31 uploads completed, 0 failed; mid-drain stats showed `negative_entries=27` with 4 uploads landed and post-drain `negative_entries=0` — live observation of insert-then-invalidate per key.
- wiped store + rebuild (same salt): **31/31 remote hits, 100 % hit rate**, build wall 1.9 s, 0 download failures.
- #564 exact sequence: two checks of one missing key = **1 S3 round trip + 1 negative-cache hit** (first answer 0.07 s, second 0.00 s, `remote_check_roundtrips` unchanged).

## Test coverage

- `remote_resilience`: classifier per class (typed `EntryNotFound`/opendal `NotFound`/`Elapsed`/`io::TimedOut` + text fallback + transient default), backoff growth/cap/overflow, jitter bounds, retry-only-transient (budget, immediate-timeout-return, stop-on-success), breaker threshold/recovery/streak-reset, negative-cache TTL expiry, disable-at-0, upload invalidation, capacity eviction, LIST coherence.
- `daemon`: second check answered from the negative cache without S3 + post-invalidation re-probe; degraded breaker gates the download path (a panic-on-GET backend proves S3 is never touched); restore deadline answers miss at ~the bound and feeds the breaker; degraded `do_upload` drops the job without a PUT; transient HEAD failures retry to the 3-attempt budget while timeouts probe once and three of them degrade the breaker.
- `config`: defaults, file override, env-over-file, `0` disables.

`cargo test --workspace`: **1929 passed, 0 failed** (kache bin 1667, kache-core 32, kache-e2e 90, kache-service 14, integration suites 126). `cargo fmt --all --check` clean. `cargo clippy --workspace --all-targets -- -D warnings` reports zero findings introduced by this diff (my local 1.95.0 clippy flags 3 pre-existing findings in untouched files — `remote_backend.rs`, `tui.rs`, `wrapper.rs` — identically on `upstream/main`, so they appear to be local-toolchain drift rather than CI-visible). Note: `incremental_policy` unit tests flake occasionally under full-suite parallelism on this machine — reproduced on pristine `upstream/main` (1645/1 on one of six runs), unrelated to this diff.

## Scope vs #327 (honest cuts, tracked as follow-ups)

Implemented: classification + jittered backoff, the op-wide breaker with download/upload/LIST gating, the deadline-bounded restore, and the #564 negative cache (in-memory).
**Not** in this PR, deliberately: **weighted permits** (pack-size-scaled semaphore acquisition), the **named per-key lock** replacing the busy-poll download join, and the **persistent (cross-restart) negative cache** — each is a separable change with its own failure modes, and a reviewable PR beats a complete one. They remain tracked by #327's list.

Also: the breaker threshold/window stay compile-time constants (as today) rather than new config; happy to lift them into config if you'd prefer.

Addresses #327 (resilience core). Fixes #564.

---

*This is a fork PR (`ZephyrCloudIO:feat/327-564-remote-resilience`), so CI needs maintainer approval to run; as noted before, the Windows CI jobs are on your side for fork PRs.*


jleni (changes requested):
The resilience direction is useful, but this is not safe to land as one 1,500-line state-machine change. Please split it into negative-cache/singleflight, typed classification+breaker, and end-to-end deadline slices.

Blocking examples:

- Concurrent cold misses still all HEAD S3; the sequential negative-cache test does not solve #564's first-wave stampede.
- Degraded uploads return success and are removed from the pending set, so they can be lost permanently rather than retried.
- Prefetch/LIST/speculative GET paths bypass the breaker and restore budget.
- One global streak lets healthy reads mask broken writes—or broken writes suppress healthy reads—and cooldown has no single half-open probe.
- Local/integrity errors are classified as remote transient failures; auth/config errors also default transient.
- The daemon's 300s restore timeout conflicts with the client's hard 3s RemoteCheck timeout and excludes queue/HEAD/extraction time.
- HEAD-vs-upload races can insert a stale negative after successful publication.

The new env settings also need `ignore_env` coverage. Do not rebase or approve the stale run yet; make the state machine reviewable first.

jleni:
Thanks for the contribution. I have adopted the pinned source head `7fff1fbd4179b674f02fda569dae270113603063` into maintainer-owned replacement #719, preserving your authorship and adding the durability, mixed-version deadline, GC-pinning, and shutdown safeguards found during review. The original PR will remain open until the replacement completes hosted CI and lands.

jleni:
Thank you, Dmitriy. We adopted the contribution into maintainer-owned [#719](https://github.com/kunobi-ninja/kache/pull/719), preserved your authorship, and added the concurrency, deadline, upload-durability, GC-protection, mixed-version, and cross-platform corrections required for landing.

The replacement passed exact-head GitHub-hosted CI and landed as GitHub-verified squash commit [`c3bac93`](https://github.com/kunobi-ninja/kache/commit/c3bac93aa0ec5643a23f86a2b85f174c106bf7d8), including `Co-authored-by: Dmitriy Shekhovtsov <valorkin@gmail.com>`.

Closing this source PR in favor of the replacement.

Issue #652, opened 2026-08-03
prefetch_enabled follow-ups: daemon-side policy reporting, negative-cache trust window, warming-barrier comment


jleni:
Follow-ups from the review of #649, which added `cache.prefetch_enabled` and `cache.remote_key_cache_refresh_secs`. None of these block that PR: the defaults reproduce the previous behavior exactly (60s refresh gives the same 300s authoritative window), so nothing changes for anyone who does not set the new knobs. They matter for the people the feature was written for.

## 1. Disabling prefetch by env var does not reach a running daemon, and `kache stats` says it did

`docs/remote-cache/ci.mdx` tells long-lived self-hosted runners to set `KACHE_PREFETCH_ENABLED=0`. The daemon captures its env once at startup. The only auto-restart triggers are the config file fingerprint watcher in `server_main` and the client-binary-newer check, so an env change never reaches a daemon that is already up. With `DEFAULT_DAEMON_IDLE_TIMEOUT_SECS = 600`, a runner with back-to-back jobs keeps the same daemon and the setting silently does nothing until an idle gap happens to reap it. Intermittent by construction, which makes it worse than a clean no-op.

`render_stats` compounds it: the "Prefetch: disabled" line is computed from the *client's* config, so `kache stats` reports disabled while the daemon is still LISTing and planning.

To do:
- Lead with the file-based `prefetch_enabled = false` in the CI doc. A file change trips the fingerprint watcher and restarts the daemon on its own, so it is the only form of the setting that self-applies.
- Say explicitly that changing `KACHE_PREFETCH_ENABLED` needs `kache daemon restart`.
- Carry the effective policy in the stats snapshot and render that instead of client config, so the line cannot disagree with the daemon. This is the real fix for the reporting half.

## 2. `remote_key_cache_refresh_secs` also scales how long a stale negative index is trusted

`KEY_CACHE_AUTHORITATIVE_MULTIPLIER = 5` makes the negative-cache trust window `refresh * 5`. The configuration doc suggests raising the refresh interval as the cheaper middle ground between full speculation and none. At `900` that means a key uploaded by a peer just after a LIST is answered as a miss with no HEAD probe for up to 15 minutes, and up to 75 minutes when refreshes are failing. On a shared remote with several runners uploading, that is a straight hit-rate loss, and it undercuts the "exact remote lookup remains enabled" framing, since the lookup can be served entirely from a stale snapshot.

[...]

Issue #131, opened 2026-05-23
In-flight build observability: heartbeat, ETA, stuck-process detection


jleni:
## Motivation

When kache spawns rustc for a cache miss, the wrapper is silent until rustc exits. For most crates this is fine (~1–10 s). For long crates (Firefox's `gkrust` at ~8 min, large LTO units, heavy proc-macro-driven compiles) the wrapper looks frozen — to the user, to mach's TTY-throttling logic, and to any tooling parsing stderr. The bench harness surfaced this concretely: 7 minutes of dead output while a single rustc ground through a staticlib.

kache already has the data to make this much better — `compile_time_s` per `(crate, key)` is recorded in every event log, the daemon already tracks spawned children, and the TUI (`kache monitor`) already exists. The gap is plumbing.

This issue groups a cluster of related improvements around the same theme: "what is kache doing *right now*."

### Headline: heartbeat + ETA (P1)

While kache is waiting on a spawned rustc, the daemon emits periodic progress events:

```
[kache] spawning rustc for gkrust (typical: 7m51s)
[kache] still compiling gkrust — 4m20s elapsed (typical: 7m51s, ETA 3m31s)
[kache] still compiling gkrust — 4m50s elapsed (typical: 7m51s, ETA 3m01s)
```

- ETA is derived from the median `compile_time_s` for prior recorded events with the same crate name.
- First-time crates omit the typical/ETA suffix.
- Default cadence: every 30 s, configurable.
- Emitted on stderr (subject to KACHE_LOG / KACHE_LOG_FILE plumbing — same gating as other warns).

### Stuck-process detection (P2)

If a spawned child rustc shows CPU usage below 1% for 60 s (sampled cheaply via `getrusage` / `/proc/<pid>/stat`), log a WARN naming the child PID and crate:

```
WARN rustc for gkrust appears stuck (no CPU for 60s, PID 3125)
```

Catches the degenerate cases — swap thrashing, deadlocks, lock contention on the cache store — that otherwise present as "kache is slow." Cheap to implement; high signal-to-noise.

### Why-am-I-missing diagnostics (P2)

On a miss for a crate that previously had a hit in this workspace, diff the keys field-by-field and emit which composition input changed:

```
miss: crate gkrust (last hit 2h ago; key changed in: codegen:linker, env_dep:OUT_DIR)
```

[...]

PR #703, opened 2026-08-08
fix(stats): render the daemon's effective config and surface CLI/daemon config mismatch (#689)

### LINKED ISSUE #689: stats: daemon-backed responses silently ignore the invoking CLI's KACHE_CONFIG/XDG_CONFIG_HOME — and auto-start flips the semantics when no daemon is running
## Summary

`kache stats` (and every daemon-first CLI read) reports the **daemon's** config, silently discarding the invoking process's `KACHE_CONFIG` / `XDG_CONFIG_HOME` / `KACHE_*` env. There is no mismatch warning. Worse, when no daemon is reachable the CLI **auto-starts one that inherits the CLI's env**, so the exact same command line produces opposite results depending on whether a daemon happened to be alive.

Observed on kache 0.13.0, macOS (apfs), local-only setup.

## Reproduction

```sh
# 1) a daemon is running with the default config (cap 50 GiB)
kache stats            # "Store: ... / 50.0 GiB"

# 2) point the CLI at a config with a different cap
printf '[cache]\nlocal_max_size = "117GiB"\n' > /tmp/kache-repro.toml
KACHE_CONFIG=/tmp/kache-repro.toml kache stats
# -> "Store: ... / 50.0 GiB"   (daemon's config; KACHE_CONFIG silently ignored)

# 3) same command, but force the daemon out of the loop with a dead socket
KACHE_SOCKET_PATH=/tmp/kache-nowhere/x.sock KACHE_CONFIG=/tmp/kache-repro.toml kache stats
# -> "Store: ... / 117.0 GiB"  (auto-started daemon inherited the CLI env)
```

Step 3 also leaves a second `kache daemon run` process behind (bound to the dead-socket path), which then answers *later* probes with **its** inherited config — a sequence of debugging probes poisons itself. `kache doctor` on this machine reported `3 daemon processes running (pid …), expected 1` after exactly such a session.

## Why this bites

The silence converts operator config mistakes into a false systemic diagnosis. Real case: an operator had `[cache] size = "120GiB"` (wrong key; correct is `local_max_size` — silently ignored by the deliberate non-`deny_unknown_fields` parse, the tradeoff documented above `RemoteFileConfig` in `src/config.rs`). Every verification attempt went through `kache stats`, whose numbers came from a long-lived daemon — so even after fixing the key **via `KACHE_CONFIG` the numbers did not move**, and the conclusion became "kache ignores config files entirely; only env vars work". It took bypassing the daemon to see that config loading is fine.

## Where in the source

- `src/cli.rs` — `fetch_stats_snapshot`: "Try daemon first, fall back to direct reads"; `max_size: resp.max_size` comes from the daemon response, and the fallback arm auto-starts a daemon (`daemon::start_daemon_background()`) that inherits the CLI env.
- `src/daemon.rs` — the config watch (`config_file_fingerprint`, "This watches only the file the daemon itself resolved") correctly reloads on *file edits*, but by design never sees a per-invocation `KACHE_CONFIG`/`XDG_CONFIG_HOME` that resolves to a *different path* than the daemon's.

## Expected

Match the project's own "never silent" principle (`ignore_env` warns which overrides it suppresses; `why-miss` "says how to turn them on rather than staying silent"): when a daemon-backed command's own resolved config fingerprint (`config_file_fingerprint()`) differs from the daemon's, print a one-line warning naming both paths, e.g.

```
warning: daemon is using ~/.config/kache/config.toml; this invocation resolved /tmp/kache-repro.toml (KACHE_CONFIG) — daemon-side values shown. `kache daemon restart` to reload.
```

That requires the daemon to report its config fingerprint/path in the stats response. Alternatively (or additionally): resolve display-only values like `max_size` CLI-side instead of trusting the daemon copy.

## Actual

No warning anywhere; daemon values presented as if they were the invocation's config; auto-start makes the behavior dependent on daemon liveness.

valorkin:
## Problem

`kache stats` mixes two sources of truth without saying so: the numbers come from the daemon, but config-derived lines (store cap, prefetch policy) are re-resolved from the invoking process's own `KACHE_CONFIG` / `XDG_CONFIG_HOME` / `KACHE_*` env. A `KACHE_CONFIG` pointing at a different file is silently ignored — and when no daemon is reachable, the CLI silently auto-starts one that inherits the CLI's env, so the exact same command produces opposite results depending on daemon liveness. A debugging session of such probes leaves extra daemons behind (`kache doctor`: "3 daemon processes running"), and an operator's config mistake escalates into "kache ignores config files entirely" (#689's real-world case).

## Fix design

Reporting only — no change to daemon config loading, restart, or auto-start behavior:

1. **The daemon reports what it loaded.** `StatsResponse` gains an `effective_config` block (`max_size`, `cache_dir`, resolved config-file path, `prefetch_enabled`, socket path, startup timestamp), captured once at `Daemon::new`. Startup capture cannot go stale: env/cwd are fixed for the process's lifetime and the existing config-file fingerprint watcher restarts the daemon on content changes.
2. **Daemon-backed lines render the daemon's view.** The `Prefetch: disabled` line — previously computed from the *client's* config (the #652 stats item) — now renders the daemon's effective policy, and the `Daemon:` line names the config file the daemon loaded. The store cap already came from the daemon; now its source is visible.
3. **Divergence warns, one line per rendered field**, naming both values and both sources (`config_mismatch_warnings` is pure, so every branch is unit-testable without a daemon). Matching values with differing paths stays quiet on purpose — the rendered output is correct then, and project-file resolution from different cwds would otherwise warn constantly.
4. **Auto-start announces itself** on stderr, so the liveness-dependent semantics flip becomes visible. The TUI passes `announce_auto_start: false` — its raw-mode alternate screen owns the terminal.
5. **Version-skew tolerant both ways**, following the `in_flight`/`prefetch` precedent: the new field is `#[serde(default)]`; an old daemon yields `None` and the CLI falls back to its own config with an explicit label (`[client config — daemon did not report its policy]`) instead of presenting a guess as a daemon fact. An old client ignores the unknown field. `kache daemon status` also prints the daemon-loaded config path when reported.

## Measured before/after

macOS (APFS, M4 Max), release builds, isolated store/socket/config (`KACHE_CACHE_DIR=/tmp/kache-store-wd-*`, own socket paths, `KACHE_S3_*` unset). Baseline = pristine `upstream/main` @ 4961602. Daemon started with `local_max_size = "50GiB"`; probe uses `KACHE_CONFIG=/tmp/kache-wd-repro.toml` (117 GiB) — the exact repro from #689.

**Before — silent wrong attribution (daemon alive):**
```
$ KACHE_CONFIG=/tmp/kache-wd-repro.toml kache stats
Store:      0 B / 50.0 GiB (0 entries, 0%)          # 117 GiB config silently ignored, no warning
Daemon:     v0.13.0 (epoch 1786229374)
```

**After — same command:**
```
warning: daemon (started 2026-08-08 23:01 UTC, config /tmp/kache-wd-daemon.toml) has local_max_size=50.0 GiB; this process's config (/tmp/kache-wd-repro.toml) says 117.0 GiB — the daemon's value is in effect; `kache daemon restart` to apply yours
Store:      0 B / 50.0 GiB (0 entries, 0%)
Daemon:     v0.13.0 (epoch 1786230081, config /tmp/kache-wd-daemon.toml)
```

**Before — dead socket flips the answer silently:**
```
$ KACHE_SOCKET_PATH=/tmp/kache-wd-nowhere/x.sock KACHE_CONFIG=/tmp/kache-wd-repro.toml kache stats
Store:      0 B / 117.0 GiB (0 entries, 0%)         # auto-started daemon inherited CLI env, nothing said
```
(and `ps` then shows 2 leftover `kache daemon run` processes — the #689 doctor complaint)

**After — same command:**
```
kache: no daemon reachable at /tmp/kache-wd-nowhere-after/x.sock; starting one inheriting this process's environment
Store:      0 B / 117.0 GiB (0 entries, 0%)
Daemon:     v0.13.0 (epoch 1786230081, config /tmp/kache-wd-repro.toml)
```
The flip still happens (auto-start behavior unchanged, out of scope) but is announced and correctly attributed.

**Prefetch policy (the #652 stats item), after:** `KACHE_PREFETCH_ENABLED=0 kache stats` against an enabled daemon no longer renders a wrong `Prefetch: disabled` line; instead:
```
warning: daemon (started …, config …) has prefetch_enabled=true; this process's config (…) says false — the daemon's value is in effect; `kache daemon restart` to apply yours
```

## Tests

- `handle_stats` carries the effective config (values, non-empty config path, startup stamp).
- Wire compat: a `StatsResponse` without `effective_config` parses to `None` (extends the existing #131 old-daemon default test).
- `render_stats`: daemon-reported policy wins in both directions; old-daemon fallback is labeled; daemon config path shown on the `Daemon:` line.
- `config_mismatch_warnings`: silent on agreement (even with differing paths), one warning per differing field naming both sources, `started_at_ms = 0` renders "unknown time" instead of 1970.
- Integration (real binary, real auto-started daemon): auto-start notice on stderr; cross-config `kache stats` renders the daemon's cap and warns naming both files; `KACHE_PREFETCH_ENABLED=0` against an enabled daemon warns; both tests stop their daemon.

Gates on this branch (macOS, toolchain-pinned 1.95.0):
- `cargo fmt --all -- --check` — clean.
- `cargo clippy --workspace --all-targets -- -D warnings` — **zero findings introduced**: my local 1.95.0 clippy flags 3 pre-existing lints (`src/remote_backend.rs:191`, `src/tui.rs`, `src/wrapper.rs:3627`) on pristine `upstream/main` too (your CI is green there, so this looks like local-vs-CI clippy skew); the before/after error sets are byte-identical apart from a 2-line shift in `tui.rs` from a doc comment.
- `cargo test --workspace --no-fail-fast` — **1911 passed / 2 failed** across 26 suites: unit binary 1647/2, all other suites green (90, 54, 32, 27, 14, 12, 10, 7, 4, 3, 3, 2, 2, 1, 1, 1, 1 passed; 8 empty). The 2 failures are `incremental_policy::tests` lock/reset tests — a module this PR does not touch; the failing names rotate between runs, and the module passes 26/26 when run alone, so this looks like a load-sensitive local flake, not a regression (happy to dig if it's not known).
- E2E gate harness (`kache-scenario --select suite:e2e --select tier:gate`, as CI's macOS smoke runs): **40 fixtures pass, 2 skip** (`e2e-cc-bench-flags-gnu`, `e2e-cc-cl-debug` — not applicable on this host), `PASS: all executed fixtures green`.

## Scope cuts / follow-ups (deliberate)

- The `Remote:` line and the key-LIST refresh cadence still render client config: extending `EffectiveConfig` with a remote descriptor is more wire surface than the issue needs, and the fields the issue names (cap, store dir, config path, prefetch policy) are covered. Happy to follow up if you want the remote block carried too.
- Auto-start/daemon lifecycle behavior is intentionally untouched (announcement only), per the issue's reporting-only framing.
- `kache doctor` could reuse `effective_config` for a config-divergence check; left out to keep this PR one change.

Fork CI note: workflow runs need maintainer approval, and the Windows jobs run on your self-hosted runners.

Fixes #689. Addresses the stats-reporting item of #652 ("Carry the effective policy in the stats snapshot and render that instead of client config").

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -42,6 +42,10 @@ pub(crate) struct StatsSnapshot {
     /// In-flight miss compiles (kunobi-ninja/kache#131); empty when the
     /// daemon is unreachable (the TUI then falls back to tailed heartbeats).
     pub in_flight: Vec<daemon::InFlightEntry>,
+    /// The daemon's effective config from the stats response
+    /// (kunobi-ninja/kache#689); `None` when the daemon is unreachable or
+    /// predates effective-config reporting.
+    pub daemon_effective_config: Option<daemon::EffectiveConfig>,
 }
 
 impl Default for StatsSnapshot {
@@ -85,6 +89,7 @@ impl Default for StatsSnapshot {
             blob_stats: None,
             prefetch: daemon::PrefetchStatsSnapshot::default(),
             in_flight: Vec::new(),
+            daemon_effective_config: None,
         }
     }
 }
@@ -108,11 +113,19 @@ pub fn compile_weighted_hit_rate(es: &daemon::EventStatsResponse) -> Option<f64>
 }
 
 /// Try daemon first, fall back to direct reads.
+///
+/// With `announce_auto_start` set, the daemon-unreachable path says on stderr
+/// that it is starting a daemon inheriting this process's environment
+/// (kunobi-ninja/kache#689) — otherwise the same command silently flips from
+/// "daemon's config" to "my config" depending on daemon liveness. The CLI
+/// passes true; the TUI passes false because its raw-mode alternate screen
+/// owns the terminal.
 pub(crate) fn fetch_stats_snapshot(
     config: &Config,
     include_entries: bool,
     sort_by: &str,
     hours: Option<u64>,
+    announce_auto_start: bool,
 ) -> StatsSnapshot {
     let event_hours = hours.or(Some(24));
     let blob_stats = || {
@@ -149,11 +162,18 @@ pub(crate) fn fetch_stats_snapshot(
             blob_stats: blob_stats(),
             prefetch: resp.prefetch,
             in_flight: resp.in_flight,
+            daemon_effective_config: resp.effective_config,
         };
     }
 
     // Daemon unreachable or stale socket: best-effort auto-start for monitor/stats UX.
     // This path is not used by compile-time hot operations.
+    if announce_auto_start {
+        eprintln!(
+            "kache: no daemon reachable at {}; starting one inheriting this process's environment",
+            config.socket_path().display()
+        );
+    }
     if daemon::start_daemon_background().unwrap_or(false)
         && let Ok(resp) =
             daemon::send_stats_request(config, include_entries, Some(sort_by), event_hours)
@@ -182,6 +202,7 @@ pub(crate) fn fetch_stats_snapshot(
             blob_stats: blob_stats(),
             prefetch: resp.prefetch,
             in_flight: resp.in_flight,
+            daemon_effective_config: resp.effective_config,
         };
     }
 
@@ -274,6 +295,7 @@ pub(crate) fn snapshot_from_direct_reads(
         blob_stats: store.as_ref().and_then(|s| s.blob_stats().ok()),
         prefetch: daemon::PrefetchStatsSnapshot::default(),
         in_flight: Vec::new(),
+        daemon_effective_config: None,
     }
 }
 
@@ -282,9 +304,20 @@ pub(crate) fn snapshot_from_direct_reads(
 /// Print a one-shot stats summary to stdout.
 pub fn stats(config: &Config, hours: Option<u64>) -> Result<()> {
     let hours = hours.unwrap_or(24);
-    let snap = fetch_stats_snapshot(config, false, "size", Some(hours));
+    let snap = fetch_stats_snapshot(config, false, "size", Some(hours), true);
     let store = Store::open(config)?;
     let blob_stats = store.blob_stats()?;
+
+    // #689: the numbers below are daemon-first, so before rendering them say
+    // when the daemon's effective config disagrees with this invocation's —
+    // otherwise a config edit that has not reached the daemon masquerades as
+    // "kache ignores config files".
+    if let Some(eff) = &snap.daemon_effective_config {
+        for warning in config_mismatch_warnings(config, &crate::config::resolve_config_path(), eff)
+        {
+            eprintln!("{warning}");
+        }
+    }
 
     for line in render_stats(&snap, &blob_stats, config, hours) {
         println!("{line}");
@@ -397,8 +430,16 @@ pub(crate) fn render_stats(
         } else {
             ""
         };
+        // Name the config file the daemon loaded (#689): the store cap and
+        // policy lines describe THAT file, which need not be the one this
+        // invocation resolved.
+        let config_note = snap
+            .daemon_effective_config
+            .as_ref()
+            .map(|eff| format!(", config {}", eff.config_path))
+            .unwrap_or_default();
         lines.push(format!(
-            "Daemon:     v{} (epoch {}){mismatch}",
+            "Daemon:     v{} (epoch {}{config_note}){mismatch}",
             snap.daemon_version, snap.daemon_build_epoch,
         ));
     } else {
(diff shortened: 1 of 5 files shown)


jleni (changes requested):
Thanks—the wire additions are compatible, but three correctness gaps need fixing before delivery:

- Different explicit `KACHE_CONFIG` paths must warn even when selected values happen to match. Otherwise #689's invalid `[cache] size=...` case remains silent; the current test pins that wrong behavior.
- Stats must not combine daemon `prefetch_enabled` with the invoking client's remote presence/cadence. Carry daemon-owned remote state or suppress those mixed-source statements when configs differ.
- `kache daemon restart` does not make a managed launchd/systemd daemon inherit the invoking shell's `KACHE_CONFIG`; the guidance can perpetuate the mismatch.

Please add path-mismatch/identical-values and both remote-mismatch direction tests. I will handle the mechanical rebase and maintainer delivery after #701; the stale fork run will not be approved.

jleni:
Maintainer adoption is now tracked in https://github.com/kunobi-ninja/kache/pull/715. It preserves Dmitriy Shekhovtsov's contributor authorship and carries the reviewed fixes on a repository-owned branch. This original PR remains open while the replacement completes hosted CI and review.

jleni:
Thank you, Dmitriy. The maintained replacement #715 has now passed the full hosted CI matrix and landed on `main` as signed commit `f73bc1cbb851493376d7a393560630ee7e9821db`, with your authorship credited in the squash commit. Closing this superseded source PR.

Issue #711, opened 2026-08-09
No put-side admission control, and max-age eviction never runs automatically


valorkin:
## Summary

There is no admission control on `put`, and the age-based eviction policy exists but is never reached by any automatic path. On a store whose working set turns over daily, this fills the cache with entries that are cheaper to rebuild than to keep.

## Evidence (v0.13.0 @ a7a041c)

- `OlderThanPolicy` is implemented and `kache gc --max-age` is wired through the CLI, but the daemon's periodic sweep passes `None` for the age policy, which routes to plain `evict()` (size pressure only) — so max-age never runs unless a human types it.
- The only compile-cost gate in the codebase is `KACHE_MIN_COMPILE_MS` (`src/daemon.rs`), applied as `.filter(|e| e.compile_time_ms >= min_compile_ms)`. It filters **remote prefetch candidates only** — never `put`.

So every compilation result is stored regardless of how cheap it was to produce, and nothing ever ages it back out automatically.

## Measured on a real 109 GiB store (24,855 entries)

| | entries | size |
|---|---|---|
| rebuild in < 1s | **13,334 (54%)** | 9.4 GiB |
| `hit_count = 0` and idle > 6h | **13,859** | **56.5 GiB** |
| live working set (accessed < 6h) | 4,860 | **16.2 GiB** |

A store holding 109 GiB to serve a ~16 GiB working set is paying for entries that will never be read again, and evicting the wrong ones when it does run (see the two duplicate/hysteresis issues filed alongside this one).

## Live reproduction

**Admission control** — a real, near-instant crate compile (measured 70 ms) stored under the default (no threshold configured):

```
$ KACHE_MIN_STORE_COMPILE_MS=200 cargo build   # (compile measured ~70ms)
$ kache list
No cached entries.
```

vs. the same build with no threshold set: the entry is stored as before (0 = off, preserves existing behavior exactly).

**Automatic max-age** — a real entry, `last_accessed` aged well past any reasonable retention window, in a store nowhere near its size budget (5 GiB cap, ~14 KB used):

```
$ kache daemon run     # periodic-sweep-equivalent: runs immediately on startup
$ kache list
Crate  ...  1 entries      # unchanged — no automatic path ever reaches OlderThanPolicy
$ kache gc --max-age 24h   # only removed once a human types the exact flag
Evicted 1 entries older than 24h.
```

## Suggested fixes

[...]

Issue #710, opened 2026-08-09
GC: evict() has no hysteresis — trigger and target are both 90%, so it thrashes at the boundary


valorkin:
## Summary

`Store::evict()` documents a hysteresis band but does not implement one: the trigger threshold and the stop target are the same number, so GC fires the moment the store crosses 90% of `max_size` and stops the instant it drops back under that same line.

## Evidence (v0.13.0 @ a7a041c)

`src/store.rs`, `Store::evict()`:

```rust
let target = self.config.max_size * 9 / 10; // evict to 90% — avoids boundary thrashing
// ...
let size_before = self.physical_size()?;
if size_before <= target {
    return Ok(GcStats::default());
}
```

The comment claims the 90% figure exists to avoid boundary thrashing, but the same value is used as the entry condition, so there is no band at all.

A 20% slack does exist — but only in the wrapper's decision to *spawn* a background GC (`AUTO_GC_SLACK_PERCENT`, `src/wrapper.rs`, comment: "evict() already targets 90% of max_size; triggering only above 110% keeps the two thresholds apart"). That comment's premise — that `evict()`'s trigger sits meaningfully below its own spawn threshold — was already true for that one caller, but every caller that invokes `evict()`/`run_gc()` without going through the wrapper's spawn decision (`kache gc`, the daemon's periodic sweep, a bare RPC) lands directly on the 90%-equals-90% knife edge.

## Live reproduction

Five real-crate units (unique content each, no dedup interaction) built into an isolated store with `local_max_size` set so the physical total sits between the old 90% line and the actual cap:

- `max_size` = 240 KiB (245,760 B); old target/trigger (90%) = 221,184 B
- Physical after 4 units: 226,560 B — **above the 90% line, below the cap**

```
$ kache gc          # (daemon startup GC, immediate-on-launch)
```
```json
{"entries_evicted": 1, "bytes_freed": 56640, "blobs_removed": 3}
```

Store dropped to 169,920 B even though it never came close to the configured 240 KiB cap. Adding exactly one more unit brought physical back to 226,560 B; a second `kache gc` call fired again, evicting another 56,640 B down to 169,920 B — the same cycle, indefinitely, with the store never actually using its configured budget.

## Suggested fix

[...]

Issue #608, opened 2026-07-25
fix(gc): size-based eviction measures logical entry bytes, not physical blob bytes


jleni:
## What

`Store::total_size()` returns `SELECT COALESCE(SUM(size), 0) FROM entries` (`src/store.rs:1727`). That is a *logical* sum. When several entries share content-addressed blobs, those bytes are counted once per referencing entry while the disk holds a single copy.

Size-pressure eviction is built on that number end to end:

- `evict()` computes `target = self.config.max_size * 9 / 10` and compares it against `total_size()` (`src/store.rs:1935`).
- `SizePressurePolicy` ranks by `EntryFeatures::size` (`src/eviction.rs:36`), the same per-entry logical size, via `size_pressure_score` (`src/eviction.rs:71`).
- The removal walk subtracts each evicted entry's logical size to decide when it has reached the target (see the comment above the `evict_with` call in `evict()`).

Blob removal is refcount-gated in `remove_entry_guarded`: a blob is unlinked only when the last entry referencing it goes away.

So the budget, the ranking, and the stop condition all speak logical bytes, while the thing being managed is physical bytes.

## Why this is wrong

The two numbers diverge, and the error runs in both directions.

**Over-eviction.** A store whose entries share blobs reports a logical size well above its real footprint. `evict()` then fires while the disk is comfortable and destroys rebuild value without reclaiming meaningful space. Cross-clone and multi-worktree stores are precisely where blob sharing is highest, so this misfires hardest in the workload kache is aimed at.

**Under-freeing.** Evicting an entry whose blobs are all shared frees zero bytes on disk, but the walk subtracts that entry's full logical size and can stop believing it reached the target. The store stays over its actual budget.

**Ranking.** Within a sweep the policy prefers larger entries, but "larger" is logical. A 500 MB entry whose blobs are all shared frees nothing; a 200 MB entry with unique blobs frees 200 MB. The current score picks the first one.

## The physical number already exists

`SELECT COALESCE(SUM(size), 0) FROM blobs` is already queried at `src/store.rs:2618`, and it is what the monitor's `Blobs: N physical` figure reports. The store already knows both numbers. Eviction reads the wrong one.

## Suggested shape

Two separable changes:

1. Make the eviction trigger and stop condition use physical bytes (`SUM(blobs.size)`) rather than `SUM(entries.size)`. This changes when GC fires, so `max_size` needs an explicit documented meaning as part of the change.
[...]

PR #712, opened 2026-08-09
fix(gc): real hysteresis band, marginal-bytes-gated duplicate eviction, admission control + auto max-age

### LINKED ISSUE #709: GC: duplicate eviction runs unconditionally, destroys the highest-hit keys, frees zero bytes
## Summary

`Store::evict_duplicate_entries()` runs on every GC pass, ahead of any size check, with no byte budget — and on a dedup-backed store it deletes the highest-hit keys while freeing zero bytes.

## Evidence (v0.13.0 @ a7a041c)

- `src/daemon.rs` — `let dedup_stats = store.evict_duplicate_entries().unwrap_or_default();` is called unconditionally in the GC path, ahead of any size-pressure test.
- `src/store.rs` — the call passes `stop_at = None`, so there is no byte budget to stop at.
- `DuplicatePolicy` groups committed entries by `content_hash` and evicts every member except the most-recently-accessed.

Because blobs are refcounted and shared (`materialize_blob`, content-addressed storage), evicting a duplicate **key** whose blob is still referenced by the surviving twin reclaims nothing. This isn't a corner case — it's structural: `DuplicatePolicy` always keeps exactly one survivor per content group, and that survivor alone keeps the shared blob(s) alive. Any duplicate group of size ≥2 therefore has zero *marginal* reclaimable bytes for every non-survivor member, by construction.

## Measured on a real 109 GiB store (24,855 entries)

| policy | entries evicted | accumulated hits destroyed | GiB actually freed |
|---|---|---|---|
| **duplicate** | **2,359** | **10,284** | **0** (blobs shared) |
| size-pressure | 189 | 23 | 43.34 |

The store's lifetime hit total was 33,439 — duplicate eviction destroyed keys carrying ~31% of it.

`gc_stats.json` from one recorded pass: `entries_evicted: 142, bytes_freed: 0, blobs_removed: 0`.

## Live reproduction

Two separate checkouts building byte-identical output from real crate sources under different `KACHE_KEY_SALT` values (simulating two worktrees converging on the same content under different cache keys — a legitimate, common scenario, since path normalization succeeding is exactly what produces this). Isolated store, `local_max_size` set below the store's physical size so GC fires:

```
$ kache gc
Evicted 2 entries to stay under size limit.
Store: 70.2 KiB (2 entries)
```
```json
{"entries_evicted": 2, "bytes_freed": 0, "blobs_removed": 0}
```

Physical store size was **unchanged** (71,861 bytes before and after) — the two evicted entries' blobs were still held by their surviving twins. Both deleted keys now miss on their next build, for zero reclaimed space.

## Why it matters more than it looks

Two worktrees converging on byte-identical output under different cache keys is path normalization **succeeding**. GC then deletes all but one of those keys, so the next build in the other worktree misses and recompiles — for zero reclaimed space.

It's also not disableable: `KACHE_AUTO_GC=0` only suppresses the wrapper-spawned pass; the daemon's periodic sweep calls `run_gc(None)` unconditionally, which reaches `evict_duplicate_entries()` regardless.

## Suggested fix

1. Gate `evict_duplicate_entries()` behind real size pressure — only scan and run it when the store is actually over its eviction trigger.
2. Skip any candidate whose removal would free no marginal bytes — i.e. whose blob's reference count would remain > 0 after removing this entry. `#608`'s `entry_blobs`/`reclaimable_bytes` tracking already computes this signal for the size-pressure policy; `DuplicatePolicy` just needs to consult it too.

Size: small — one gate plus a marginal-bytes filter in `DuplicatePolicy::select`.

valorkin:
## Summary

Three related GC/store defects, found by analyzing a real 109 GiB production store (24,855 entries) and fixed with live verification against real crate builds.

- `evict_duplicate_entries()` ran unconditionally on every GC pass, ahead of any size check, with no byte budget. Because blobs are content-addressed and refcounted, evicting a duplicate key whose blob is still held by its surviving twin always frees zero bytes while destroying that key's hit history.
- `evict()`'s trigger and target were both `max_size * 9/10`, so the doc comment claiming a 90% headroom band was fictional — eviction fired the instant the store crossed 90% and stopped the instant it dropped back under that same line.
- No put-side admission control existed, and the periodic sweep never reached `OlderThanPolicy` even though it and `kache gc --max-age` already existed, so nothing aged out automatically.

Fixes #709, #710, #711.

### #709 — duplicate eviction destroys keys for zero freed bytes

`DuplicatePolicy` groups entries by `content_hash` and evicts every member except the most-recently-accessed. On a refcounted blob store this is structural, not a corner case: the survivor alone keeps the shared blob(s) alive, so every non-survivor member of a duplicate group has zero *marginal* reclaimable bytes, always.

**Fix:** `evict_duplicate_entries()` is now gated behind the same size-pressure trigger `evict()` uses (`over_eviction_trigger`), and `DuplicatePolicy::select` skips any candidate whose backfilled `reclaimable_bytes` is a known `0`. `None` (not yet backfilled into `entry_blobs`) falls back to logical size, matching how the size-pressure policy already treats that same unknown — GC backfills those rows at the top of every sweep, so it's a one-pass transient, not a standing exemption.

On a fully backfilled, refcounted store this makes duplicate eviction select almost nothing. That's the point: those bytes were never reclaimable. Size-pressure eviction — which ranks on the same marginal-bytes signal (#608) — is what actually frees space.

### #710 — no hysteresis band

Fix lives centrally in `Store::evict()`: trigger fires at `max_size` (100%), target stays at `max_size * 9/10` (90%). Every caller (`kache gc`, the daemon's periodic sweep, the post-upload check) inherits the same real 10% band from one change, rather than each call site needing its own fix.

### #711 — no admission control, max-age never runs automatically

Two knobs, deliberately separate from each other and from `KACHE_MIN_COMPILE_MS`:

- `KACHE_MIN_STORE_COMPILE_MS` (default `0`, off) — put-side admission control. Gates whether an entry that **doesn't exist yet** is worth local disk. `KACHE_MIN_COMPILE_MS` gates whether an entry that **already exists** remotely is worth downloading — different cost model, same signal, must not share a knob.
- `KACHE_GC_MAX_AGE_HOURS` (default `168` = 7 days) — every unattended GC path (the daemon's periodic sweep, the wrapper's background auto-GC worker, a bare `kache gc`) now composes an age-retention pass with size-pressure eviction. An explicit `kache gc --max-age <duration>` is unaffected — it still means exactly what was typed, running only that policy, never composed with anything else.

## Live verification (TAP-derived reproduction)

Built real crates from a large downstream monorepo (verbatim source, zero external dependencies: a 23-line and a 60-line crate) through both an unpatched `a7a041c` binary and this patch, in fully isolated `KACHE_CACHE_DIR`/`KACHE_SOCKET_PATH`/`.kache.toml` environments — no contact with any default store or socket.

**#709 — duplicate eviction, before/after** (two simulated worktrees, same content under different `KACHE_KEY_SALT`-derived keys, physical store 71,861 B, `local_max_size` set below it so GC fires):

| | entries evicted | bytes freed | entries surviving |
|---|---|---|---|
| unpatched | 2 | **0** | 2 (physical size unchanged: 71,861 → 71,861) |
| patched | **0** | 0 | 4 (all survive; physical size unchanged: 71,861 → 71,861) |

**#710 — hysteresis, before/after** (5 real-crate units, `max_size` = 240 KiB / 245,760 B; old 90% line = 221,184 B):

| | physical | pass 1 | physical after | added 1 unit → | pass 2 |
|---|---|---|---|---|---|
| unpatched | 226,560 B (> old 90% line) | evicted 1, freed 56,640 B | 169,920 B | 226,560 B | evicted 1 **again**, freed 56,640 B |
| patched | 226,560 B (< new 100% trigger) | evicted **0** | 226,560 B (unchanged) | 283,200 B (> trigger) | evicted 2, freed 113,280 B, down to 169,920 B (90% target) |

Unpatched fires on essentially every marginal crossing of the old 90% line; patched gives the store genuine use of its configured budget before acting.

**#711 — admission control and auto max-age, live:**

- A real crate compile measured at 70 ms, with `KACHE_MIN_STORE_COMPILE_MS=200`: `kache list` → `No cached entries.` (not stored). Same build with no threshold set: stored as before (default `0` preserves existing behavior exactly).
- A real entry aged well past any reasonable retention window, store nowhere near its size budget: unpatched daemon startup (periodic-sweep-equivalent) leaves it untouched indefinitely — only an explicit `kache gc --max-age 24h` removes it. Patched daemon startup, with `gc_max_age_hours = 24` configured, evicts it automatically (`entries_evicted: 1, bytes_freed: 13845`) with no `--max-age` ever typed.

Real-store analysis (Section B of the original write-up) was skipped: available disk (120 GiB) was not comfortably above the store's own size (127 GiB), and copying it — even via APFS clonefile — was judged not worth the risk on an active development machine. The live TAP-derived reproduction above is the primary evidence.

## Test coverage

New/updated regression tests:
- `eviction.rs`: `duplicate_skips_victim_whose_blob_is_still_referenced`, `duplicate_still_evicts_victim_with_positive_marginal_bytes`
- `store.rs`: `evict_duplicate_entries_spares_a_pair_sharing_one_blob`, `evict_duplicate_entries_skips_the_scan_under_budget`, `evict_noop_within_the_hysteresis_band`, `evict_fires_at_the_trigger_and_stops_at_the_target` (replacing the now-incorrect `evict_fires_and_stops_at_the_ninety_percent_target`)
- `wrapper.rs`: `store_admits_compile_gates_on_configured_threshold`
- `daemon.rs`: `run_gc_none_applies_configured_max_age_even_under_size_budget`, `run_gc_none_skips_age_eviction_when_max_age_hours_is_zero`
- Three existing dedup tests updated to reflect the corrected behavior (a genuine content-duplicate pair now survives) rather than the old destructive one.

`cargo test --workspace --all-features`: **1920 passed, 0 failed.**
`cargo fmt --check`: clean.
`cargo clippy --workspace --all-targets --all-features -- -D warnings`: same 3 pre-existing warnings as a pristine `upstream/main` checkout (`remote_backend.rs:191`, `tui.rs:516`, a `wrapper.rs` collapsible-if), none introduced by this change — verified by diffing clippy's output against both trees.

## Caveats and scope

- The real-store analysis (Section B) was skipped for the disk-safety reason above; all evidence here is from the isolated live reproduction plus the unit/integration suite.
- `evict_duplicate_entries()` now selects almost nothing on a normally-operating store, by design — see the #709 discussion above. If a future storage layout stops fully sharing blobs across content-identical entries, this becomes live again; it isn't dead code, just usually inert today.
- Docs updated (`docs/getting-started/configuration.mdx`, `docs/commands/reference.mdx`) to describe the real band and the two new knobs; the stale "GC evicts down to 90%" framing is gone.

## Manual Testing

Not applicable beyond the automated suite above — this is a backend GC/store change with no UI surface. The live TAP-derived reproduction functions as the manual/live verification this change needed.

---

Fork CI needs maintainer approval to run. Windows CI is understood to be the maintainer's stated responsibility for fork PRs.


jleni (changes requested):
This should be split before CI; several current behaviors are unsafe:

- #709 still fails open: `reclaimable_bytes=None` is treated as logical size, so bounded backfill can delete legacy duplicates for zero reclaimed bytes. Select only `Some(bytes > 0)`.
- A default 168-hour expiry silently changes every existing installation. Keep the default off unless maintainers explicitly choose a breaking retention policy.
- Size eviction runs before age eviction, causing avoidable over-eviction; age should run first and pressure must be recomputed.
- Manual, periodic, and background GC modes are conflated, so output and `--max-age` semantics become inaccurate.
- Admission also suppresses remote publication although it is presented as a local-disk policy.

Recommended landing order: #710 hysteresis; #709 with fail-closed legacy coverage; #711 admission; then #711 automatic age policy after mode/default/order decisions. Please also fix the `#1/#2/#4` references. Do not rebase or approve the stale run yet; split/design comes first.

jleni:
Thanks for the contribution. I have adopted the pinned source head `9db8a715c34c8af08f22868dcdfabe284e24dff4` into maintainer-owned replacement #718, preserving your authorship while separating the four policy changes and applying the safety fixes found during review. The original PR will remain open until the replacement completes hosted CI and lands.

jleni:
Thank you, Dmitriy. We adopted this contribution in maintainer-owned PR #718, preserved your authorship and exact source pin, added the safety and compatibility corrections found during review, and merged it after full GitHub-hosted CI passed. Closing this source PR in favor of the merged replacement: https://github.com/kunobi-ninja/kache/pull/718

PR #739, opened 2026-08-12
Fix kache doctor probe gcc panic on non-English locales (e.g. LANG=zh_CN.UTF-8)

AzureZee:
## Summary

When in a non-English language environment, running `kache doctor` will panic.
The issue lies in `filter(|_| base[base.len() - 4..].eq_ignore_ascii_case(".exe"))` inside the `extract_gnu_cc1_line()` function in `src/probe/resolve.rs`.
It can be fixed by changing it to `filter(|&n| base.is_char_boundary(n) && base[n..].eq_ignore_ascii_case(".exe"))`.

Before fix:
```
❯ kache doctor

thread 'main' (651368) panicked at src/probe/resolve.rs:245:29:
start byte index 8 is not a char boundary; it is inside '内' (bytes 6..9) of `使用内建`
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After fix:
```
❯ ./target/release/kache doctor

  kache v0.14.0    rustc 1.97.1 (8bab26f4f 2026-07-14)
#...skip something 
  ✓ Compiler probe  cc -### resolves (cc (GCC) 16.1.1 20260515 (Red Hat 16.1.1-2))

  All checks passed.
  Daemon checks are informational: no remote cache or planner configured (the daemon is optional for local-only use).
```

## Test plan

cargo test --bin kache extract_gnu_cc1_line_handles_non_ascii_chars

### DIFF
--- a/src/probe/resolve.rs
+++ b/src/probe/resolve.rs
@@ -242,7 +242,9 @@ pub fn extract_gnu_cc1_line(stderr: &str) -> Option<&str> {
         let base = base
             .len()
             .checked_sub(4)
-            .filter(|_| base[base.len() - 4..].eq_ignore_ascii_case(".exe"))
+            // `is_char_boundary` so non-ASCII basenames (e.g. `使用内建`
+            // from a localized gcc banner) don't panic the slice.
+            .filter(|&n| base.is_char_boundary(n) && base[n..].eq_ignore_ascii_case(".exe"))
             .map_or(base, |cut| &base[..cut]);
         base.eq_ignore_ascii_case("cc1") || base.eq_ignore_ascii_case("cc1plus")
     })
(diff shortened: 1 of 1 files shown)


jleni:
Thanks @AzureZee ! great finding!

jleni:
Thanks for tracking this down. The diagnosis is right, and it checks out end to end: `使用内建` is 12 bytes with boundaries at 0, 3, 6, 9 and 12, so `len() - 4` lands inside `内` and the slice panics exactly as you reported. I also reverted your guard while keeping your test, and it reproduces the panic at the same line and byte index, so the test is a real regression test rather than a passing formality.

One thing had to be fixed before it could land: `cargo fmt --all -- --check` rejects the new test's line wrapping, which fails the required `Check (Linux)` job. Nothing caught it here because CI never ran, since runs on fork PRs need maintainer approval first.

Rather than send you back for a whitespace change, I adopted the PR: https://github.com/kunobi-ninja/kache/pull/740 carries your commit `a912ea9` unchanged, with a separate formatting commit on top. Your authorship and your original commit object are preserved exactly. Once that merges I will close this one as superseded, with the credit staying with you.

One aside, not something you need to do: none of the probe spawn sites pin `LC_ALL=C`, which is the underlying reason the banner was localized in the first place. That is a pre-existing gap and belongs in its own change.

*(AI-assisted review; a maintainer owns the call.)*

jleni:
Landed in `main` as 8117f3f via #740, so I am closing this as superseded rather than rejected. The fix is yours; the only thing added on top was the formatting commit.

It went in as a squash, which is this repo's convention, so `main` carries a single commit with a `Co-authored-by: AzureZee <azureze@qq.com>` trailer rather than your original commit object. The trailer is linked to your account, so it registers as your contribution.

Thanks again for finding this. It is the kind of bug that stays invisible to everyone working in an English locale, and the report had everything needed to confirm it: the exact panic, the byte offset, and the localized banner that triggered it.

If you run into anything else in a non-English environment, please do open an issue. That path clearly gets less exercise than it should.

PR #758, opened 2026-08-14
fix(cli): report reclaimable bytes, not raw sizes, in kache clean

aqeelat:
## Summary

`kache clean` reported raw directory sizes as space it would free, but bytes shared with kache's store (hardlink/reflink) are NOT returned to disk when the directory is deleted — the store blob still references them. On a real 9-dir scan it claimed "would free 233.4 GiB" when only ~142.7 GiB is actually reclaimable.

The data was already in hand (`TargetEntry.cached_bytes` from the sharing probe), so this is a display/arithmetic fix only:

- `render_clean_dry_run`: "would free" now sums `size − cached_bytes` (saturating) instead of raw sizes
- `draw_clean` TUI header: `Selected:` total sums `size − cached_bytes`; per-row display and the raw dir totals are unchanged
- `remove_targets` (+ `--yes` and post-TUI callers): `freed` sums `size − cached_bytes` over dirs actually removed; tuple widened to carry `cached_bytes`
- Each summary adds a one-time note — e.g. `Dry run: would free 142.7 GiB (90.7 GiB shared with the store is kept)` — so the smaller number doesn't read as undercounting; the note is omitted when nothing is shared
- Windows / probe-less filesystems keep `cached_bytes = 0` (the sharing walk is `cfg(unix)`-gated), so their output is byte-for-byte unchanged

No changes to `kache list`, `kache stats`, or the scan/classifier logic.

## Test plan

- [x] `draw_clean_renders_target_table`: mixed shared/unshared entries; asserts `Selected: 1 (1.9 MiB)` (reclaimable) appears and the raw `4.8 MiB` total doesn't; header raw totals asserted unchanged
- [x] `render_clean_dry_run_formats_plural_profiles_and_fallback_paths`: asserts `would free 25 B` + `(5 B shared with the store is kept)` and that the raw `30 B` is not presented as freed
- [x] `remove_targets_*`: freed = `size − cached` over actually-removed dirs; shared remainder reported separately; failed dirs still not counted
- [x] `tests/cli_commands_test.rs` populated dry-run (non-macOS): asserts exact `would free 10.0 KiB` and that no shared note is printed for plain copies
- [x] `cargo test --bin kache -- clean remove_targets` (50 passed), `cargo test --test cli_commands_test clean` (1 passed), `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`: the only 3 errors are pre-existing on `main` in `remote_backend.rs` / `tui.rs` / `wrapper.rs` (files this PR does not touch); this PR introduces no new warnings

## Checklist

- [ ] `just check` passes locally (fmt + clippy + tests)

### DIFF
--- a/docs/commands/reference.mdx
+++ b/docs/commands/reference.mdx
@@ -256,6 +256,8 @@ kache clean -y           # non-interactive: delete all target/ dirs (alias: --ye
 
 This is useful for freeing disk space when you're done with a project or before a fresh build.
 
+Freed-space totals (`Selected:` in the selector, `would free` in a dry run, and `freed` after removal) count only bytes that return to disk: storage kache shares with the directory (hardlink or reflink) is kept by the store blob and reported as `shared with the store is kept`, not as freed.
+
 <Callout type="info">
 On macOS, the scan skips TCC-protected locations (`/System`, `/Library`, `/private`, `/Applications`, `/Volumes`, and home folders like `~/Desktop`, `~/Documents`, `~/Downloads`, `~/Library`, `~/Pictures`) to avoid permission prompts, so `target/` directories under those paths are never found or cleaned.
 </Callout>
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -2175,11 +2175,15 @@ fn draw_clean(
     use ratatui::prelude::*;
     use ratatui::widgets::*;
 
+    // Reclaimable bytes, not raw sizes: storage shared with the store
+    // (hardlink/reflink) is not returned to disk when the dir is deleted —
+    // the store blob still references it. The per-row and header totals
+    // above stay raw on purpose; only the "would delete" number shrinks.
     let selected_size: u64 = targets
         .iter()
         .zip(selected.iter())
         .filter(|(_, s)| **s)
-        .map(|(t, _)| t.size)
+        .map(|(t, _)| t.size.saturating_sub(t.cached_bytes))
         .sum();
     let selected_count = selected.iter().filter(|s| **s).count();
     let total_size: u64 = targets.iter().map(|t| t.size).sum();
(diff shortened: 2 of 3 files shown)


aqeelat:
Hey @jleni — `cargo clippy --workspace --all-targets -- -D warnings` (clippy 1.95.0) fails on `main` with 3 pre-existing errors, none related to my fixes. Do you want me to include them in this PR, or open a new PR for them?

**1. `src/tui.rs` — `clippy::collapsible_match`**

before:
```rust
KeyCode::Backspace => {
    if state.filter.pop().is_some() {
        reset_filtered_viewports(state);
    }
}
```
after:
```rust
KeyCode::Backspace if state.filter.pop().is_some() => {
    reset_filtered_viewports(state);
}
```

**2. `src/wrapper.rs` — `clippy::collapsible_if`**

before:
```rust
if let Ok(metadata) = std::fs::symlink_metadata(marker) {
    if metadata.file_type().is_symlink() || !metadata.file_type().is_file() {
        return None;
    }
    #[cfg(windows)]
    {
        use std::os::windows::fs::MetadataExt;
        if (metadata.file_attributes() & 0x400) != 0 {
            return None; // Refuse reparse points explicitly
        }
    }
}
```
after — note: clippy's machine-applied suggestion would leave the Windows reparse-point check as dead code inside the collapsed branch, silently disabling it; this rewrite folds it into the condition instead:
```rust
if let Ok(metadata) = std::fs::symlink_metadata(marker)
    && (metadata.file_type().is_symlink()
        || !metadata.file_type().is_file()
        || is_reparse_point(&metadata))
{
    return None;
}

#[cfg(windows)]
fn is_reparse_point(metadata: &std::fs::Metadata) -> bool {
    use std::os::windows::fs::MetadataExt;
    metadata.file_attributes() & 0x400 != 0
}

#[cfg(not(windows))]
fn is_reparse_point(_metadata: &std::fs::Metadata) -> bool {
    false
}
```

**3. `src/remote_backend.rs` — `clippy::nonminimal_bool`** (first five terms shown; the trailing-dot/segment checks are unchanged)

before:
```rust
let canonical = valid_empty
    || (!key.is_empty()
        && !original.starts_with('/')
        && !key.contains('\\')
        && !key.chars().any(char::is_control)
        && !(self.is_filesystem() && key.contains(':'))
        && /* ... */);
```
after:
```rust
let canonical = valid_empty
    || !(key.is_empty()
        || original.starts_with('/')
        || key.contains('\\')
        || key.chars().any(char::is_control)
        || (self.is_filesystem() && key.contains(':'))
        || /* ... */);
```

jleni:
Thanks for the contribution!
I can merge in a branch and then fix that!

jleni:
Thanks @aqeelat. We’ve adopted the contribution from pinned head `e042817a70bc77c35c757122883e2c2c5f7e4f51` into the maintainer-owned replacement https://github.com/kunobi-ninja/kache/pull/765. Your original commit and Author metadata remain unchanged as the first commit; our integration fixes are separate. We’ll keep this PR open until the replacement passes hosted CI and lands.

jleni:
Thanks @aqeelat — replacement #765 landed as [`fac24e8`](https://github.com/kunobi-ninja/kache/commit/fac24e860ca7061f22660e63af68b7b4b8d194bf). Your contribution is credited through its `Co-authored-by` trailer; source commit `e042817a` itself is not in `main` because the replacement was squash-merged. Closing this PR as superseded by #765, not rejected. Thank you for the contribution.

PR #775, opened 2026-08-19
fix(probe): pin LC_ALL=C on compiler probe spawns (#741)

### LINKED ISSUE #741: probe: pin `LC_ALL=C` on compiler probe spawns so localized output cannot reach the parsers
#739 fixed a panic in `extract_gnu_cc1_line` that only reproduced under a non-English `LANG`: `gcc -###` printed a localized banner (`使用内建 specs。` under `zh_CN.UTF-8`), and the `.exe`-suffix strip sliced through a multi-byte char. The guard is the right fix for that slice, but the underlying reason the parser ever saw localized text is that the probe spawns inherit the user's locale.

None of these pin a locale:

- `src/probe/mod.rs:378` (`cc -###`, `resolve_invocation`)
- `src/probe/mod.rs:469` (`cc --version`, `live_probe_diagnostic_for`)
- `src/probe/mod.rs:526` (`cc -###` re-run for the unresolved-probe stderr head)
- `src/probe/mod.rs:797` (`cc -###`)

Other subprocess spawns in the repo already do this, for example `src/cache_key.rs:3408` and `:3419` use `.env("LC_ALL", "C")`.

Setting `LC_ALL=C` on the probe spawns makes the parsed surface locale-independent, which matters for anything reading the text and not just for that one slice. Worth checking whether `version_line` is affected too, since it becomes key material.

This is defense in depth rather than a live bug: the `cc1`/`cc1plus` subprocess line itself is not localized, so the resolved tokens were correct even in the panicking case.

Ysh204:
## Description
Fixes #741

Pins `LC_ALL=C` on compiler probe subprocess spawns so that localized output (such as non-ASCII banner lines under non-English locales or localized `--version` strings) cannot reach probe parsers.

### Summary of Changes
- **`CcProber::probe`**: Added `.env("LC_ALL", "C")` to `Command::new(req.compiler).arg("--version")` so `version_line` (which is cache key material) is locale-independent.
- **`run_family_probe`**: Added `.env("LC_ALL", "C")` to `child_cmd` (`-E -P -x c`).
- **`resolve_invocation`**: Added `.env("LC_ALL", "C")` to `Command::new(compiler).arg("-###")`.
- **`live_probe_diagnostic_for`**: Added `.env("LC_ALL", "C")` to `--version` probe and unresolved stderr head re-run.
- **Test diagnostic fallback**: Added `.env("LC_ALL", "C")` to `cc -###` test re-run.
- **Test**: Added `probe_spawns_pin_lc_all_c` unit test verifying that compiler child processes receive `LC_ALL=C`.

## Verification
- `just fmt-check` passed cleanly.
- `cargo test probe_spawns_pin_lc_all_c` passed.
- `cargo test probe::` passed (64/64 tests).
- `just test` passed across all workspace crates.

### DIFF
--- a/src/probe/mod.rs
+++ b/src/probe/mod.rs
@@ -127,6 +127,7 @@ impl Prober for CcProber {
     fn probe(&self, req: &ProbeRequest<'_>) -> Result<ResolvedConfig> {
         // Compiler identity — `cc --version`.
         let output = Command::new(req.compiler)
+            .env("LC_ALL", "C")
             .arg("--version")
             .output()
             .with_context(|| format!("running `{} --version`", req.compiler))?;
@@ -265,6 +266,7 @@ fn run_family_probe(program: &str) -> Result<Option<ProbedFamily>, ()> {
     child_cmd
         .args(["-E", "-P", "-x", "c"])
         .arg(source_file.path())
+        .env("LC_ALL", "C")
         .env("KACHE_FAMILY_PROBE_ACTIVE", "1")
         .stdin(Stdio::null())
         .stdout(Stdio::piped())
@@ -376,6 +378,7 @@ fn resolve_invocation(
     per_tu_paths: &[String],
 ) -> Option<Vec<String>> {
     let output = Command::new(compiler)
+        .env("LC_ALL", "C")
         .arg("-###")
         .args(args)
         .output()
@@ -466,7 +469,11 @@ pub fn live_probe_diagnostic() -> LiveProbeDiagnostic {
 fn live_probe_diagnostic_for(compiler: &str) -> LiveProbeDiagnostic {
     // Absence is the one informational state; every other failure below is a
     // diagnostic that could not run and must surface as such.
-    match Command::new(compiler).arg("--version").output() {
+    match Command::new(compiler)
+        .env("LC_ALL", "C")
+        .arg("--version")
+        .output()
+    {
         Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
             return LiveProbeDiagnostic::NoCompiler;
         }
@@ -524,6 +531,7 @@ fn live_probe_diagnostic_for(compiler: &str) -> LiveProbeDiagnostic {
     // The prober discards the raw `-###` output; re-run it for the head,
     // which is the one fact an "unresolvable probe" report needs.
     let stderr_head = Command::new(compiler)
+        .env("LC_ALL", "C")
         .arg("-###")
         .args(&args)
         .output()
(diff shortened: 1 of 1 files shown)


jleni (commented):
Thanks for this. The `src/probe/mod.rs` change looks good: pinning `LC_ALL=C` on all five probe spawns is the right fix, `version_line` is cache key material so locale independence matters there, and the unix test that asserts every spawn saw `LC_ALL=C` is a nice touch. GNU gettext also ignores `LANGUAGE` when the locale is `C`, so there is no escape hatch left open.

The `Cargo.lock` change should come out though. It is two separate things and neither belongs in this PR.

**The `h2` bump is already on main.** `h2 0.4.16` landed in 977ad4d (#772). It shows up in your diff only because this branch is based on 13899ad, from before that commit. RUSTSEC-2026-0258 was a real advisory and the intent was fine, it is just redundant now.

**The `windows-sys` changes are a regression.** The lock moves 12 dependency edges down from `windows-sys 0.61.2`:

| crate | main | this PR |
|---|---|---|
| rustix, tempfile, seize, rustls-platform-verifier, winapi-util, errno | 0.61.2 | 0.52.0 |
| anstyle-query, anstyle-wincon, dirs-sys, nu-ansi-term, socket2, termina | 0.61.2 | 0.60.2 |

Those crates declare wide ranges (`rustix`, `tempfile` and `errno` use `>=0.52, <0.62`, `socket2` uses `>=0.60, <0.62`), and the workspace depends on `windows-sys = "0.61"` directly in `Cargo.toml`, so cargo unifies them onto 0.61.2. A fresh `cargo generate-lockfile` on current main moves one more edge *up* to 0.61.2 and moves nothing down, so this is not a resolution cargo would produce today. It is most likely a stale registry index or a hand resolved lock conflict. The practical cost is three copies of `windows-sys` compiled where main needs only the unified one. CI will not flag it because `--locked` accepts any valid lock.

Could you rebase on main and restore the lock so the PR touches `src/probe/mod.rs` only?

```
git fetch upstream main
git rebase upstream/main
git checkout upstream/main -- Cargo.lock
```

If the rebase leaves `Cargo.lock` untouched relative to main, the diff should be the probe change on its own.

jleni:
To add some context on the ask above: we have been bitten by lock file churn in PRs before, so as a general rule I would rather keep dependency updates in their own change. For this one please limit the fix to the `LC_ALL` change only for now. Security bumps like the `h2` one are worth doing, just as a separate PR where the lock diff is the whole point and easy to review on its own.

Ysh204:
Rebased on `upstream/main` and restored `Cargo.lock` so this PR touches `src/probe/mod.rs` only.

jleni:
Thanks!

PR #776, opened 2026-08-19
fix(probe): eliminate env-mutation race in probe_key tests (#742)

### LINKED ISSUE #742: flaky test: `probe_key_is_stable_for_the_same_binary` races with env-mutating tests
`cargo test --workspace` fails intermittently on macOS:

```
thread 'probe::cache::tests::probe_key_is_stable_for_the_same_binary' panicked at src/probe/cache.rs:304:9:
assertion `left == right` failed
  left: "5344d764a59b52eb2fd7d9c5aeafdd754b1b0eb47efc28a118e2747be5750ae0"
 right: "4cdf2ef2ffbb459e645aca032eca4d01b5d122c15a5474f6745a92446e8ed63f"
```

It passes reliably in isolation and only fails inside the full parallel run.

Cause: `probe_key` hashes the entire process environment (`env_fingerprint()` at `src/probe/cache.rs:30`, defined at `:81`). The test calls `probe_key` twice and asserts the results match, so any other test mutating the environment between the two calls changes the second hash. There are many such mutators running concurrently, for example in `src/config.rs` (around `:2444`, `:2812`, `:3107`) and `src/extra_inputs.rs:2732`.

The repo already knows about this hazard and serializes it elsewhere: `ENV_LOCK` in `src/cache_key.rs:6608` and `ENV_VAR_TEST_LOCK` in `src/store.rs`. This test takes neither.

Since CI runs plain `cargo test --workspace` (`Justfile:62`) with default parallelism, this can fail any PR's `Check (Linux)` job for reasons unrelated to the change.

Options: have this test take the shared env test lock, or compute the two keys from an explicit env map rather than the live process environment.

Found while reviewing #739; unrelated to that change.

Ysh204:
## Description
Fixes #742

Fixes flaky intermittent failure of `probe::cache::tests::probe_key_is_stable_for_the_same_binary` (and other `probe_key` tests) during parallel `cargo test --workspace` runs.

### Cause
`probe_key` incorporates `fingerprint_env(std::env::vars())` in the cache key calculation. The unit tests called `probe_key` across multiple points to assert stability or variation. When other tests running concurrently in parallel mutate process environment variables, the second `probe_key` invocation observes a mutated environment and produces a different hash, causing `assertion left == right failed`.

### Solution
- Extracted `probe_key_with_env(prober_id, req, env_vars: impl Iterator<Item = (String, String)>)` following the same pattern as `fingerprint_env`.
- Implemented `probe_key` to delegate to `probe_key_with_env(prober_id, req, std::env::vars())`.
- Updated `src/probe/cache.rs` unit tests (`probe_key_is_stable_for_the_same_binary`, `probe_key_changes_when_the_binary_changes`, `probe_key_changes_when_key_args_change`, `probe_key_differs_per_prober_id`) to provide an explicit, fixed environment rather than reading the live mutable process environment.
- Added `probe_key_changes_when_env_changes` test.

## Verification
- `just fmt-check` passed cleanly.
- `cargo test probe::cache::tests` passed (19/19 tests).
- `just test` (full workspace test suite) passed cleanly under parallel execution.

### DIFF
--- a/src/probe/cache.rs
+++ b/src/probe/cache.rs
@@ -24,10 +24,21 @@ const PROBE_SUBDIR: &str = "probes";
 /// compiler change `mtime` alone would miss. Deriving the key is a
 /// `stat`, not a process fork.
 pub fn probe_key(prober_id: &str, req: &ProbeRequest<'_>) -> Option<String> {
+    probe_key_with_env(prober_id, req, std::env::vars())
+}
+
+/// Content-addressed key for a probe evaluated against an explicit
+/// environment. Split from [`probe_key`] so key stability and variance
+/// are testable without racing against concurrent environment mutations.
+pub(crate) fn probe_key_with_env(
+    prober_id: &str,
+    req: &ProbeRequest<'_>,
+    env_vars: impl Iterator<Item = (String, String)>,
+) -> Option<String> {
     let resolved = resolve_program(req.compiler)?;
     let meta = std::fs::metadata(&resolved).ok()?;
     let fingerprint = compiler_fingerprint(&meta);
-    let env_fp = env_fingerprint();
+    let env_fp = fingerprint_env(env_vars);
 
     let mut h = blake3::Hasher::new();
     h.update(b"probe_schema:");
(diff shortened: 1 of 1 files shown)


jleni:
@Ysh204 Again Yash, Thanks! 🦾

Issue #610, opened 2026-07-25
packaging/nix/package.nix: document or drop the crates User-Agent override


jleni:
## Summary

Three small things in `nix/package.nix` that a review flagged. None is urgent and none blocks anything; they are grouped because they are all one-line-ish and all in the same file. Each is independent, so they can land separately or be closed individually if we decide the current behaviour is deliberate.

## 1. `apple-sdk_15` may be unnecessary

`nix/package.nix` adds `apple-sdk_15` to `buildInputs` on Darwin. nixpkgs' Darwin stdenv already provides an SDK, and in the currently locked nixpkgs the default is **14.4** while `apple-sdk_15` is **15.5**. Current nixpkgs guidance is to use the default unless the package needs APIs the default lacks.

If kache only uses ordinary filesystem, process, networking and xattr APIs, this is dead weight and pins us to a newer SDK than needed. Worth noting that SDK selection and deployment target are separate concerns, so this does not by itself mean "requires macOS 15", but a newer SDK does make accidental compatibility regressions easier to introduce.

Only a real `aarch64-darwin` build settles it. See the comment below for a local result.

## 2. The crates.io User-Agent override reaches into private plumbing

```nix
buildRustPackage = rustPlatform.buildRustPackage.override {
  importCargoLock = rustPlatform.importCargoLock.override {
    fetchurl = fetchurlWithCratesUserAgent;
  };
};
```

This works, but it overrides through `importCargoLock`'s internals rather than a supported interface, so a nixpkgs refactor can break it silently. Current nixpkgs already fetches crates from `static.crates.io` specifically to avoid the crates.io API rate limit and intermittent 403s, which is likely the problem this was working around in the first place.

Suggested: try a build with the locked nixpkgs and the override removed. If it succeeds, drop it. If there is still a reproducible CDN failure, keep it and add a comment linking the actual failure or upstream issue so the next person knows why it is there. Not a reason to switch to `cargoHash`; the `cargoLock` approach is the right one for an in-repo flake.

## 3. `channel = "1.95"` floats across patch releases

`rust-toolchain.toml` selects a minor series, not an exact compiler. rust-overlay resolves it to the newest `1.95.x` available in the locked overlay revision, so bumping the rust-overlay input can move the patch release without any visible change here. `flake.lock` keeps a given checkout reproducible, so this is not a correctness bug.

[...]

PR #788, opened 2026-08-22
fix(nix): drop unneeded apple-sdk_15 dependency (#610)

Ysh204:
## Summary
Drops `apple-sdk_15` from `nix/package.nix`.

nixpkgs' Darwin `stdenv` provides SDK 14.4 by default. `kache` only uses standard filesystem, process, networking, and xattr APIs and builds cleanly without macOS 15 SDK features. Dropping `apple-sdk_15` removes the extra dependency and avoids pinning to a newer SDK than needed.

Ref: #610 (Item 1)

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -2,7 +2,6 @@
   lib,
   rustPlatform,
   fetchurl,
-  apple-sdk_15,
   cacert,
   stdenv,
 }:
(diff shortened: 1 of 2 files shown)


jleni:
Merged main into your branch to resolve a conflict: we both fixed the same new Rust 1.98 clippy lint in native_archive.rs (yours in 09d4202, mine on main via #789), so the merge keeps main's spelling and your commit's intent is preserved. No action needed on your side; CI is rerunning.

PR #790, opened 2026-08-22
chore(ci): pin the rust toolchain in mise.toml

jleni:
The `[tools]` rust entry said `latest` while the comment right under it says versions are pinned so a tool release cannot break CI, and claimed rust deferred to rust-toolchain.toml. Neither held: an explicit `[tools]` version beats the idiomatic version file in mise's precedence, so `latest` won — and Rust 1.98.0 reaching the runners turned `-D warnings` red on every open PR at once (#789, #788's spurious failures).

Pin 1.98.0 (what main is green on today). Bump deliberately in a PR that also clears any new lints, matching how the mise CLI itself is pinned (#346). Part of #610's remaining cleanup discussion.

Validation: TOML parses; the pinned version is exactly what CI resolved today, so behavior is unchanged until the next deliberate bump.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -4,11 +4,16 @@
 idiomatic_version_file_enable_tools = ["rust"]
 
 [tools]
-rust = { version = "latest", components = "rustfmt,clippy" }
+# Pinned exactly, and this entry — not rust-toolchain.toml — is what CI
+# actually runs: an explicit [tools] version takes precedence over the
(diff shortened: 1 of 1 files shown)


jleni:
Closing: keeping rust on latest stable is the preferred tradeoff; the lint fallout when a new stable lands is accepted as routine maintenance.

Issue #811, opened 2026-08-24
L1: trusted node-local cache lifecycle for ephemeral CI runners


jleni:
Parent: #810

Deliver a bounded trusted node-local L1 cache for ephemeral self-hosted CI jobs, with explicit ownership and no public/fork exposure.

## Required behavior

- Define one supported per-node store/daemon lifecycle across one-job ARC runner pods.
- Preserve Kache's local SQLite/blob correctness under concurrent trusted jobs.
- Isolate trust domains so public, fork, or otherwise untrusted jobs cannot read or mutate the trusted cache.
- Bound disk with GC, reservations, health, and node-local hit/miss/byte metrics.
- Fail open to ordinary per-job L2/v3 behavior when L1 is unavailable.
- Document runner/platform wiring and operational rollback.

## Acceptance

- Warm trusted nodes download only misses.
- Isolation tests prove untrusted jobs receive no node-cache mount/socket/credentials.
- Concurrent-job, daemon restart, corrupt local entry, disk-pressure, and fallback tests.
- Five clean cold-node and warm-node representative runs.

Issue #810, opened 2026-08-24
Two-level cache for ephemeral CI: node-local L1 plus packed remote prefetch


jleni:
## Problem

ARC creates a fresh runner pod for each job, so every Linux job cold-hydrates Kache from the S3 remote. A representative large Rust job reported:

- 4,188 cached crates / 20 compiled
- 5.0 GB downloaded across 4,371 objects at 78 MB/s
- 12.6 GB restored, all copied
- 169 ms average hit overhead; about 12 minutes aggregate
- aggregate remote phases: 139 s open/setup, 65 s transfer, 104 s extract, 172 s import

The hit rate is excellent, but the transport is doing a distributed rebuild for every ephemeral pod.

## Design

Implement both layers:

### L1: bounded node-local cache for trusted runner pools

- Persist a local Kache store per builder node with explicit multi-job and daemon lifecycle support.
- Never mount writable persistent cache state into public or fork-capable runners.
- Bound storage with GC and expose bytes, entries, evictions and local-hit metrics.
- Preserve content validation and isolate trust domains; a job must not poison another trust domain.

### L2: immutable packed prefetch for cold and untrusted runners

- Publish content-addressed transport packs selected by manifest, namespace and Cargo.lock shards.
- Fetch a small number of packs instead of thousands of individual S3 objects.
- Verify cache recipe, entry metadata and blob hashes before import.
- Keep existing v3 objects as a compatibility fallback and use create-only publication.
- Avoid unbounded duplication with bounded pack size, retention and GC metadata.

## Runner prerequisites

Larger builder disks and scheduling that accounts for ephemeral storage. Both are tracked in the private platform repositories.

## Acceptance criteria

- At least 90% fewer remote requests on a representative cold Rust job.
- No increase above 10% in bytes downloaded for the same build shape.
- Warm trusted-node runs download only misses, with reported L1/L2 byte counts.
- Demonstrated fallback from missing or corrupt packs to existing v3 objects.
- Isolation test proving an untrusted job cannot read or mutate another trust-domain cache.
- Before and after wall-clock benchmark across at least five clean runs per arm.

jleni:
## Delivery graph — revision 16

Progress sequence: 16

| Node | Scope | State | Evidence / dependency |
| --- | --- | --- | --- |
| D1 | Pin scope and architecture against current `origin/main` | complete | base refreshed to `77ef004`; #811 and #812 remain native open subissues until their action/rollout acceptance is complete |
[...]

PR #813, opened 2026-08-24
feat: separate runtime state from cache store

jleni:
## Summary

- add `KACHE_RUNTIME_DIR` / `[cache] runtime_dir`, defaulting to the existing cache directory for compatibility
- keep the persistent store, SQLite index, upload queue, and GC coordination under `KACHE_CACHE_DIR`
- move job-owned daemon IPC/state/logs, transfer logs, and build-session markers under the runtime directory
- preserve old-daemon effective-config compatibility and config-TUI round trips
- document the trusted shared-store boundary and per-job runtime setup

## Why

Ephemeral ARC runner pods can safely reuse one trusted node-local store only when each concurrent job owns distinct daemon/runtime state. This is the Kache prerequisite for #811; runner isolation, fallback probes, disk-pressure handling, and five representative runs remain in the parent issue.

## Validation

- `cargo check --tests`
- runtime config precedence/default/ignore-env tests
- concurrent shared-store / isolated-runtime tests
- session-marker isolation test
- socket/event path tests
- config-TUI round trip and old-daemon wire compatibility tests

Part of #811 and #810.


jleni:
Exact head `374e70c` is now fully green. Authoritative CI run `32732389908` completed successfully with Linux/macOS/Windows tests and E2E, Nix, audit, Kani, core/service mutation suites, both changed-code mutation shards, and coverage/check. @emmanuelm41 @ainhoa-a ready for independent review.

jleni:
Exact current head `e882a6b18f08330ae7bb5c3755479b549c980a8d` is fully green in authoritative run `32735003800` attempt 2. The rerun passed the unchanged probe test that previously hit transient Linux `ETXTBSY`; both changed-code mutation shards, core/service mutations, Kani, Linux/macOS/Windows tests and E2E, Nix, audit/check, and the mutation aggregate all passed. Ready for independent review.

Issue #812, opened 2026-08-24
L2: publish batch packs and catalogs


jleni:
Parent: #810

Deliver immutable multi-entry remote prefetch packs that remove the cold-run small-object fan-out without weakening Kache's content-addressed trust boundary.

## Required behavior

- Select pack contents from the existing manifest/shard/namespace planning context.
- Use content-addressed, create-only pack objects and bounded metadata.
- Verify every cache-key binding, `meta.json`, artifact name/size, and blob hash before publication.
- Import verified entries in a batch; do not re-hash bytes already verified while streaming extraction.
- Missing, corrupt, incompatible, or incomplete packs fall back to current v3 objects.
- Keep on-demand v3 reads available while a pack is fetched.
- Report pack/v3 request counts, bytes, validation/fallback, and real plan wall time.

## Acceptance

- At least 90% fewer remote requests on the representative cold-node arm.
- No more than 10% extra remote bytes versus object-by-object v3 prefetch.
- Deterministic corruption, traversal, partial-pack, concurrent publication, and fallback tests.
- Five clean representative runs for both object and pack arms.

jleni:
Reader, verification and v3 fallback shipped in https://github.com/kunobi-ninja/kache/pull/814. Nothing in production publishes packs or catalogs yet, so the request-count acceptance cannot be measured. The open work is the publisher.

PR #814, opened 2026-08-24
feat(remote): add immutable packed prefetch

jleni:
## Summary

- add deterministic immutable multi-entry pack/catalog codecs, selectors, and content-addressed keys
- reject digest, length, recipe, duplicate, traversal, framing, expiry, entry-count, and 256 MiB cap violations
- add create-only backend support, including S3 `If-None-Match: *`
- discover the newest valid catalog from manifest/namespace/Cargo.lock shard context
- fetch bounded packs, stream-verify each artifact once, and batch-import verified entries in one transaction
- preserve valid siblings when one packed entry is invalid and fall back per key to existing v3 objects
- keep packed prefetch asynchronous and leave demand v3 reads available while a pack GET is blocked
- report pack/v3 requests and bytes, validation/fallback counts, and real plan wall time

## Remaining scope

The immutable pack publisher is not implemented yet. It must group entries deterministically, verify or publish v3 fallback before publishing a catalog, create pack metadata before the catalog, and prove concurrent create-only publication. Adding that remote-publication capability is explicitly approval-gated; no live S3/R2 publication is authorized by this PR.

Representative object/pack five-run proof remains part of #812 acceptance after the publisher, compatible release, and controlled caller rollout.

## Validation

- full binary suite: 2,094 passed
- `cargo clippy --all-targets -- -D warnings`
- `git diff --check`
- rebased onto current `origin/main` (`f2f8be1`); exact-head CI is running for `cb696e4`

Part of #812 and #810.

### DIFF
--- a/.cargo/mutants.toml
+++ b/.cargo/mutants.toml
@@ -149,6 +149,16 @@ gitignore = true
 # than the already-observed assertion failure. Exclude only that whole-body
 # replacement; the override resolver and every branch inside the real method
 # remain mutation-tested.
+#
+# The packed-prefetch format has three inclusive-boundary operator mutants
+# that cannot change an observable result. A catalog which passed structural
+# validation cannot approach the 64 MiB encode cap: its 65,536 fixed-shape
+# entries, 16,384 pack wrappers, and two 4 KiB selector fields remain far
+# below it. On decode, an object of exactly the 16-byte header length has no
+# JSON index/payload and is rejected either way; likewise an index ending at
+# EOF has no payload for the required non-empty, non-zero-length entry set.
+# Exact-limit and first-real-overshoot tests retain coverage of every reachable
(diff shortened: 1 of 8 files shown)


jleni:
Exact-head verification for `0c32ac50573259a6775be3bfdbebfd6684d782d8`:

- authoritative CI run `32745034768`: success, 38 executed jobs green
- all 23 changed-code mutation shards passed, plus service/core mutation and Kani
- formerly surviving shard 3: 10 mutants — 9 caught, 1 unviable
- formerly surviving shard 19: 9 mutants — 8 caught, 1 unviable
- local post-fix gate: 2,123 unit tests plus every integration binary, clippy `-D warnings`, fmt, and focused probes (6 caught, 2 unviable)

Publisher scope remains intentionally absent pending explicit authorization; this evidence covers the current reader/import slice only.

PR #815, opened 2026-08-24
fix(report): measure observed transfer throughput

jleni:
## Summary

- record exact transfer start/finish timestamps and include SQLite import in download elapsed time
- separate import lock wait from import execution
- report bytes over the observed wall span plus peak download concurrency
- retain legacy service-time rates for compatibility, but label them accurately
- keep transfer-event v2 and existing report JSON consumers compatible

## Why 78 MB/s was misleading

The old `body_throughput_mbps` divided bytes by the sum of each GET body duration. With concurrent downloads, that is a per-request service-time diagnostic—not aggregate link throughput. It also excluded object-open latency, extraction, and the serialized import stage. The representative run spent more cumulative time opening thousands of ~1 MiB objects than reading their bodies, then serialized verification/import behind a global store mutex.

This PR makes the next representative run show honest wall-span throughput and peak concurrency. Packed prefetch and batched import remain in #812.

## Validation

- report tests: 47 passed
- prefetch transfer-event regression
- CLI report tests
- JSON schema contract test
- `cargo clippy --all-targets -- -D warnings`

Part of #810 and #812.


jleni:
Exact head `9e86eba` is now fully green, including all 11 changed-code mutation shards, core/service mutation suites, Kani, Linux/macOS/Windows tests, E2E, Nix, audit, and coverage. @emmanuelm41 @ainhoa-a ready for independent review.

jleni:
Rebased exact head `932b152` is fully green on authoritative run `32735000255`, including all 11 changed-code mutation shards, core/service mutation suites, Kani, Linux/macOS/Windows tests and E2E, Nix, audit, coverage/check. @emmanuelm41 @ainhoa-a independent approval is the remaining merge gate.

Issue #117, opened 2026-05-20
cc: key Firefox debug-info and clang argument-wrapper flags


jleni:
## Context

Firefox C/C++ compiles on macOS frequently pass through kache because debug-info and clang argument-wrapper flags are not accepted by the cc flag allow-list.

Evidence source:

```sh
jq -r 'select(.result=="passthrough" and (.passthrough_reason|contains("cc: unsupported flag(s):"))) | .passthrough_reason' ~/Library/Caches/kache/events.jsonl
```

In the sampled Firefox build/configure log, these flags appeared in 4,275 single-source passthroughs:

```text
4275 -gdwarf-4
4275 -gsimple-template-names
4275 --start-no-unused-arguments
4275 -mllvm=-dwarf-linkage-names=Abstract
4275 --end-no-unused-arguments
```

## Proposed Work

Support these Firefox debug-info and clang wrapper flags:

```text
-gdwarf-4
-gsimple-template-names
--start-no-unused-arguments
-mllvm=-dwarf-linkage-names=Abstract
--end-no-unused-arguments
```

## Acceptance Criteria

- These flags no longer force passthrough for otherwise cacheable single-source `-c` invocations.
- Cache keys distinguish output-affecting debug-info variants.
- Tests cover the paired `--start-no-unused-arguments` / `--end-no-unused-arguments` case and the `-mllvm=` form.

## Out of Scope

- General support for every possible `-mllvm` flag.
- Link-mode caching.

jleni:
Shipped via #140 (squash-merged into `dev` as commit `5a51fef`). The five debug-info and clang argument-wrapper flags are now classified in `CC_FLAGS` — three `CapturedByProbe`, two `NoObjectEffect`. Closing manually because GitHub's auto-close only fires when a PR with a close keyword merges into the *default* branch (`main`); these PRs land in `dev` first, so the keyword is inert until the next `dev → main` promotion.

PR #843, opened 2026-08-26
feat(cc): key cc-rs's Apple debug flag -gdwarf-2 via the resolved probe

### LINKED ISSUE #838: cc: cache Apple debug builds using -gdwarf-2
## Problem

kache refuses otherwise cacheable C/C++ object compiles on Apple targets when `cc-rs` adds `-gdwarf-2`:

```text
unsupported|cc unsupported flag(s): -gdwarf-2 — not yet
```

A 24-hour local event sample on Apple Silicon contained 1,084 C/C++ passthroughs carrying this flag. It was the only unsupported flag in 710 of them; the remaining 374 also carried separately unmodeled flags. The sampled roots were `aws-lc-sys` 0.40.0 and 0.44.0, but the source is generic: [`cc-rs` selects DWARF 2 for every Apple target](https://github.com/rust-lang/cc-rs/blob/b27570c6644e4f5a43ea4561878a80ea334c5a14/src/lib.rs#L4118-L4131) when debug info is enabled.

Apple clang 21 resolves the driver flag under `-###` as:

```text
-debug-info-kind=standalone -dwarf-version=2 -debugger-tuning=lldb
```

kache already classifies exact `-gdwarf-4` as `CapturedByProbe` because the resolved-token hash distinguishes its DWARF version. The same evidence exists for version 2, but the current table intentionally contains only the Firefox-observed version 4 from #117.

Environment for the sample:

- kache 0.15.1; verified still unmodeled on current `main` at `dfd326672f6a`
- macOS 26.6.2, arm64
- Apple clang 21.0.0
- rustc 1.97.0
- local cache only

## Proposed solution

Add an exact `Matcher::Exact("-gdwarf-2")` row classified as `CapturedByProbe`. Keep the existing exact-version policy rather than accepting `-gdwarf-*` broadly.

Acceptance criteria:

- `-gdwarf-2` no longer refuses an otherwise cacheable single-source object compile.
- The resolved invocation is required and contributes `-dwarf-version=2` to the key.
- Tests distinguish DWARF 2 from DWARF 4 and from the default/no-flag invocation.
- A compiler whose resolved probe cannot establish the flag's effect still fails closed to passthrough.

## Alternatives considered

`[cc] extra_allowlist_flags = ["-gdwarf-2"]` raw-keys the exact token and is a workable local override. It leaves a standard `cc-rs` Apple debug configuration uncached by default on every installation and does not record the stronger resolved-probe contract already used for `-gdwarf-4`.

## Use case

Rust workspaces with native dependencies built through `cc-rs` on macOS debug/test profiles. In the sample, supporting this one flag would make 710 currently refused translation units eligible for caching immediately and remove one blocker from another 374.

Xuanwo:
## Summary

Closes #838.

`cc-rs` adds `-gdwarf-2` for every Apple target when debug info is enabled, so effectively every native-dep TU of a macOS Rust debug build refused caching with `unsupported flag(s): -gdwarf-2` (24h sample: the sole blocker in 710 of 1,084 passthroughs carrying it; sampled roots aws-lc-sys 0.40.0/0.44.0).

This adds an exact `Matcher::Exact("-gdwarf-2")` row classified `CapturedByProbe`, mirroring `-gdwarf-4` (#117). Apple clang resolves the driver flag under `-###` to `-debug-info-kind=standalone -dwarf-version=2 -debugger-tuning=lldb`, so the resolved-token hash keys DWARF 2, DWARF 4 and the no-flag baseline apart; a driver whose probe cannot resolve keeps refusing fail-closed. The exact-version policy is kept: `-gdwarf-3`/`-gdwarf-5`/`-gdwarf` still refuse.

## Test plan

- [x] `cargo test --bin kache -- compiler::cc::tests probe::resolve::tests` — 232 passed, including new `classifier_accepts_gdwarf2_issue_838` (classified + probe forced), `gdwarf_version_changes_resolved_tokens` (real Apple clang 21 `-###` fixtures: DWARF 2 vs 4 vs no-flag key apart), and the extended fail-closed guard (`-gdwarf-2` refuses when `-###` resolves nothing)
- [x] `cargo test --test integration_test` — 30 passed, including new `test_cc_gdwarf2_caches_and_keys_dwarf_version_issue_838`: cold `-gdwarf-2` compile stores (miss, not passthrough), identical invocation hits, `-gdwarf-4`/no-flag do not hit the DWARF-2 entry
- [x] `cargo fmt --check`
- [ ] `just check` (clippy): 3 pre-existing `-D warnings` failures on base (`remote_backend.rs`, `tui.rs`, `wrapper.rs`), unrelated to this change; none in touched files

## Checklist

- [x] New behaviour is covered by tests
- [x] Commit messages follow conventional commits
- [x] User-visible changes reflected in docs — no CLI/config/output change; classifier rows are documented at the flag table itself


jleni:
this seems to need rebase too

jleni:
Thanks @Xuanwo — we’ve adopted the contribution at pinned head `9e1b2c95c514888f761586289b2136f85d26f09a` into maintainer-owned replacement https://github.com/kunobi-ninja/kache/pull/851 because current `main` conflicts after #842/#845. Your authorship is preserved in the imported commit; #843 will remain open while the replacement is reviewed and validated.

PR #845, opened 2026-08-26
fix(cc): classify -fsanitize-undefined-strip-path-components=-1 so aws-lc-sys 0.44 jitterentropy TUs cache (#840)

### LINKED ISSUE #840: cc: cache aws-lc-sys jitterentropy builds using -fsanitize-undefined-strip-path-components=-1
## Problem

kache refuses the `aws-lc-sys` 0.44.0 jitterentropy translation units that carry clang's UBSan path-stripping option:

```text
unsupported|cc unsupported flag(s): -gdwarf-2 -mno-omit-leaf-frame-pointer -fsanitize-undefined-strip-path-components=-1 — not yet
```

A 24-hour local event sample on Apple Silicon contained six such passthroughs. `aws-lc-sys` adds the exact option after checking compiler support because [Clang UBSan metadata at `-O0` can retain the full source path](https://github.com/aws/aws-lc-rs/blob/f464440d1fd3983ce9fb023e9eaf1698530919a2/aws-lc-sys/builder/cc_builder.rs#L541-L548).

Apple clang 21 preserves the option in the resolved `-cc1` invocation:

```text
-fsanitize-undefined-strip-path-components=-1
```

The option affects path metadata rather than executable semantics, but it still changes object bytes and should be keyed. Its value is visible in the resolved invocation, so `CapturedByProbe` can distinguish it without treating it as no-effect.

Environment for the sample:

- kache 0.15.1; verified still unmodeled on current `main` at `dfd326672f6a`
- macOS 26.6.2, arm64
- Apple clang 21.0.0
- rustc 1.97.0
- local cache only

## Proposed solution

Support the observed `-fsanitize-undefined-strip-path-components=-1` spelling through the resolved invocation. An exact matcher is sufficient for the reported workload; a value-bearing matcher is also sound only if every accepted value is preserved by the resolved-token key.

Acceptance criteria:

- The exact flag no longer refuses an otherwise cacheable single-source object compile.
- The resolved invocation is required and its path-component value contributes to the key.
- Tests distinguish `=-1`, absence of the flag, and any other accepted value.
- Unsupported compiler or probe shapes remain passthroughs.

## Alternatives considered

`[cc] extra_allowlist_flags = ["-fsanitize-undefined-strip-path-components=-1"]` raw-keys the exact token and resolves the local case. Upstream support avoids leaving a reproducibility flag from a widely used TLS dependency as a permanent cache bypass.

## Use case

Rust workspaces using `aws-lc-sys` 0.44.0 on macOS. The count is small today, but after the separately reported `-gdwarf-2` and `-mno-omit-leaf-frame-pointer` gaps are fixed, these six jitterentropy units become the remaining fixed flag-classification floor.

Xuanwo:
## Summary

Fixes #840.

aws-lc-sys 0.44 adds `-fsanitize-undefined-strip-path-components=-1` to its six jitterentropy TUs after a compiler support probe (Clang's UBSan check metadata at `-O0` can embed the full source path). The flag was unclassified, so every one of those TUs refused into passthrough.

The option is keyed, not ignored: it changes object bytes. Verified against Apple clang 21.0.0 `-###`, the driver forwards the option — value included — verbatim into the resolved `-cc1` line, so the row is `CapturedByProbe` and the resolved-token hash carries the value. The matcher is `Exact` on the one observed value: every integer value clang accepts rides through to cc1 (`=0`/`=2` verified), so widening is sound when a workload needs it, but unobserved spellings keep refusing (the `-gdwarf-4` precedent). Unsupported shapes stay passthroughs: gcc rejects the option outright, a value cc1 cannot parse (`=abc`) fails the compile before anything can be cached, and an unresolvable `-###` probe refuses rather than under-keys.

## Test plan

- [x] `compiler::cc` unit tests: `=-1` classifies as `CapturedByProbe`, the full aws-lc-sys 0.44 jitterentropy invocation no longer refuses, and the probe is forced; `=0`/`=2`/`=-2`/value-less spellings keep refusing
- [x] `probe::resolve` unit test on a frozen Apple clang 21 `-###` fixture (`src/probe/testdata/clang_ubsan_strip_path.txt`): the value survives into the semantic tokens; presence, absence, and a different value are three distinct token lists
- [x] Live end-to-end test `test_cc_ubsan_strip_path_components_caches_issue_840` (gated on a clang driver and a resolvable probe; skips elsewhere): cold compile caches, repeat hits, flag-less compile keys apart — run for real on Apple clang 21.0.0 / arm64 macOS
- [x] `cargo test --bin kache`: 2166 passed, 1 pre-existing environment flake (`daemon::tests::find_daemon_pids_*`, also fails on clean `origin/main`)
- [ ] `just lint` — blocked by 3 pre-existing clippy errors on clean `origin/main` with the pinned 1.95 toolchain (`src/remote_backend.rs:212`, `src/tui.rs:724`, `src/wrapper.rs:4396`); untouched by this PR

## Checklist

- [x] `just check` passes locally (fmt + clippy + tests) — fmt and tests pass; clippy fails only on the pre-existing errors above
- [x] User-visible changes (CLI, config, output) are reflected in the README or relevant docs — no user-facing surface change; the flag table is the doc


jleni:
Thanks @Xuanwo ! You are on fire today ! :)

Xuanwo:
> Thanks @Xuanwo ! You are on fire today ! :)

I'm using kache locally and found that it didn't cache correctly on my projects, so working on fix it!

jleni:
Keep them coming!!!
Nothing better than using things to fix the bugs!

jleni:
could you rebase?

Xuanwo:
> could you rebase?

done

PR #871, opened 2026-08-28
fix(store): read the executable bit from the compiler output, not the staged blob

glebpom:
#648 taught restore to honour the mode bit recorded at insert time, because a
`[[test]] harness = false` target supplies its own `main` and is compiled with
neither `--test` nor `--crate-type` — nothing in the rustc argv says
"executable", so the recorded bit is the only signal.

#822 then moved that read from the compiler's output to the staging snapshot:

    -  let executable = is_executable(&metadata);      // the compiler's output
    +  let executable = is_executable(&staged_meta);   // the staging temp

`stage_blob_from_source` prefers `try_reflink`, whose Linux implementation
creates the destination with `File::create` before the FICLONE ioctl. The
snapshot therefore lands at `0o666 & !umask` whatever the source was, so on
every CoW filesystem (btrfs, XFS-with-reflink, ZFS >= 2.2, bcachefs) a 0o755
binary is recorded `executable: false`. Restore then falls back to
`Other("rustc:unknown")`'s `Hardlink` strategy, never applies 0o755, and cargo
fails the run with "Permission denied (os error 13)" — the exact failure #648
fixed.

`fs::copy` and macOS `clonefile` both preserve the mode, which is why ext4 CI
and macOS saw nothing. In a 0.16 store on ZFS the harness-shaped entries
(`crate_types: []`, extensionless ELF output) split 144 executable / 237 not,
nondeterministically: FICLONE fails for blocks not yet written out, and the copy
fallback then preserves the bit by accident.

The snapshot is authoritative for *bytes* — hashing it rather than the live
output is the whole point of #822 — but never for permissions, and
`CachedFile::executable` already documents itself as "whether the source file
had the executable bit set at store time". A failed stat becomes an error rather
than a silent `false`, which is precisely the wrong value; staging opens the same
path a line later regardless, so only the message changes.

Two layers of coverage, because CI could observe none of this:

  - A `#[cfg(test)]` thread-local makes staging emulate a Linux reflink (copy,
    then chmod 0o644) on any filesystem. Without it ext4 only ever exercises the
    mode-preserving copy fallback, and the new tests hold for free.
  - A `cow-filesystem` job runs the filesystem-sensitive suites on a loopback
    btrfs mount, gated on a `cp --reflink=always` probe so a mount that cannot
    clone fails the job instead of silently becoming a second ext4 lane.
    `scratch_dir()` honours `KACHE_TEST_SCRATCH_DIR` to get there.

`custom_harness_test` reproduces the reported failure on ZFS without this fix
and passes with it.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -633,6 +633,73 @@ jobs:
       - name: Build flake checks
         run: nix flake check --keep-going --print-build-logs
 
+  # --- Copy-on-write filesystem coverage ---
+  # Every hosted runner is ext4, where `try_reflink` always fails and the
+  # `fs::copy` fallback preserves permissions. That makes an entire class of bug
+  # unobservable in CI: the store's reflink ingest creates its staging temp with
+  # `File::create` (umask mode, no +x), so a mode-losing ingest looks perfectly
+  # healthy here and only breaks on btrfs / XFS-with-reflink / ZFS >= 2.2 /
+  # bcachefs / APFS. #648 fixed the restore half of that and #822 silently
+  # reverted it through the ingest half; both shipped green. This lane exists so
+  # a third round can't.
+  cow-filesystem:
+    name: CoW filesystem (btrfs)
+    runs-on: ubuntu-latest
+    needs: changes
+    if: github.event_name != 'pull_request' || needs.changes.outputs.code == 'true'
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
+        with:
+          # Pin mise CLI — see the `check` job's note (v2026.6.10 regression).
+          version: 2026.6.9
+          install_args: rust
+          cache: false
+      - name: Mount a loopback btrfs volume
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y btrfs-progs
+          # Sparse: the image only occupies what the tests actually write.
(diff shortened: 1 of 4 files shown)


jleni:
Thanks @glebpom !
We will have a look!

jleni (commented):
Thanks @glebpom — the mechanism checks out. `#822` moved `is_executable` from the compiler output onto the staging snapshot, and Linux `try_reflink` really does `File::create` before `FICLONE`, so a 0o755 binary is recorded `executable: false` on every Linux CoW filesystem. Restore then takes `Other("rustc:unknown")`'s `Hardlink` path and never applies 0o755. That is the `#648` failure again. The two unit tests actually force the mode drop, so they cannot pass because ext4's copy fallback kept `+x`.

We're taking this onto a maintainer-owned branch so we can land the store fix and the btrfs lane without approving fork CI on a workflow change. `#871` stays open; I'll link the replacement here once it exists. Authorship stays yours.

I read the code against `#822` and `link::try_reflink`. I did not run the new tests.

jleni:
Adopted at pinned head `407deec99e6eb9cf417c4040b18f3b2c08e51f24` into maintainer-owned replacement https://github.com/kunobi-ninja/kache/pull/877. Your authorship is preserved in the imported commit; #871 stays open while the replacement is reviewed and validated.

jleni:
Thanks @glebpom — this landed through maintainer replacement https://github.com/kunobi-ninja/kache/pull/877 as https://github.com/kunobi-ninja/kache/commit/847bb6d1115183463d6e0fafdf4fea0f530e9320.

GitHub squash-merged, so the individual imported commit was not kept on `main`. Your authorship is on that squash as `Co-authored-by: Gleb Pomykalov <glebpom@gmail.com>`. Closing this PR as superseded, not rejected.

PR #877, opened 2026-08-29
fix(store): read the executable bit from the compiler output, not the staged blob

jleni:
## Summary

Maintainer-owned adoption of #871 from @glebpom, pinned at contributor head [`407deec99e6eb9cf417c4040b18f3b2c08e51f24`](https://github.com/nervix-io/kache/commit/407deec99e6eb9cf417c4040b18f3b2c08e51f24).

`#822` started recording `CachedFile.executable` from the staging snapshot. Linux `try_reflink` creates that snapshot with `File::create` before `FICLONE`, so a 0o755 binary is stored `executable: false` on btrfs / XFS-with-reflink / ZFS ≥ 2.2 / bcachefs. Restore then takes `Other("rustc:unknown")`'s `Hardlink` path and never applies 0o755 — the `#648` failure again. This reads the mode from the compiler output and adds a btrfs CI lane so the next regression cannot hide on ext4.

This replacement is maintainer-opened so we do not approve fork CI on a workflow change.

## Provenance and attribution

| Source commit | Imported commit | Original author | Disposition |
| --- | --- | --- | --- |
| `407deec99e6eb9cf417c4040b18f3b2c08e51f24` | `539916092c9065dce26102ef2e98677533bf33b9` | Gleb Pomykalov `<glebpom@gmail.com>` | Cherry-picked with `-x`; logical patch identical (matching patch-ids). Source OID and signature not retained because `main` is linear. |

Maintainer work is separate:

- `0584b5924aadeddfe07d57a8d6272c10c10376ed` pins the CoW job to `-p kache` and drops APFS from the affected-filesystem list (`clonefile` preserves mode).

## Review

- Safety: store and tests are ordinary Rust. The CoW job is GitHub-hosted `ubuntu-latest` with `sudo` only to mount a loopback btrfs volume; no secrets, no self-hosted runners, no trigger/permission/environment change. `publish-crates.yaml` only adds the job name to `require-ci-green` (release-only). CODEOWNERS still applies to `.github/workflows/`. Original fork CI was not approved.
- Purpose: **ACCOMPLISHES** — ingest records the source mode bit; restore already honours it.
- Tests: **COVERED** — two unit tests force a mode-dropping ingest so ext4 cannot hide the bug; `custom_harness_test` plus `KACHE_TEST_SCRATCH_DIR` is the real CoW path.
- Docs: **NOT REQUIRED** — no user-facing enumeration this change falsifies. Existing 0.16 CoW-store entries that already recorded `executable: false` stay wrong until rebuilt; call that out in the next release notes.

## Validation

- Static combined-diff and provenance review: clean.
- `git diff --check origin/main..HEAD`: clean.
- Both local commits carry valid maintainer signatures; imported author identity is preserved.
- Contributor code and tests were not executed on a maintainer machine.
- Fresh full hosted CI against current `main`: pending this replacement PR.

## Test plan

- [ ] Required Linux, macOS, Windows, Nix, E2E, and mutation checks pass on the exact replacement head.
- [ ] New `CoW filesystem (btrfs)` job is green (not a required check today; it does gate tag releases and crates publish).
- [x] Contributor and maintainer changes remain separate and attributable.
- [x] No public documentation claim becomes stale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -652,9 +652,9 @@ jobs:
   # unobservable in CI: the store's reflink ingest creates its staging temp with
   # `File::create` (umask mode, no +x), so a mode-losing ingest looks perfectly
   # healthy here and only breaks on btrfs / XFS-with-reflink / ZFS >= 2.2 /
-  # bcachefs / APFS. #648 fixed the restore half of that and #822 silently
-  # reverted it through the ingest half; both shipped green. This lane exists so
-  # a third round can't.
+  # bcachefs. macOS clonefile preserves mode, so APFS is not in that set.
+  # #648 fixed the restore half of that and #822 silently reverted it through
+  # the ingest half; both shipped green. This lane exists so a third round can't.
   cow-filesystem:
     name: CoW filesystem (btrfs)
     runs-on: ubuntu-latest
@@ -707,8 +707,8 @@ jobs:
           RUSTC_WRAPPER: ""
         run: |
           set -euo pipefail
-          cargo test --test custom_harness_test
-          cargo test --bins -- store:: link:: wrapper::tests::materialize_cached_artifact
+          cargo test -p kache --test custom_harness_test
+          cargo test -p kache --bins -- store:: link:: wrapper::tests::materialize_cached_artifact
       - name: Report mount usage
         if: always()
         run: df -h /mnt/cow || true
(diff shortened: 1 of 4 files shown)


jleni:
Babysitting this replacement. Windows test compile is red on `ModeDroppingIngest` (`-D dead-code` because the helper is test-only and the tests that use it are Unix-only). Fixing that as a maintainer commit; Gleb's imported commit stays untouched.

PR #888, opened 2026-08-31
test: stabilize process and environment-sensitive cases

jleni:
Two test races were failing checks before feature code was evaluated.

- The daemon PID test now waits for its spawned fixture to complete exec before checking the process list.
- Manifest-key tests now use the shared environment lock and restore `KACHE_MANIFEST_KEY`.

The first race reproduced twice in the local mutation baseline. The second reproduced in Windows CI, which read `only-this` from a neighboring test. Production behavior is unchanged.

Validation:

- `just pr`
- Positive discovery test passed 20 consecutive runs
- Positive and negative PID discovery tests passed together
- Manifest-key group passed 20 repeated parallel runs
- Mutation diff: 0 applicable mutants, 0 survivors, 0 timeouts

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -10351,7 +10351,17 @@ mod tests {
             .expect("spawn fake daemon");
         let fake_pid = child.id();
 
-        let found = super::find_daemon_pids();
+        // `Command::spawn` returns after fork, before the child necessarily
+        // completes exec and exposes the fake daemon argv to `pgrep -f`.
+        // Poll that transition instead of racing it once.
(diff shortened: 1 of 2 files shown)


jleni:
Monitoring exact head `c5672bc` through required CI.

jleni:
Windows CI on `c5672bc` exposed a second shared test race: manifest-key readers could observe `KACHE_MANIFEST_KEY=only-this` from a neighboring test. `929a425` serializes and restores those readers. Full local `just pr` is green on the new signed head; monitoring run 33390127174.

Issue #904, opened 2026-09-01
chore: fewer entries at the repository root


jleni:
The repository root has 41 tracked entries. Most are pinned there by tooling, but about a third are not, and the packaging material in particular is spread across six sibling directories holding 18 files between them.

This issue records what can move, what cannot, and every reference that would have to change. A cleanup PR can be written from it without rediscovering the inventory.

## Already handled

Four committed store staging files (`stage-1-0.tmp`, `stage-60717-191.tmp`, `stage-new-1.tmp`, `stage-old-1.tmp`) are scratch from `src/store.rs` staging tests that a `git add -A` picked up. They are removed, with a `/stage-*.tmp` ignore rule, in the wave-0 fixes PR.

## Pinned to the root, cannot move

`Cargo.toml`, `Cargo.lock`, `rust-toolchain.toml`, `.cargo/` (cargo) · `flake.nix`, `flake.lock` (Nix requires a flake at the root) · `Cross.toml` (cross) · `Justfile` (just) · `mise.toml` (mise) · `deny.toml` (cargo-deny default) · `README.md`, `LICENSE` (crates.io and convention) · `AGENTS.md` · `.github/`, `.gitignore`, `.gitattributes`, `.dockerignore`, `.pinact.yaml` · `src/`, `crates/`, `tests/`, `fuzz/`, `docs/`, `scripts/`, `assets/`

## Candidate: fold packaging into `packaging/`

`packaging/` already exists and holds only `debian/`. Six root entries could live under it:

| Move | Files |
| --- | --- |
| `apt_config/` → `packaging/apt/` | 1 |
| `aur/` → `packaging/aur/` | 2 |
| `charts/` → `packaging/charts/` | 9 |
| `docker/` → `packaging/docker/` | 2 |
| `nix/` → `packaging/nix/` | 2 |
| `docker-bake.hcl` → `packaging/` | 1 |

References that must be updated, found by grepping the workflow, Nix, TOML, shell, Justfile and HCL files:

- **apt_config** — `.github/workflows/package-publish.yml`, `scripts/apt/generate-repo.sh`
- **aur** — `.github/workflows/ci.yml`, `.github/workflows/package-publish.yml`, `Justfile`, `scripts/aur/test-vcs-pkgver.sh`, `scripts/aur/vcs-pkgver.sh`
- **charts** — `.github/workflows/ci.yml`, `Justfile`, `scripts/check-version-consistency.sh`
[...]

PR #906, opened 2026-09-01
chore: gather packaging under packaging/

jleni:
Follows #905 and closes the main item in #904. Six root entries held 18 files of distribution material between them; they now sit together.

```
apt_config/     -> packaging/apt/
aur/            -> packaging/aur/
charts/         -> packaging/charts/
docker/         -> packaging/docker/
nix/            -> packaging/nix/
docker-bake.hcl -> packaging/docker-bake.hcl
```

`packaging/debian/` was already there. `CONTRIBUTING.md` and `SECURITY.md` move to `.github/`, which GitHub reads for both.

Root entries: **36 to 28.** `flake.nix`, `flake.lock`, `Cross.toml`, `Justfile`, `mise.toml`, `rust-toolchain.toml` and `deny.toml` stay because their tools require the root.

## The edit that could break a release

`packaging/nix/package.nix` reaches out of its own directory for the source fileset — `root = ../.`, plus `../Cargo.toml`, `../Cargo.lock`, `../assets`, `../crates`, `../src`, `../tests/fixtures` and the `lockFile`. Every one became `../../`. That is the change here that fails loudly rather than quietly.

Verified by forcing the derivation to evaluate, which is what exercises the fileset:

```
$ nix eval ".#packages.<system>.kache.drvPath"
"/nix/store/m4pk5w86r0rhs284wcfdq4ka76h7b5nj-kache-0.16.1.drv"
```

`packaging/nix/module.nix` refers to `./package.nix` beside itself, so it needed nothing.

## Every reference updated

Found by grepping the workflow, Nix, TOML, shell, Justfile and HCL files, then re-grepped afterwards. The only remaining match is the phrase "no nix/installer surface" in a `publish-crates.yaml` comment, which is English, not a path.

- **apt** — `package-publish.yml` `CONF_DIR`, `scripts/apt/generate-repo.sh` default
- **aur** — `package-publish.yml` (the `ls -d` discovery, the `sed`, the `pkgbuild:` input, three comments), `scripts/aur/vcs-pkgver.sh`, `scripts/aur/test-vcs-pkgver.sh`
- **charts** — `Justfile` (`helm lint`, two `Chart.yaml` rewrites), `scripts/check-version-consistency.sh`, `ci.yml` `helm package`
- **docker** — `Justfile` e2e build, `docker-bake.hcl` `dockerfile =`, `service-image.yml` path filters
- **docker-bake.hcl** — three `Justfile` recipes, `service-image.yml` filter and bake call
- **nix** — `flake.nix` (`callPackage`, both `moduleFor` calls)

## Validation

Each moved path exercised with its own tool rather than by inspection:

| Check | Result |
| --- | --- |
| `nix eval .#packages.<system>.kache.drvPath` | derivation evaluates |
| `helm lint packaging/charts/kache-service` | 1 chart, 0 failed |
| `docker buildx bake -f packaging/docker-bake.hcl --print service` | resolves `packaging/docker/service.Dockerfile` |
| `scripts/check-version-consistency.sh` | `kache == kache-core == dep-pin == 0.16.1` |
| `cargo fmt --check`, `scripts/check-docs.sh` | clean |

`scripts/aur/test-vcs-pkgver.sh` reports 5 passed, 3 failed — **pre-existing**. I ran the same script on unmodified `origin/main` in a separate worktree and got the identical 5/3. The failures are the throwaway repo needing signed tags, which `fix/e2e-fixture-signing` addresses separately.

## What a reviewer should still watch

`package-publish.yml` runs on tags, so its APT and AUR paths are not exercised by this PR. They are covered by grep and by the version-consistency and pkgver scripts, but the first real proof is a release. The AUR and APT changes are confined to their own hunks if a surgical revert is ever needed.


jleni:
Pushed `e85a31a`. The `Nix package` job was right to fail.

`flake.nix` builds a `nixSources` fileset so that editing Rust does not invalidate the formatting check, and one of its elements is the directory itself — a bare `./nix` with no file component. My sweep looked for `nix/`, so it matched `./nix/package.nix` and `./nix/module.nix` and walked straight past this one. After the move the path does not exist, `lib.fileset.unions` refuses it, and every `formatting` check plus the `treefmt-check` derivation goes down with it.

The real mistake was in how I validated. I evaluated the package derivation, which proved the `../../` fileset paths resolve but never touched this one. Re-verified with the command CI actually runs:

```
$ nix flake check --all-systems --no-build
✅ checks.{x86_64,aarch64}-linux.package
✅ checks.{x86_64,aarch64}-linux.formatting
✅ checks.aarch64-darwin.{package,formatting}
✅ checks.{x86_64,aarch64}-linux.nixos-module
EXIT=0
```

Also re-swept for bare directory references with no trailing slash, which is the pattern that hid this. The only remaining matches are already-updated paths, `just` group labels such as `[group(*docker*)]`, the `docker` command itself, and prose.

Issue #762, opened 2026-08-15
epic(cc): transparent compile-to-link cache coverage


jleni:
## Goal

Make Kache a practical, transparent C/C++ compiler cache: ordinary build-system recompiles remain cacheable, compiler-name shims can intercept builds without per-project configuration, and opt-in link caching safely covers the largest remaining passthrough class.

## Active workstreams

- [x] #744 — restore caching for ordinary recompiles whose output file already exists
- [x] #310 — support compiler-name shims on PATH without recursion
- [x] #1029 — classify `--` as end-of-options (shipped in #1031)
- [x] #259 — cache whole-program link outputs and their complete sidecar set (PR https://github.com/kunobi-ninja/kache/pull/1040)
- [x] #825 — verify the resolved probe on cross gcc drivers (PR https://github.com/kunobi-ninja/kache/pull/1040)
- [x] #826 — key the remaining closed-set target `-m` flags (PR https://github.com/kunobi-ninja/kache/pull/1040)
- [x] #1028 — include candidate under a file makes the compile uncacheable (ENOTDIR) (PR https://github.com/kunobi-ninja/kache/pull/1040)
- [x] #1030 — Windows linker wrapper script makes rustc links uncacheable (PR https://github.com/kunobi-ninja/kache/pull/1040)

## Delivery order

1. Fix #744 without regressing special-file and symlink write semantics.
2. Add the transparent compiler entry point in #310.
3. Build opt-in link caching in #259 after determinism and portability gates are proven.

## Completed context

#337 established compiler-driver conformance, launcher scenarios, structured passthrough reasons and coverage gates. It remains closed; this epic tracks only the concrete remaining product path.

## Exit criteria

- CMake configuration changes can recompile into existing regular object paths while still reading from and writing to Kache.
- `/dev/null`, device, FIFO, directory and symlink behavior remains compatible with the real compiler.
- A POSIX compiler-name shim resolves and invokes the real compiler without re-entering Kache; unsupported platforms and missing compilers fail clearly.
- Link keys cover ordered objects, archives, libraries, response files, linker flags and linker identity.
- Binary, import-library, map and debug sidecars are stored and restored atomically.
- Link caching stays opt-in until path portability and reproducibility negative controls pass.
- Representative isolated-checkout CMake/Firefox scenarios demonstrate object and link hits without false hits or path leakage.

## Non-goals

- A persistent C/C++ preprocess memo from #562 until workload evidence and an invalidation proof exist.
[...]

PR #928, opened 2026-09-03
perf(cc): let the preprocessor memo cross checkouts

jleni:
Refs #762. Follows #927.

#927 made the memo survive new metadata for unchanged bytes, which fixed rebuilding in the tree where the memo was written: the same-tree warm phase went from 575 preprocessor spawns to zero and halved its key time. It did nothing for a second checkout, and that is the phase this benchmark is built on. Measured on hk after #927:

| phase | preprocessor spawns | aggregate key |
| --- | ---: | ---: |
| same-tree warm | 0 | 20.5s |
| cross-clone warm | 575 | 42.2s |

## Why it could not cross

Two things pinned it to one directory.

- The memo key folded the raw working directory and the raw preprocess argv. Both carry the checkout root, so two trees never computed the same key.
- Recorded input paths were absolute, so even a shared key could not have resolved them elsewhere.

## What changed

Both now go through the prefix maps the expansion is already hashed under, so the memo is expressed in the same coordinates as the thing it memoises.

- The key folds the mapped working directory, the mapped argv, and the map *targets* only. The sources are the per-checkout roots this change exists to exclude; the targets are the sentinels both trees share, and they are what decides whether two mappings mean the same thing.
- Inputs are recorded under their mapped name and resolved back through the inverse at lookup. The recorded path is tried first, so a rebuild in place still matches on metadata and reads nothing.
- Schema label to v3, so no record written under the old rules is read under the new ones.

## On the inverse being inexact

Two roots can share one sentinel, so the inverse offers every candidate, most specific first. Picking wrong is not a correctness problem: the caller compares contents, so a candidate whose bytes differ is a miss and the expansion is recomputed. And two files with equal contents at one mapped name contribute the same expansion by construction, so either answer is the same answer. A mapped name matching no sentinel yields nothing and the memo is skipped.

## Verification

- Unit tests for the mapping and its inverse: longest source wins, a name recorded in one checkout resolves into the other, two roots behind one sentinel offer both, an unmapped absolute path is its own answer, a relative name has none.
- End to end with a real compiler over two checkouts of one source: before this change the second preprocessed; after it, the second hits the memo and does not.
- `mise exec -- just check`: fmt, clippy `-D warnings`, full workspace tests. Draft for the mutation shards.

## What a reviewer should still watch

- The benchmark number is the claim to check: the cross-clone warm phase should drop from 575 spawns to zero and its key time toward the same-tree figure. I will dispatch it once this is on main.
- Inputs outside every mapped root, system headers and the toolchain, are recorded verbatim and behave exactly as before.

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -2564,6 +2564,13 @@ enum FileHashCache<'db> {
 /// are exactly the ones metadata alone rejects.
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub(crate) struct CcPreprocessMemoInput {
+    /// The input's prefix-mapped spelling: what it is called in the form the
+    /// expansion was hashed in, and therefore the same string from any
+    /// checkout the maps cover. The reader turns it back into a local path.
+    name: String,
+    /// Where the file was, and what its metadata said, when recorded. The
+    /// path here is local to the recording checkout and is only a fast path;
+    /// a reader that resolved `name` elsewhere ignores it.
     #[serde(flatten)]
     fingerprint: FileFingerprint,
     /// blake3 of the file's contents when the expansion was hashed.
@@ -2710,7 +2717,11 @@ impl<'db> FileHasher<'db> {
     /// written; a content hash can, because a file that changes afterwards
     /// simply fails the next comparison. Any database, decoding, or metadata
     /// uncertainty is still a miss.
-    pub(crate) fn cc_preprocess_memo_lookup(&self, memo_key: &str) -> Option<String> {
+    pub(crate) fn cc_preprocess_memo_lookup(
+        &self,
+        memo_key: &str,
+        resolve: impl Fn(&str) -> Vec<PathBuf>,
+    ) -> Option<String> {
         let cache = self.cache.as_ref()?;
         let record = match cache.get_cc_preprocess_memo(memo_key) {
             Ok(record) => record?,
@@ -2739,7 +2750,7 @@ impl<'db> FileHasher<'db> {
             return None;
         }
         for expected in &inputs {
-            if !self.memo_input_is_unchanged(expected) {
+            if !self.memo_input_is_unchanged(expected, &resolve) {
                 return None;
             }
         }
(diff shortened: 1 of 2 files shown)


jleni:
Force-pushed after the e2e caught a real hole in the first attempt, worth recording here rather than only in the commit message.

The original resolution tried the path the record was written at before the reader's own, on the reasoning that a rebuild in place would then match on metadata and skip the read. The `relocate-modified` phase failed with `min_misses: expected >= 1, actual 0`: a second checkout with an **edited** source was answered by the untouched original still sitting at the recorded path, so the memo returned an expansion of source that was no longer being compiled, and the edited unit hit. That is precisely the stale-artifact class the phase exists to detect.

Now only the reader's own resolution counts. When the two trees are the same it returns that path anyway, so the metadata fast path is unchanged; when they are not, the record's own path is never consulted. There is a unit test for exactly this shape: identical copy in another tree hits, edited copy in another tree misses, and the second half of the assertion makes sure it is not simply rejecting everything foreign.

End to end over two checkouts: first tree preprocesses, identical second tree hits with zero preprocessor runs, edited second tree misses.

jleni:
Second force-push. The Windows lanes found a second real defect, and it is the same class as the first: my mapping was written as if paths were UTF-8 text everywhere.

`cc_mapped_path` was running paths through `cc_memo_os_bytes`, which encodes UTF-16 on Windows, and then interpreting the result as UTF-8. Every mapped name came out mangled, matched no sentinel, and so no Windows memo could ever resolve. `e2e-cpp-hello` caught it exactly where it should: `max_preprocessor_runs: expected <= 0, actual 1` on the warm phase, a scenario that passes on main. It now uses the path as text, the way the maps are applied to argv.

My unit test was Unix-centric too and failed on Windows for its own reason: a sentinel spelling like `/kache/root` is not an absolute path there, so the "an unmapped absolute path is its own answer" case needs a host-appropriate spelling. It now uses one, and the half-empty-map case asserts the property that holds on both hosts rather than an exact path.

Worth saying plainly: three defects in this change so far, and CI found all three where local testing did not. The e2e phases and the Windows lanes are doing exactly the job they exist for.

PR #941, opened 2026-09-04
bench: show the perf gate where the warm build's time went

jleni:
The gate compares one number — warm wall-clock, head against merge base. That
catches a build getting slower and misses a change that moves time *between*
wrapper phases, because the total does not move. Shifting 200 ms per crate out
of `compile` and into `store` passes today.

The phase totals already existed: the report aggregates them, and the Perfetto
trace draws them per crate. The gate never read them, because the bench result
JSON did not carry them.

## What changed

`PhaseMetrics` gains a `phases` object — startup, key, the dep-info pre-pass
inside it, lookup, wait, restore, store, and the unattributed remainder,
summed over the cacheable crates of each build phase.

The gate prints them as a collapsed second table under the wall-clock rows:

```
| phase                  | merge base | PR head |   delta |
| ---------------------- | ---------: | ------: | ------: |
| startup                |     500 ms |  500 ms |   +0.0% |
| key                    |    8000 ms | 3000 ms |  -62.5% |
|   dep-info pre-pass    |    6500 ms |    0 ms | -100.0% |
| lookup                 |     900 ms |  900 ms |   +0.0% |
| restore                |    2000 ms | 2000 ms |   +0.0% |
| store                  |    1200 ms | 1400 ms |  +16.7% |
| unattributed           |     400 ms |  400 ms |   +0.0% |

dep-info pre-pass spawns: 330 (merge base) / 0 (PR head).
```

**Reported, never blocking.** A phase can move a long way on a small absolute
number, and moving time between phases is often the intended effect of a
change rather than a fault. The reviewer is the one who can tell which. What
this removes is the blind spot, not the judgement.

The phases are summed over cacheable crates, so they deliberately do not add
up to the wall clock: a parallel build overlaps them and passthrough compiles
are absent. The table says so.

## Edge cases

- Rows where both sides are zero are dropped rather than printed as an
  unchanged `0.0%`, which would read as "measured and did not move".
- A result with no phase data at all — an external cache backend, or events
  predating the schema that carries these fields — prints no table.

## Verification

Exercised against synthetic result JSON for three cases: a phase shift (table
above), a result with no phase data (no table, exit 0), and a real regression
(still fails, `FAIL — warm build +10.0%`).

### DIFF
--- a/crates/kache-e2e/src/bench_runner.rs
+++ b/crates/kache-e2e/src/bench_runner.rs
@@ -3711,6 +3711,59 @@ struct PhaseMetrics {
     /// Storage savings — restore method (reflink/hardlink/copy) and
     /// content dedup, from the report's `storage` section.
     storage: StorageInfo,
+    /// Where the phase's wrapper time went, summed over cacheable crates.
+    /// The wall-clock above says a build got slower; this says which part of
+    /// the wrapper did it.
+    phases: PhaseTimes,
+}
+
+/// Wrapper time by phase, in wrapper order, summed over every cacheable crate
+/// of one build phase.
+///
+/// These are the same numbers the Perfetto trace draws per crate. Carried in
+/// the result JSON so the per-PR gate can compare them between the head and
+/// its merge base: a change that moves 200 ms per crate out of `compile` and
+/// into `store` leaves the total wall-clock alone and is invisible without
+/// this breakdown.
+///
+/// Every field is zero for a build whose events predate schema 17, and for
+/// external cache backends, which have no wrapper to instrument.
+#[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct PhaseTimes {
+    /// Process start to wrapper entry.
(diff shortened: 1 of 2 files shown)


jleni:
Also fixes a pre-existing Linux failure in this crate that this PR's mutation lane surfaced.

`write_cache_otlp_or_warn_is_false_when_kache_exits_nonzero` writes a shell stand-in and execs it. Linux refuses to exec a file any process still holds open for writing (ETXTBSY), and this suite runs in parallel and spawns processes, so a child can inherit the writable fd of a stand-in another test just wrote. macOS does not enforce that at all.

Nothing runs this crate's tests on Linux except the mutation lane's baseline, and only when a change touches the crate — so the failure sat here until now:

```
[bench] warning: kache telemetry write failed: Text file busy (os error 26)
assertion failed: marker.is_file()
error: cargo test failed in an unmutated tree, so no mutants were tested
```

The two tests that spawn the stand-in now retry, bounded, until it actually ran.

PR #996, opened 2026-09-09
fix(macos): resolve dsymutil objects from executable output directory

### LINKED ISSUE #995: macOS: store-time dsymutil cannot resolve output-relative OSO paths
## Bug

On Apple Silicon macOS with kache 0.18.0, compiling a tiny Rust executable with debug info emits this warning during kache's store-time `dsymutil` invocation:

```text
warning: (arm64) demo.demo.1a701118a94152c5-cgu.0.rcgu.o unable to open object file: No such file or directory
```

The build succeeds and the executable runs. The debug object exists, but its path resolves only when `dsymutil` runs from the output directory.

## Minimal reproduction

https://github.com/Dzejkop/kache-dsym-repro

On Apple Silicon macOS with Nix and Xcode Command Line Tools:

```sh
nix run github:Dzejkop/kache-dsym-repro
```

The flake pins kache 0.18.0 and Rust 1.93.0. Apple's linker and `dsymutil` come from the host; the script records tool versions alongside the results. Each run uses a fresh temporary directory and isolated kache cache/config/runtime, with remote caching disabled.

The program has no dependencies:

```rust
#[inline(never)]
fn answer() -> u32 { 42 }
fn main() { println!("{}", answer()); }
```

All four cases use `-Cdebuginfo=2 -Csplit-debuginfo=unpacked` to keep the debug objects available. Verified results:

| Case | Build warnings | `dsymutil --dump-debug-map` warnings from run root | From output directory |
| --- | ---: | ---: | ---: |
| Plain rustc | 0 | 0 | 0 |
| rustc with kache's `-oso_prefix` flag | 0 | 1 | 0 |
| kache, executable caching enabled | 1 | 1 | 0 |
| kache, executable caching disabled | 0 | 0 | 0 |

All four executables print `42`. The script asserts these results. Exit zero means the bug was reproduced.

The flag-only case adds `-Clink-arg=-Wl,-oso_prefix,<absolute-output-directory>/`; it does not invoke `dsymutil` during compilation. The kache-enabled case emits the warning during compilation itself.

## Suspected cause

In v0.18.0, `macos_oso_prefix_flag_inner` in `src/compiler/rustc.rs` injects `-oso_prefix` for cached debug links, making OSO paths relative to the output directory. `MacOsPlatform::package_debug_bundle` in `src/compiler/platform.rs` then invokes `dsymutil` without setting its working directory.

Running the same binary's debug-map inspection from its output directory resolves the warning without changing or rebuilding any artifact. This points to a mismatch between output-relative paths and the inherited working directory.

A potential fix is to run bundle generation from the appropriate output directory, using absolute binary and bundle paths. I have not tested an upstream source patch or verified the completeness of the generated `.dSYM`.

Expected behavior: store-time debug bundle generation resolves the existing objects without warnings.

Related history: #319, #685, #880.

Dzejkop:
## Summary

Fixes #995.

Cached macOS debug links use `-oso_prefix` to make OSO object paths relative to the executable's output directory. Store-time `dsymutil` inherited the caller's working directory, so it warned that existing objects could not be opened and produced bundles missing their source debug info.

Resolve the binary and bundle paths before running `dsymutil` from the executable's output directory. Regression coverage checks relative command arguments and verifies that a real bundle contains source and function debug info after linking with `-oso_prefix`.

Reproduction: https://github.com/Dzejkop/kache-dsym-repro

## Test plan

- Local Nix flake override of the reproduction: kache build warnings drop from 1 to 0, and the generated `.dSYM` contains source debug info. All four comparison cases pass.
- Reporter also confirmed a fresh `world-id-indexer` build succeeds without these warnings after installing the pinned fix.
- Original v0.18.0 fails the same `--expect-fixed` assertions.
- All 18 targeted debug-bundle tests pass.
- Replacing the child cwd with the original caller cwd makes the new regression test fail on the missing source compile unit; the fix was restored afterward.
- `CARGO_HOME=<isolated home> RUSTUP_TOOLCHAIN=1.98.1 just check` (local installed toolchain; the repository pin is 1.98.0). The isolated home avoids a machine-global Cargo `[env] KACHE_CONFIG` overriding fixtures that test project config discovery.

The reproduction's README includes a pinned command for this fork. Existing cached bundles are not repaired by this change.

## Checklist

- [ ] CI changed-line mutation testing passes (draft PR per AGENTS.md; not run locally)
- [x] New behavior is covered by regression tests
- [x] Commit follows conventional commits
- [x] Reproduction and before/after verification are documented in the linked repository

### DIFF
--- a/src/compiler/platform.rs
+++ b/src/compiler/platform.rs
@@ -253,6 +253,7 @@ impl Platform for MacOsPlatform {
 
 /// Resolve paths before changing the child's cwd: cached macOS debug links
 /// use `-oso_prefix` to make OSO records relative to the binary's output dir.
+#[cfg_attr(not(target_os = "macos"), allow(dead_code))]
 fn debug_bundle_command(binary: &Path, bundle_dir: &Path) -> Result<Command> {
     let binary = std::path::absolute(binary)?;
     let bundle_dir = std::path::absolute(bundle_dir)?;
--- a/src/compiler/platform.rs
+++ b/src/compiler/platform.rs
@@ -621,14 +621,12 @@ pub(crate) mod tests {
             command.get_current_dir(),
             Some(cwd.join("target/debug/deps").as_path())
         );
-        assert_eq!(
-            command.get_args().collect::<Vec<_>>(),
-            vec![
-                cwd.join(binary).into_os_string(),
-                "-o".into(),
-                cwd.join(bundle).into_os_string()
-            ]
-        );
+        let args = command.get_args().collect::<Vec<_>>();
+        assert_eq!(args.len(), 3);
+        // Windows absolute() normalizes separators; compare paths, not argv bytes.
+        assert_eq!(Path::new(args[0]), cwd.join(binary));
+        assert_eq!(args[1], "-o");
+        assert_eq!(Path::new(args[2]), cwd.join(bundle));
     }
 
     #[test]
--- a/src/compiler/platform.rs
+++ b/src/compiler/platform.rs
@@ -219,12 +219,7 @@ impl Platform for MacOsPlatform {
         //   2. lldb's adjacent-bundle lookup is by `<binary>.dSYM` sibling
         //      path, which is exactly this location.
         let bundle_dir = binary.with_file_name(format!("{file_name}.dSYM"));
-        let status = match Command::new("dsymutil")
-            .arg(binary)
-            .arg("-o")
-            .arg(&bundle_dir)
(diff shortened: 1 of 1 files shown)


jleni:
Thanks @Dzejkop !!

jleni:
@Dzejkop somethign similar happened in Windows :/

Dzejkop:
oh sick, thanks for merging!

PR #999, opened 2026-09-10
fix(native): admit DWARF-bearing Mach-O members to the structural archive hash

### LINKED ISSUE #998: native_archive: DWARF-bearing Mach-O members take the path-bound fallback, re-keying every cc-built -g archive per checkout
## What happened

On macOS, a `cc`-built `static=` archive whose objects carry DWARF is refused by the Mach-O gate in `portable_static_archive_hash` and lands in the `path-ar-v1` fallback, which binds the lexical absolute archive path. That is every dev-profile build of a cc-rs crate, and `ring` on Apple targets unconditionally (`-gfull`, #857). Two checkouts of the same commit therefore never share the crate's rlib entry, and the miss cascades through every dependent's `--extern` content hash. This is the per-checkout re-key that the closing comment on #691 asked to confirm on the next macOS run.

Evidence, `ring 0.17.14` inside a workspace, two checkouts of the same commit on one machine, `KACHE_LOG=trace cargo build -p ring 2>&1 | grep '\[key:ring\]'` in each: 187 of 189 key lines are identical. The only differences:

```
[key:ring] link_lib_content:static=ring_core_0_17_14_=path-ar-v1:345ec (<checkout-a>/target/debug/build/ring-89d9a8590d815664/out/libring_core_0_17_14_.a)
[key:ring] link_lib_content:static=ring_core_0_17_14_=path-ar-v1:32ba3 (<checkout-b>/target/debug/build/ring-89d9a8590d815664/out/libring_core_0_17_14_.a)
[key:ring] final=ca7bb6a9c95ce4fb
[key:ring] final=2ef18cc4c2aec700
```

The two archives are byte-identical: same size (478064), same SHA-1, 24 members, `ar x` + per-member SHA-1 shows zero differences. The member objects' DWARF is already path-normalized by `kache cc` (`DW_AT_comp_dir "/kache/cc-root"`, `DW_AT_name "/kache/cc-root/crypto/…"`), and cc-rs's `ZERO_AR_DATE` leaves every member timestamp at 0. `kache why-miss ring` classifies the second build as `dup` (identical output, different key). Cross-checkout hit rate for the workspace: 39%; `ring` plus its 12 dependents recompile in the second checkout.

Control experiment: with `CFLAGS=-g0` (no DWARF, everything else unchanged) the same archive hashes `bsd-ar-v2:091ebd…` in both checkouts, `ring`'s final key converges, the second checkout's `ring` rebuild is a `local_hit` (128 ms), the 13-crate cascade drops from 17.7 s to 4.75 s, and the hit rate goes from 39% to 72%. So the gate, not the bytes, is what re-keys.

## Expected behaviour

A `cc`-built archive whose members are structurally valid Mach-O objects should get the structural BSD hash whether or not they carry DWARF:

- the DWARF bytes are hashed as member content like everything else, so a path a compiler left in `DW_AT_comp_dir` is identity-bearing, never a false hit;
- the `archive(member)` `N_OSO` concern that motivated the exclusion is already handled for every cached macOS debug link by ld64 `-oso_prefix` relative to `--out-dir` (`src/compiler/rustc.rs`) and the store-time `.dSYM` (#319) — exactly how the debug-bearing Rust objects in every dev-profile rlib are handled today. Once rustc has bundled a `static=` archive into the rlib, its members are indistinguishable from those.

`-oso_prefix` is only injected on the cache-miss link of an executable kache is about to cache, and a passthrough link keeps absolute `N_OSO` paths — but nothing is restored from such a link, so admitting DWARF members cannot produce a restored binary whose debug map is wrong.

## Steps to reproduce

1. `kache init` on macOS; a workspace with `ring` (or any cc-rs crate) in the dev profile.
2. A second checkout of the same commit (`git worktree add ../b`).
3. In both: `cargo clean -p ring && KACHE_LOG=trace cargo build -p ring 2>&1 | grep '\[key:ring\]'`; diff the two outputs — only `link_lib_content:…=path-ar-v1:…` and `final=` differ.
4. `kache why-miss ring` → `dup` for the second build; `cmp` of the two `libring_core_0_17_14_.a` files → identical.

## Environment

- `kache --version`: kache 0.19.0
- OS and version: macOS 26.6.2, arm64
- Rust toolchain: rustc 1.95.0 (59807616e 2026-04-14); Apple clang 21.0.0 (clang-2100.1.1.101)
- Install method: cargo install
- Remote backend: none

## Logs

```
Why `ring` missed:

  Last dup: 2026-09-10T11:58:07 (key: b149ea1259d8)
  Stored entries for `ring` (2 total):
    - key: b149ea1259d8 (size: 7.8 MiB, hits: 0, profile: dev, type: lib, target: host) <-- entry-miss key (stored after compile)
    - key: ca7bb6a9c95c (size: 7.8 MiB, hits: 0, profile: dev, type: lib, target: host)
  Diagnosis: key mismatch -- 1 other entry exist but it matched the current build inputs
```

Trace diff between the two checkouts (everything else identical):

```
< [key:ring] link_lib_content:static=ring_core_0_17_14_=path-ar-v1:345ec (…/obscura/target/debug/build/ring-89d9a8590d815664/out/libring_core_0_17_14_.a)
> [key:ring] link_lib_content:static=ring_core_0_17_14_=path-ar-v1:32ba3 (…/obscura-kache-tune/target/debug/build/ring-89d9a8590d815664/out/libring_core_0_17_14_.a)
< [key:ring] link_lib_content:static=ring_core_0_17_14__test=path-ar-v1:83607 (…)
> [key:ring] link_lib_content:static=ring_core_0_17_14__test=path-ar-v1:430b9 (…)
```

## Additional context

- Root cause in `src/native_archive.rs`: `macho_segment_is_portable` rejects `__DWARF`, and `macho_section_is_portable` rejects any `S_ATTR_DEBUG` / `__debug_*` section, so `parse_known_no_debug_macho_object` returns `None` and `compute_static_lib_hash` (`src/cache_key.rs`) takes the fallback that hashes `std::path::absolute(path)` together with the bytes. `paths.base_dirs` / `KACHE_BASE_DIR` cannot help: the fallback uses the raw path, not the normalized one.
- #857 / #860 made `ring`'s C compiles cacheable *with* `-gfull`; this is the archive-level half of the same shape — the objects hit, the archive they form still re-keys per checkout.
- Workaround until fixed: `CFLAGS=-g0` / `CXXFLAGS=-g0` in `~/.cargo/config.toml` `[env]` (drops C/C++ debug info for build-script compiles only).
- Fix proposed in the linked PR: admit `__DWARF,__debug_*` / `__apple_*` sections flagged `S_ATTR_DEBUG` to the shared Mach-O gate; STABS, bitcode/LTO carriers, ARM64_32, and every other spelling of a debug section stay fail-closed.

yanyaoer:
## Summary

Fixes #998.

On macOS every `cc`-built `static=` archive whose objects carry DWARF (all dev-profile builds; `ring` on Apple targets unconditionally via `-gfull`) fails the Mach-O gate in `portable_static_archive_hash` and takes the `path-ar-v1` fallback, which binds the lexical absolute archive path. The crate's rlib key then differs per checkout even though the archives are byte-identical, and the miss cascades through every dependent's `--extern` content hash — the per-checkout re-key that #691's closing comment asked to confirm.

This admits DWARF to the shared Mach-O object gate, for both the BSD and the GNU arm, and invalidates the memo rows that would otherwise keep the old answer:

- `macho_segment_is_portable` no longer rejects `__DWARF`;
- a new `macho_section_is_dwarf` accepts exactly clang's spelling — a `__debug_*` or `__apple_*` section in the `__DWARF` segment flagged `S_ATTR_DEBUG` — and `macho_section_is_portable` admits it through an early return, so the existing conjunction is untouched; every other spelling of a debug section (debug name in a code segment, `__DWARF` section without the flag, `__zdebug_*`) stays fail-closed;
- STABS (`N_STAB` symbols), bitcode/LTO carriers, ARM64_32, and unknown shapes are unchanged;
- the static-lib memo namespace moves `static-ar-v5` → `static-ar-v6` (`src/cache_key.rs`): a v5 row would keep serving the path-bound digest of an unchanged archive, so existing builds would never converge — the same treatment the ELF and Mach-O LTO gates got before; `hash_static_lib_ignores_legacy_namespaces` covers v5;
- `is_known_no_debug_macho_object` keeps its name (its doc comment now says what it proves: "no debug" is about STABS). Renaming it to `…_portable_…` is the one cosmetic change I kept out of the diff — say the word if you'd like it in. The module docs and the BSD-arm comment carry the argument below.

Why this is safe:

- The DWARF bytes are hashed as member content like everything else, so a path a compiler left in `DW_AT_comp_dir` is identity-bearing — it can only cause a miss, never a false hit.
- The `archive(member)` `N_OSO` record ld64 writes for a debug-bearing member was the reason for the exclusion. Its path half is already checkout-independent for every cached macOS debug link: `-oso_prefix` relative to `--out-dir` (`src/compiler/rustc.rs`) plus the store-time `.dSYM` (#319). That is how the debug-bearing Rust objects in every dev-profile rlib are handled today, and once rustc bundles a `static=` archive into the rlib its members are indistinguishable from those. The member name and timestamp — the other half of the record — were already committed to the hash by both arms.
- A passthrough link keeps absolute `N_OSO` paths, but nothing is restored from a passthrough, so no restored binary can end up with a wrong debug map.

No hashed-content definition changes for archives the arms already accepted, so `BSD_DOMAIN` / `GNU_DOMAIN` and `CACHE_KEY_VERSION` are left alone; archives that previously took the fallback simply re-key once from `path-ar-v1` to `bsd-ar-v2` / `gnu-ar-v2`. Happy to bump if you prefer a hard cut.

## Test plan

- [x] `cargo test -p kache --bin kache native_archive` and `… hash_static_lib` — `bsd_debug_sections_and_stabs_fall_back` becomes `bsd_dwarf_hashes_structurally_and_stabs_fall_back` (DWARF member → `bsd-ar-v2:` / `gnu-ar-v2:`, DWARF bytes re-key on change, STABS still fall back with and without DWARF under both layouts); new `only_clang_shaped_dwarf_is_admitted` (segment, flag, and name each required); `macho_section_policy_checks_each_marker_independently` extended with direct `macho_section_is_dwarf` cases so each conjunct and the `||` are observable to mutants; `hash_static_lib_ignores_legacy_namespaces` gains the v5 sentinel.
- [x] `just check` (fmt, clippy `-D warnings`, workspace tests)
- [x] Real archive, `ring 0.17.14`, two checkouts of the same commit, macOS 26.6.2 / Apple clang 21 / rustc 1.95, `CFLAGS` untouched (objects carry DWARF): before, `link_lib_content:static=ring_core_0_17_14_=path-ar-v1:…` differs per checkout and the second build is a `dup`; after (patched binary as `RUSTC_WRAPPER`, isolated store, objects carrying 8 `__DWARF` sections), both checkouts hash `link_lib_content:static=ring_core_0_17_14_=bsd-ar-v2:5dfd5b…`, `ring`'s `final=2f9c5a4e26657596` converges, and the second checkout's rebuild is a `local_hit` (1.4 s against a 17.6 s miss). Before/after trace excerpts in #998.
- [ ] Mutation shards (CI) — opened as a draft per AGENTS.md.

## Checklist

- [x] `just check` passes locally (mutants left to CI per AGENTS.md)
- [x] Commit messages follow conventional commits
- [x] User-visible changes are reflected in the docs — none; the module docs carry the design change

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/cache_key.rs
+++ b/src/cache_key.rs
@@ -3806,12 +3806,14 @@ impl<'db> FileHasher<'db> {
         // things — `hash` stores plain blake3, this stores a structural or
         // path-bound archive digest). This restores the warm-build fast path the whole-file
         // hasher had: an unchanged large `static=` archive (e.g. rocksdb) is not
-        // re-read on every incremental build. `v5`: v1/v2 rows used older
-        // identity definitions, v3 predates the fail-closed ELF gate, and v4
-        // predates the GCC Mach-O LTO gate. None may be served after the final
-        // archive hardening.
+        // re-read on every incremental build. `v6`: v1/v2 rows used older
+        // identity definitions, v3 predates the fail-closed ELF gate, v4
+        // predates the GCC Mach-O LTO gate, and v5 predates admitting
+        // DWARF-bearing Mach-O members (a v5 row would keep serving the
+        // path-bound digest of an unchanged archive). None may be served
+        // after the final archive hardening.
         let key = FileFingerprint {
-            path: format!("static-ar-v5\0{}", fingerprint.path),
+            path: format!("static-ar-v6\0{}", fingerprint.path),
             size: fingerprint.size,
             mtime_ns: fingerprint.mtime_ns,
             ctime_ns: fingerprint.ctime_ns,
@@ -7632,8 +7634,12 @@ mod tests {
             path: format!("static-ar-v4\0{}", fingerprint.path),
             ..fingerprint.clone()
         };
-        let current_key = FileFingerprint {
+        let legacy_v5_key = FileFingerprint {
             path: format!("static-ar-v5\0{}", fingerprint.path),
+            ..fingerprint.clone()
+        };
+        let current_key = FileFingerprint {
+            path: format!("static-ar-v6\0{}", fingerprint.path),
             ..fingerprint
         };
         let cache = fh.cache.as_ref().expect("persistent cache opens");
@@ -7647,6 +7653,9 @@ mod tests {
         cache
             .put(&legacy_v4_key, "legacy-unguarded-macho-sentinel")
             .unwrap();
+        cache
+            .put(&legacy_v5_key, "legacy-path-bound-dwarf-sentinel")
+            .unwrap();
 
         let computed = fh.hash_static_lib(&lib).unwrap();
         assert!(computed.starts_with("gnu-ar-v2:"));
@@ -7654,6 +7663,7 @@ mod tests {
         assert_ne!(computed, "legacy-member-sentinel");
         assert_ne!(computed, "legacy-unguarded-object-sentinel");
         assert_ne!(computed, "legacy-unguarded-macho-sentinel");
+        assert_ne!(computed, "legacy-path-bound-dwarf-sentinel");
         assert_eq!(cache.get(&current_key).unwrap(), Some(computed.clone()));
         assert_eq!(fh.hash_static_lib(&lib).unwrap(), computed);
     }
--- a/src/native_archive.rs
+++ b/src/native_archive.rs
@@ -49,12 +49,19 @@
 //! header size INCLUDES those bytes. The path-derived `cc` name therefore
 //! lives inside the member data itself. The exact stored name bytes (including
 //! encoding/padding), parsed timestamp, and object DATA are all hashed.
-//! - only structurally valid, known Mach-O `MH_OBJECT` files
-//!   without debug sections, STABS, or embedded compiler bitcode/LTO. Darwin's
-//!   linker
-//!   records `archive(member)` (and archive-member time) in `N_OSO` debug-map
-//!   entries. Unsupported, malformed, debug-bearing, bitcode, and ARM64_32
-//!   objects therefore use the path-bound fallback;
+//! - only structurally valid, known Mach-O `MH_OBJECT` files without STABS
+//!   or embedded compiler bitcode/LTO. DWARF (`__DWARF,__debug_*` /
+//!   `__apple_*`, S_ATTR_DEBUG) is hashed like any other member bytes, so a
+//!   path a compiler left in `DW_AT_comp_dir` is identity-bearing, never a
+//!   false hit. Darwin's linker records `archive(member)` (and archive-member
+//!   time) in `N_OSO` debug-map entries for such members, but every cached
+//!   macOS debug link already makes those records checkout-independent —
+//!   ld64 `-oso_prefix` relative to `--out-dir` (`compiler/rustc.rs`) plus
+//!   the store-time `.dSYM` (#319) — exactly as for the debug-bearing Rust
+//!   objects every dev-profile rlib bundles; a `cc`-built archive is no
+//!   different once rustc has bundled its members into the rlib. Unsupported,
+//!   malformed, STABS, bitcode, and ARM64_32 objects therefore use the
+//!   path-bound fallback;
 //! - the ranlib member (`__.SYMDEF[ SORTED]` / `__.SYMDEF_64[ SORTED]`) DATA
 //!   **as-is**, tagged with its trimmed name: `_64` reads the same bytes with
 //!   a different word width and ` SORTED` changes the lookup contract, so the
@@ -71,7 +78,7 @@
 //!
 //! ## Out of scope -> path-bound fallback (`None`)
 //! Thin archives (`!<thin>`), Windows COFF `.lib`, GNU/BSD mixed layouts,
-//! debug-bearing Mach-O, compiler bitcode/LTO, ARM64_32, unknown object formats,
+//! STABS-bearing Mach-O, compiler bitcode/LTO, ARM64_32, unknown object formats,
 //! and anything malformed. Thin archives are made uncacheable because their
 //! external member bytes are absent from the container. Other fallbacks bind
 //! both the archive bytes and lexical absolute archive path.
@@ -206,7 +213,7 @@ fn gnu_archive_hash(bytes: &[u8]) -> Option<String> {
                 // bitcode uses archive-path-derived LTO identifiers, and an
                 // unknown format may have equally path-sensitive semantics.
                 let known_object = if has_macho_magic(data) {
-                    is_known_no_debug_macho_object(data)
+                    is_known_portable_macho_object(data)
                 } else {
                     is_known_elf_relocatable_object(data)
                 };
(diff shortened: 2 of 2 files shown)


jleni (commented):
Thanks for this, and for the careful write-up in #998. I built your head and ran two checkouts of a crate that bundles a `cc -g` archive into an rlib. The lib and the downstream bin both hit in the second checkout, and the bin's debug map names `libosoprobe-<hash>.rlib(probe.o)`, which `-oso_prefix` already makes relative. So the rlib case, including `ring`, checks out. The gate conjunction, the v5→v6 memo bump, and leaving `CACHE_KEY_VERSION` alone all look right to me, and your targeted tests pass locally.

One thing needs to change before this lands. `link_lib_content` is folded for every crate type, not only rlibs. When a bin, test, or cdylib crate links its own build script's archive directly, ld64 records the archive's absolute path. `-oso_prefix` only strips `--out-dir` (`target/debug/deps/`, `src/compiler/rustc.rs:409-417`), and OUT_DIR is under `target/debug/build/<pkg>/out/`.

Repro: a bin crate whose build.rs compiles `probe.c` with `-g` and path-normalized DWARF (as `kache cc` produces) into `OUT_DIR/libprobe.a` and links `static=probe`. Two copies share one store. The second copy gets `local cache hit for osoprobe`, and its binary's debug map reads:

```
N_OSO '<S>/a/target/debug/build/osoprobe-b0f9cce596d10f95/out/libprobe.a(probe.o)'
```

Before this PR, `path-ar-v1` kept that key per-checkout. The restored `.dSYM` hides the problem while it exists. Packaging is best-effort, though (`src/compiler/platform.rs:222-235`), and without a `.dSYM` lldb follows the path into the other checkout. It would also make `docs/how-it-works/cache-key.mdx:26` untrue for these binaries.

Suggested fix: keep admitting DWARF members when the output is an rlib, and keep the path-bound digest when the invocation itself links (`is_executable_output()`, which covers bin, test, dylib, cdylib and proc-macro). `hash_static_lib` is memoized per file, so the simplest shape may be for the structural hash to report whether any member carried DWARF, and let the caller at `src/cache_key.rs:1627` choose. A test that folds a DWARF archive into a bin key and asserts it stays path-bound would pin it. `ring` still converges with that change, since it is an rlib.

I read the code and reproduced both shapes on macOS. I didn't run the mutation shards.

Happy to take this over and finish it if you'd rather. Just say so.

*(AI-assisted review. A maintainer still owns the call.)*

jleni:
Thanks @yanyaoer. I've taken this over as offered in the review: https://github.com/kunobi-ninja/kache/pull/1021 carries your two commits with your authorship, plus a maintainer commit that keeps direct links path-bound. I'll leave this PR open until that one lands.

jleni:
This landed in #1021 as 4ec5500 (squash merge). Your two commits are credited through a `Co-authored-by: yanyao <yanyao@xiaomi.com>` trailer, and the commit message keeps both source hashes (37c9c6b, 76f3095). That email isn't linked to your GitHub account, so GitHub may not show the co-author badge.

The maintainer commit on top keeps the path-bound digest when an invocation links a DWARF archive itself (bin, test, cdylib, and so on), so `ring` and other rlibs converge across checkouts and direct links stay correct.

Thanks again for the careful fix and the write-up in #998. Closing this one as superseded by #1021.

PR #1043, opened 2026-09-13
fix(shims): keep Homebrew compiler shims upgrade-safe

achembarpu:
## Summary

Make compiler shims installed by Homebrew Kache survive formula upgrades.

- When Kache runs from a Homebrew Cellar keg, create shims through the matching `opt/<formula>/bin/kache` path instead of the versioned keg path.
- Preserve the formula name, so stable (`kache`) and prerelease (`kache-unstable`) installations each use their own opt target.
- Treat the active opt target as the expected shim identity, allowing `kache init` run from an older keg to refresh stale shims after an upgrade.
- Keep non-Homebrew installation behavior unchanged.

## Verification

- `just check`
- `cargo test shim_install_tests`
- Manual mutation check: changing the opt path makes the targeted test fail.
- Simulated Homebrew Cellar-to-opt end-to-end `kache install-shims` run.


jleni:
thanks @achembarpu !

jleni (approved):
Thanks for the fix, Arvind. I checked the fork's workflow files against the base repository and verified the new Homebrew target logic, including the separate `kache-unstable` formula.

Safety: CLEAR
Purpose: PARTIAL — newly created shims survive upgrades; existing versioned shim links need a one-time migration, which we will handle in a follow-up PR.
Tests: COVERED for the new opt-path behavior.
Docs: NOT REQUIRED — the current instructions remain accurate.

I'm approving this scoped change with that migration tracked separately.

(AI-assisted review — a maintainer still owns the call.)

achembarpu:
@jleni  Thanks for merging! I also have an open PR on the tap to fix version staleness for the launch service by using homebrew directly: https://github.com/kunobi-ninja/homebrew-kunobi/pull/99

PR #1100, opened 2026-09-17
Forward rustc metadata notifications

blaginin:
## Summary

Forward rustc metadata notifications while capturing stdout and stderr concurrently, so cargo can start compiling other crates before upstream finishes

## Test plan

On my local project (~813k Rust LOC, 131 workspace crates), rebuilding five affected crates after a branch switch fell from 67.35 s to 60.78


jleni:
This is fantastic @blaginin !!! Let me have a look!

And please!! keep them coming! It is great to see these contributions and performance improvemnts!!

jleni:
Thanks, this is a good change: buffering stderr until exit has been blocking Cargo's pipelining on every miss.

Two things before merge:

1. `tests/metadata_streaming_test.rs` builds the expected path with `join("out/libmetadata_fixture.rmeta")`. On Windows rustc reports `out\libmetadata_fixture.rmeta`, so the comparison will likely fail there. Joining `"out"` and the file name separately should fix it.
2. In `run_rustc`, if the stderr loop or the stdout thread stops on a read error, the pipe is no longer drained and `child.wait()` can block forever while rustc waits on a full pipe. The old `wait_with_output` returned an error in that case. Could you keep draining (or kill the child) when a reader fails?

Optional: `pending_stderr: Option<String>` might read more simply as a `String` that always holds what is left to replay.

jleni:
We have some problems with mutation testing... I know it can be frustrating :/ but this avoids a lot of problems later one.

PR #1156, opened 2026-09-20
ci: stop treating a public dev branch as a release lane

jleni:
#1153 added `dev` next to `main` in CI and Service Image. There is no public `dev` branch (see CONTRIBUTING).

This restores `main` + tags as the only push lanes. The Zondax `service` branch keeps its own workflow copy for integration CI.


jleni:
Superseded: rewritten `main` already has no public `dev` trigger.

